PR #8, opened 2017-03-04
fix include in installed header

ibaned:
The headers in kokkos-kernels/src/impl/ get
installed to the include/ directory same
as the rest, so they should not include
each other as "impl/foo.hpp" because that
will not work in the installed location.
This fixes a compile error when building
Albany against current Trilinos.

@crtrott @mhoemmen 
Its pretty urgent that this gets through to Trilinos,
I'm pretty sure all compiles against installed
kokkos-kernels will fail until then.
The Albany CDash failure that spotted this is here:
http://my.cdash.org/viewBuildError.php?buildid=1146035

### DIFF
--- a/src/impl/Kokkos_Sparse_impl_spmv_def.hpp
+++ b/src/impl/Kokkos_Sparse_impl_spmv_def.hpp
@@ -47,7 +47,7 @@
 #include "Kokkos_InnerProductSpaceTraits.hpp"
 #include "Kokkos_CrsMatrix.hpp" // only included for backwards compatibility
 #include "Kokkos_Blas1_MV.hpp"
-#include "impl/Kokkos_Sparse_impl_spmv_omp.hpp"
+#include "Kokkos_Sparse_impl_spmv_omp.hpp"
 
 #ifdef HAVE_KOKKOSKERNELS_ETI_ONLY
 #define KOKKOSSPARSE_ETI_ONLY


mhoemmen:
@ibaned wrote:
> I'm pretty sure all compiles against installed kokkos-kernels will fail until then.

It's legal to push KokkosKernels changes to Trilinos/develop if urgent.  Just note that the changes will go away at the next snapshot unless they also go into KokkosKernels.

mhoemmen:
@ibaned we all just got back from work travel a couple days ago so please pardon the delay.

crtrott:
We apply the fix to trilinos separately.

ibaned:
I'm currently running the checkin-test-sems script to push this to Trilinos develop. Unfortunately @bmpersc merged Trilinos develop into master before I could do this, so applications that use Trilinos may see trouble with Trilinos master soon.

ibaned:
That was taking way too long. I just pushed 0a2eee9.

mhoemmen:
> That was taking way too long. I just pushed 0a2eee9.

Sorry, we're all just getting back from travel (and I just moved offices -- I've been unpacking boxes this morning).

ibaned:
I'm sorry, that statement was vague. I was referring to the checkin-test-sems process, not the responsiveness of developers. I have no problem with the latter.

mhoemmen:
Next time, if you're in a hurry, try adding these options:

`--enable-packages=KokkosKernels,Tpetra,Ifpack2 --no-enable-fwd-packages`

PR #38, opened 2017-07-20
KokkosKernels: Performance Testing Prototype

MicheldeMessieres:
@srajama1 

This is a first implementation of a performance archiver based on
Teuchos_XMLPerfTestArchive but using yaml instead of XML.

**Summary:**

  On first run, the Performance archiver adds the test info to a yaml database.

  On subsequent runs the test configuration is compared to prior results.
    If the configuration does not match a new entry is added to the database.
    If the configuration matches, the values are compared and failure events are
      provided if they exceed the specified tolerances.

**Unit Tests/Demos** (enabled with TPL_ENABLE_yaml-cpp):

  _performance_demo_ - a lightweight example which uses mock values and
    demonstrates usage of the Performance class. Eventually to be deleted.

  _performance_validate_ - a true unit test which runs through all the possible
    functions of the archiver and verifies proper response. The test is designed
    to mimic running one test repeatedly and validates that new input is properly
    compared to old values in the database. This test may serve as a good
    guide for learning which features are currently available with the archiver.

  _performance_demo_blas1_ - takes the blas1 example and applies the
    Performance archiver to it. Currently uses the Teuchos TimeMonitor outputs
    as inputs for the archiver. This test will be deleted and the performance class
    can be applied directly to the blas1 test once further design discussion has
    taken place. This was intended as a minimal application to a real test.

This is work in progress and needs further discussion.


perf_test/performance_demo.cpp
```diff
@@ -66,7 +66,7 @@
 
 #include "Kokkos_Performance_impl.hpp" // provides performance archiver
 
-bool run_example() {
+bool run_example(Teuchos::RCP<const Teuchos::Comm<int> > comm) {
```
mhoemmen: I'm pretty sure Kokkos isn't supposed to depend on Teuchos.

perf_test/performance_demo.cpp
```diff
@@ -86,8 +86,12 @@ bool run_example() {
   // Create an archiver - steps are create, fill with members, then run
   Performance archiver;
 
+  // Example of how to set customized machine config - to be developed
+  // Change to generate new entries in the yaml under MachineConfiguration
+  archiver.set_machine_config("Kokkos Config", "some node type");
+
   // Fill config
-  archiver.set_config("MPI_Ranks", 2); // would be set to mpi comm size
+  archiver.set_config("MPI_Ranks", comm->getSize());
```
mhoemmen: See above.

perf_test/performance_demo.cpp
```diff
@@ -150,7 +154,7 @@ int main(int argc, char *argv[]) {
 
   // for now other ranks can just be done - let rank 0 update the archive
   if(comm->getRank() == 0) {
```
mhoemmen: See above.

mhoemmen:
@MicheldeMessieres never mind my review -- it looks like you don't intend for this to depend only on kokkos-kernels.

MicheldeMessieres:
@mhoemmen - I think the plan is to have them not use Teuchos at all and the yaml reader is free of Teuchos. I had picked the blas1 test somewhat arbitrarily and just realizing now it's the only one that has Teuchos links so probably a poor first choice. To make it more clear this runs Teuchos free I removed the blas1 demo and updated the other two demos to be without any Teuchos calls. 

I think a good next step will be to pick one of the other tests, sketch out the features we'd like to see, then I could work on making sure that runs nicely with the archiver in different Kokkos modes, etc.

mhoemmen:
@MicheldeMessieres Thanks for the clarification!

crtrott:
Hm the diff seems weirdly off, it shows me stuff which is already in develop ??

MicheldeMessieres:
@crtrott Regret the mess. Was trying to push from a different server and got mangled. I will clean this up soon.

crtrott:
OK closing this. Check out #44 , I took your branch merged it into develop and summited as a new one. It just felled off that there are supposed to be over 100k lines changed for this, and the fixed up pull request only has 1.3k lines change, which makes more sense.

crtrott:
Still funny how this could happen. Somehow git got confused, because a simple merge from your branch worked perfectly fine, so its not the actual content ..

MicheldeMessieres:
@crtrott  I'll work off the new branch. I had forgotten to set authorship originally and amended before pushing which perhaps caused some issue, though the diff seemed fine.  Now it looks like it's supposed to. Thanks!

Issue #31, opened 2017-05-21
ETI System and file structure


crtrott:
Ok I think I am finally close to make this ETI stuff work properly. There is some funky compiler stuff with regards to using extern template instantiations for classes, in particular if you want to allow instantiations of other types but I believe my solution is now fool proof ......

Furthermore I believe the file structure and naming etc needs some cleanup. In particular this focus on MultiVector which historically comes from Tpetra is confusing for standalone users. 

Lets start with some requirements what we need to be able to do::
* pre-compile functions, and prevent them from being implicitly instantiated (ETI)
* Even with ETI on, allow other input types (say for example extended precision, or nonstandard data layouts)
* call TPLs (MKL, CUBLAS etc.) for input types which allow it
* disallow anything other than ETI types if requested
* check what type of instantiation gets hit in apps (ETI, Non-ETI, TPL)

In order to do all this we came up with a design which has  3 functionality layers (I will go into details later):

1. User Interface: void foo(ViewType a, Scalar alpha): takes views accepts all kinds of combinations; calls the specialization layer
2. Specialization Layer: struct Foo { static void foo(ViewInternalType a, Scalar alpha); }; makes sure that only the minimally necessary number of instantiations exists, serves as ETI specialization layer, serves as TPL specialization layer
3. Implementation Layer: This is called by the specialization layer, and has the actual functors etc.

Now I want to go through a couple of design aspects in the next posts.

crtrott:
I thought now long and hard about a sane, sustainable way for organizing this into files. Here is what I came up with:
```
src:
  KokkosBlas.hpp: includes all the KokkosBlas function header files
      KokkosBlas1_foo.hpp (contains user interface functions for foo)
src/impl:
  KokkosBlas1_foo_impl.hpp: The actual implementation of the functions (Functors etc.)
  KokkosBlas1_foo_spec.hpp: The specialization layer
src/impl/tpl
  KokkosBlas1_foo_spec_tpl_avail.hpp: Availability of TPLs for particular types
  KokkosBlas1_foo_spec_tpl_decl.hpp: The Specialization declaration for using tuples
src/impl/generated_specializations_hpp
  KokkosBlas1_foo_spec_avail.hpp: Availability declarations for ETI types
  KokkosBlas1_foo_spec_declhpp: Specialization declarations for ETI types
src/impl/generated_specializations_cpp/foo
[...]

PR #40, opened 2017-07-31
KokkosSparse Updates

mndevec:
- ETI modifications for Sparse classes for spmv and trsv functions.
- Unit tests for spmv and trsv
- ETI generation scripts for sparse classes that requires ordinal and
  size types.
  scripts/generate_specialization_function_sparse.bash
  scripts/generate_specialization_type_sparse.bash
  The use examples are in
  scripts/generate_specialization2.bash

- Restructure of some folders, and filenames. We will need to change
  some includes in Trilinos from Kokkos_Sparse* to KokkosSparse*
  Moved KokkosBlas files to blas folder.


mhoemmen:
@mndevec If you want to change header file names, please add backwards compatibility headers with the original names that include the headers with the new names.  Thanks!

mndevec:
Sure Mark, I will add the list of the changed headers, maybe here? 
By the way, #31 talks about the new standard for file names and namespaces. I am following those for now.

srajama1:
Next release is the first major one. IMHO, we can break backwards compatibility in that as long as we change Trilinos packages that include the headers.

crtrott:
I am seconding Siva here. The old naming stuff was just a mess. We need to get this on a clean slate. And we will fix everything inside of Trilinos for the Release, and outside of it we don't have many users, and the ones we have are prepared to adapt.

PR #54, opened 2017-08-23
Batched modification

kyungjoo-kim:
@ambrad This will fix your problem in ifpack2.

### DIFF
--- a/src/batched/KokkosBatched_Util.hpp
+++ b/src/batched/KokkosBatched_Util.hpp
@@ -43,6 +43,12 @@ namespace KokkosBatched {
 #pragma message "KOKKOS_BATCHED_DISABLE_AVX512"
 #endif
 
+#if defined(__FMA__)
+#pragma message "KOKKOS_BATCHED_ENABLE_FMA"
+#else
+#pragma message "KOKKOS_BATCHED_DISABLE_FMA"
+#endif
+
     // #ifdef KOKKOS_ENABLE_PRAGMA_IVDEP
     // #pragma ivdep
     // #endif
--- a/unit_test/batched/Test_Batched_VectorArithmatic.hpp
+++ b/unit_test/batched/Test_Batched_VectorArithmatic.hpp
@@ -42,71 +42,71 @@ namespace Test {
         /// test : vec + vec
         c = a + b;
         for (int k=0;k<vector_length;++k) 
-          EXPECT_NEAR_KK( c[k], a[k]+b[k], eps);      
+          EXPECT_NEAR_KK( c[k], a[k]+b[k], eps*c[k]);      
       
         /// test : scalar + vec
         c = alpha + b;
         for (int k=0;k<vector_length;++k) 
-          EXPECT_NEAR_KK( c[k], alpha+b[k], eps);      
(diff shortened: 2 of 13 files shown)


src/batched/KokkosBatched_Util.hpp
```diff
@@ -43,6 +43,12 @@ namespace KokkosBatched {
 #pragma message "KOKKOS_BATCHED_DISABLE_AVX512"
 #endif
 
+#if defined(__FMA__)
+#pragma message "KOKKOS_BATCHED_ENABLE_FMA"
```
ambrad: I'm OK with these pragma messages, but are others?

unit_test/batched/Test_Batched_VectorArithmatic.hpp
```diff
@@ -86,24 +86,24 @@ namespace Test {
         for (int k=0;k<vector_length;++k) 
           EXPECT_NEAR_KK( c[k], b[k]*alpha, eps);      
       }
-      {
-        /// test : vec / vec
-        c = a / b;
-        for (int k=0;k<vector_length;++k) 
-          EXPECT_NEAR_KK( c[k], a[k]/b[k], eps);      
+      // {
```
ambrad: This is fine, but I'm just checking to make sure you want this commented-out code.

src/batched/KokkosBatched_Util.hpp
kyungjoo-kim: I put the message (I saw this from bcrs mimic sparc) because I want to see what intrinsics are used when compiling this code.

unit_test/batched/Test_Batched_VectorArithmatic.hpp
kyungjoo-kim: I removed the whole division test as it is incomplete. 
- division complex (SIMD, SIMD) is fine
- division complex (SIMD, Scalar) is not fine due to Kokkos complex templates allows all possible combinations
- division complex (AVX, AVX) is not yet impleted. 

If you want to test division for double, then I can guard it for double and float only.

src/batched/KokkosBatched_Util.hpp
srajama1: This is ok with me.

srajama1: May be you can make it "Using ...."

nmhamster: @ambrad @kyungjoo-kim @srajama1 - guys, you know that even with all this front end handling, the compiler may choose to fuse intrinsics/instructions during the final code generation/optimisation passes? The only to really stop FMAs being generated is to use `-fno-fma` I think which tells the code generation it is not permitted to create any instructions of this type. Just want you to know what you might end up running evening if it isn't explicitly encoded.

kyungjoo-kim: @nmhamster @srajama1 @ambrad 

this is not about "not generating fma" but just "showing a message" that fma instruction is available. When benchmarking code, I would like if fma is available or not while compiling the code and this throw the message.

PR #57, opened 2017-08-30
numerous fixes...

kyungjoo-kim:
1. remove pragma note for compiler setting
2. debugged trsv  and unit test for trsv and gemv
3. complex overload for simd ( need to provide full complex overload not to be ambiguous with kokkos complex)
4. tested with @ambrad  ifpack2 patch and pass the unit test.

what else do I miss ?

### DIFF
--- a/src/batched/KokkosBatched_Util.hpp
+++ b/src/batched/KokkosBatched_Util.hpp
@@ -16,6 +16,7 @@
 
 #include "Kokkos_Core.hpp"
 #include "Kokkos_Complex.hpp"
+#include "Kokkos_ArithTraits.hpp"
 #include "impl/Kokkos_Timer.hpp"
 
 namespace KokkosBatched {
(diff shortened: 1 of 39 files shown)


src/batched/KokkosBatched_Util.hpp
```diff
@@ -24,42 +25,43 @@ namespace KokkosBatched {
...
+    void print_compiler_info();
 
-#if defined(__AVX512F__)
-#pragma message "KOKKOS_BATCHED_ENABLE_AVX512"
-#else
-#pragma message "KOKKOS_BATCHED_DISABLE_AVX512"
-#endif
+    // we do not allow mixed precision
+    template<typename Ta, typename Tb>
+    struct is_convertible {
```
mhoemmen: Calling this `is_convertible` will break user code that does `using namespace std;`.  I know users aren't supposed to do that, but they do :-) .

crtrott:
So Mark you are OK with the way it is?

mhoemmen:
> So Mark you are OK with the way it is?

Yes; thanks!

Issue #66, opened 2017-08-31
generate makefile tpl option


mndevec:
Is there a way of defining MKL and cuSPARSE tpls on generate makefile?

Issue #55, opened 2017-08-28
Build warnings in sparse matrix-matrix multiply when MKL TPL is enabled


mhoemmen:
```
.../Trilinos/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp:111:62: warning: unused typedef 'device3' [-Wunused-local-typedef]
    typedef typename in_nonzero_value_view_type::device_type device3;
                                                             ^
.../Trilinos/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp:99:36: warning: unused typedef 'idx_array_type' [-Wunused-local-typedef]
    typedef in_row_index_view_type idx_array_type;
                                   ^
.../Trilinos/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp:110:62: warning: unused typedef 'device2' [-Wunused-local-typedef]
    typedef typename in_nonzero_index_view_type::device_type device2;
                                                             ^
.../Trilinos/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp:109:58: warning: unused typedef 'device1' [-Wunused-local-typedef]
    typedef typename in_row_index_view_type::device_type device1;
                                                         ^
In file included from .../CHECKIN-CLANG-3.9.0/MPI_DEBUG_REAL/packages/tpetra/core/src/Tpetra_Details_packCrsMatrix_DOUBLE_INT_LONG_LONG_SERIAL.cpp:71:
In file included from .../Trilinos/packages/tpetra/core/src/Tpetra_Details_packCrsMatrix_def.hpp:52:
In file included from .../Trilinos/packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp:64:
In file included from .../Trilinos/packages/kokkos-kernels/src/sparse/KokkosSparse.hpp:60:
In file included from .../Trilinos/packages/kokkos-kernels/src/sparse/KokkosSparse_spgemm.hpp:51:
.../Trilinos/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_mkl2phase_impl.hpp:97:56: warning: unused typedef 'device1' [-Wunused-local-typedef]
  typedef typename in_row_index_view_type::device_type device1;
                                                       ^
[...]

PR #70, opened 2017-09-01
Fixes to MKL warning for issue #55

mndevec:
I needed to compile with gcc and enable MKL. I could not get the warnings with intel compiler.

I compiled it within trilinos, as I don't think we have a way of setting tpls from generate makefile -> #66

### DIFF
--- a/src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp
+++ b/src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp
@@ -121,7 +121,8 @@ namespace Impl{
       return;
     }
 */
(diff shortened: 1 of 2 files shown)


src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp
```diff
@@ -121,6 +121,7 @@ namespace Impl{
       return;
     }
 */
+    std::cout << "runing mkl" << std::endl;
```
mhoemmen: Please remove the extra std::cout statement before pushing.  Thanks!

mndevec:
@mhoemmen sorry for the forgotten couts, it should be okay now.

Issue #76, opened 2017-09-06
Testing with memkind library in bowman


mndevec:
For one of our projects, we are using kokkos-kernels with multilevel memory systems, such as Cuda Pinned Memory as well as HBM and DDR memory using memkind library. 

On bowman, we compile it with below to explicitly manage HBM and DDR memory allocations. 
  --with-memkind=/home/projects/x86-64-knl/memkind/20160811/gcc/4.8.3/

kokkos-kernels compilation currently fails if we enable this. Would that be possible to support at least the compilation of this? Could we also add a jenkins test on bowman that practices this?

@ambrad @kyungjoo-kim @crtrott @srajama1

nmhamster:
@mndevec - when you say it fails, what happens?

mndevec:
The compilation fails for some of the batched kernels. We don't use them, but the library compilation fails.

mndevec:
```
icpc -I./ -I/home/mndevec/work/kokkos-kernels/example/buildlib/kokkos/install/include -I/home/mndevec/work/kokkos-kernels/example/buildlib/kokkos/install/include -I/home/mndevec/work/kokkos-kernels/example/buildlib/kokkos/install/include -I/home/projects/x86-64-knl/memkind/20160811/gcc/4.8.3//include -I/home/mndevec/work/kokkos-kernels/example/buildlib/install/include --std=c++11 -xMIC-AVX512 -fopenmp -I/home/mndevec/work/kokkos/tpls/gtest -I/home/mndevec/work/kokkos-kernels/unit_test/ -I/home/mndevec/work/kokkos-kernels/unit_test/blas -I/home/mndevec/work/kokkos-kernels/unit_test/sparse -I/home/mndevec/work/kokkos-kernels/unit_test/graph -I/home/mndevec/work/kokkos-kernels/unit_test/common -I/home/mndevec/work/kokkos-kernels/unit_test/batched -I/home/mndevec/work/kokkos-kernels/unit_test/openmp -Wall -pedantic -Werror -O3 -g -Wshadow -Wsign-compare -Wignored-qualifiers -Wempty-body -Wclobbered -Wuninitialized  -c /home/mndevec/work/kokkos-kernels/unit_test/openmp/Test_OpenMP_Batched_VectorArithmatic.cpp
icpc: command line warning #10006: ignoring unknown option '-Wignored-qualifiers'
icpc: command line warning #10006: ignoring unknown option '-Wempty-body'
[...]

PR #80, opened 2017-09-08
Batched - disable team trsv test + fix for issue #76

kyungjoo-kim:
With a half day debugging with Christican, I learned that I should not use or at least avoid from using the same functor name. When the same functor names are used and linked together, it link to any functor randomly. So, even if I thought I test team interface, it mostly test serial interface only as I use the same functor name for both. 

After changing the functor names, team level trsv is not correct and I disabled those test.


kyungjoo-kim:
Ah... I also want to break the test code into two. I will update this branch.

kyungjoo-kim:
No it can be merged  after you review

PR #83, opened 2017-09-12
Spgemm and SGS eti 

mndevec:
XL compiler compilation time should be significantly reduced.


mndevec:
@kyungjoo-kim  you can use this branch for debugging XL issues.

PR #84, opened 2017-09-18
Multi level memory instantiations 

mndevec:
On KNLs, to enable explicit HBM and DDR management: run generate makefile with:
    --with-memkind=/home/projects/x86-64-knl/memkind/20160811/gcc/4.8.3/
    
    This will add HBWSpace to kokkos devices, and ETI'ed functions will instantiate
    for both HostSpace and HBWSpace. I updated these for all blas and sparse functions,
    because HBWSpace becomes the default memory space when memkind is enabled, and
    instantiating HBWSpace becomes necessity.
    
    For GPUs, use KOKKOS_DEVICES=CudaHostPinnedSpace,...
    This will generate ETI functions for this memory space as well.
    This is only added to multi-level memory sparse functions. This is an optional space,
    and "NOT Necessary" to instantiate for blas kernels. I did not add
    them for this reason.

With this set of commits, we can set a Jankins test.


scripts/generate_makefile.bash
```diff
@@ -66,6 +66,7 @@ do
       HWLOC_PATH="${key#*=}"
       ;;
     --with-memkind*)
+      KOKKOS_DEVICES="${KOKKOS_DEVICES},HBWSpace"
```
crtrott: I think this doesn't do anything useful

scripts/generate_makefile.bash
```diff
@@ -278,6 +279,7 @@ if [ ${#HWLOC_PATH} -gt 0 ]; then
   KOKKOS_USE_TPLS="${KOKKOS_USE_TPLS},hwloc"
 fi
 
+echo "MEMKIND_PATH=${MEMKIND_PATH}"
```
crtrott: Remove this

Issue #72, opened 2017-09-03
CUDA 7.5 (GCC 4.8.4) build errors


mhoemmen:
For the version of kokkos-kernels in Trilinos develop as of today, CUDA 7.5 with GCC 4.8.4 reports build errors of the following form on the file `kokkos-kernels/src/common/KokkosKernels_SparseUtils.hpp`:
```
.../Trilinos/packages/kokkos-kernels/src/common/KokkosKernels_SparseUtils.hpp: In function ‘void KokkosKernels::Impl::kk_sort_by_row_size_parallel(lno_t, const size_type*, lno_t*, int, int)’:
.../Trilinos/packages/kokkos-kernels/src/common/KokkosKernels_SparseUtils.hpp:1137:172: error: default arguments are only permitted for function parameters
   Kokkos::parallel_for( my_exec_space(0, nv),
                                                                                                                                                                            ^
```
Errors of this form appear for lines 1137, 1150, 1156, 1162, 1598, 1612, 1656, 1675, 1692, 1761, 1767, 1798, and 1829.

mhoemmen:
I'm not sure what's wrong with this code, btw.  It looks harmless.  I'm using `-std=c++11 -expt-extended-lambda` in my `CMAKE_CXX_FLAGS`, and I've set `Kokkos_ENABLE_Cuda_Lambda:BOOL=ON`.

mhoemmen:
Oddly enough, this works fine with CUDA 8.0 and GCC 5.3.0.

ibaned:
I believe I've seen issues where if a lambda is defined in the body of a function and the function has default arguments, NVCC gets really confused by this. It could be that its a bug that was fixed in CUDA 8.0. The workaround is to refactor the code so that the function containing the lambda doesn't take default arguments.

mndevec:
Thanks @ibaned , I will give that a shot.

crtrott:
That said: Trilinos officially dropped Cuda 7.5 support. KokkosKernels should still support it but not necessarily with Lambdas enabled.

crtrott:
Also: we do not test Cuda 7.5 right now. I will add it later today ...

mhoemmen:
@crtrott wrote:
> Trilinos officially dropped Cuda 7.5 support.

Not until the next minor release, according to this:

https://github.com/trilinos/Trilinos/issues/1278

mndevec:
@mhoemmen 
I have not yet been able to replicate this issue. 

[...]

PR #85, opened 2017-09-18
Protecting lambda functions, in case cuda is enabled.

mndevec:
Should solve #72

### DIFF
--- a/perf_test/graph/KokkosGraph_triangle.cpp
+++ b/perf_test/graph/KokkosGraph_triangle.cpp
@@ -333,7 +333,7 @@ int main (int argc, char ** argv){
 
   std::cout << "Sizeof(idx):" << sizeof(idx) << " sizeof(size_type):" << sizeof(size_type) << std::endl;
 
-#if !defined (KOKKOS_HAVE_CUDA)
+#if !defined (KOKKOS_ENABLE_CUDA)
 #if defined( KOKKOS_HAVE_OPENMP )
 
   if (params.use_openmp) {
@@ -358,7 +358,7 @@ int main (int argc, char ** argv){
 #endif
 
 
-#if defined( KOKKOS_HAVE_CUDA1 )
+#if defined( KOKKOS_ENABLE_CUDA1 )
   if (params.use_cuda) {
     Kokkos::HostSpace::execution_space::initialize();
     Kokkos::Cuda::initialize( Kokkos::Cuda::SelectDevice( 0 ) );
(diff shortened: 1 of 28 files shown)


src/common/KokkosKernels_SparseUtils.hpp
```diff
@@ -1582,6 +1581,7 @@ void kk_create_incidence_tranpose_matrix_from_lower_triangle(
     bool use_dynamic_scheduling = false,
     bool chunksize = 4){
 
+#ifndef KOKKOS_HAVE_CUDA
```
crtrott: Rename to KOKKOS_ENABLE_CUDA

mndevec:
@mhoemmen 
It is basically disabling the code that is written with lambdas when Cuda is enabled. This code is not used in any downstream.

mndevec:
I have changed the use of KOKKOS_HAVE_CUDA to KOKKOS_ENABLE_CUDA all over the package. 
We will need to the same of OpenMP, Serial, and Pthreads. 
@srajama1 @kyungjoo-kim  @ambrad

perf_test/graph/KokkosGraph_triangle.cpp
```diff
@@ -358,7 +358,7 @@ int main (int argc, char ** argv){
 #endif
 
 
-#if defined( KOKKOS_HAVE_CUDA1 )
+#if defined( KOKKOS_ENABLE_CUDA1 )
```
mhoemmen: What's "CUDA1"?  Did you mean to type `KOKKOS_ENABLE_CUDA` instead?

mhoemmen:
@mndevec Argh, I should have looked harder: what is `KOKKOS_ENABLE_CUDA1`?  Did you mean to type `KOKKOS_ENABLE_CUDA` instead?

perf_test/graph/KokkosGraph_triangle.cpp
crtrott: Jup thats a typo we overlooked ...

mndevec:
@mhoemmen 
No this code is disabled in purpose. I need to make couple more optimizations to run the code on GPUs. 
For example, it is currently using parallel GNU sort.

perf_test/graph/KokkosGraph_triangle.cpp
mndevec: It is not a typo, it is intentional to disable the code.

PR #103, opened 2017-10-17
kokkos-kernels: Fix #101, and add unit test

### LINKED ISSUE #101: Sparse mat-vec of double vectors with float matrix is done in float precision
Sparse matrix-vector multiply, with double-precision vectors and a single-precision matrix, is (incorrectly) done in single precision.  @vbrunini discovered this and has written a patch that fixes the issue.  See e.g.,

https://github.com/kokkos/kokkos-kernels/blob/cd6c5f8b0d48a76350f114a13a13bed3a45ac31a/src/sparse/impl/KokkosSparse_spmv_impl.hpp#L197

The problem is that `value_type` is the type of the matrix's entries.  The temporary row sums should instead be stored as the type of the output vector's entries.

mhoemmen:
This commit fixes #101.  Thanks to @vbrunini for the patch.  I modified it slightly.  I also added a unit test.  It failed before applying Victor's patch, thus verifying the issue.  After applying the patch, the test passes.

The issue came about because someone added a new sparse matrix-vector multiply implementation for the single-vector case, which used the matrix's value type rather than the output vector's value type for intermediate row sums.  The unit test exercises both the single-vector and multiple-vector cases.


crtrott:
Thanks for putting this in, but we need to rework how the test is implemented. The test should go into a header file, and utilize the macros TestCategory and TestExecSpace for instantiation. Then we add cpp files for each execution space which simply includes that header file.

mhoemmen:
> The test should go into a header file, and utilize the macros TestCategory and TestExecSpace for instantiation. Then we add cpp files for each execution space which simply includes that header file.

I would really _really_ like the test to stay for now, and be reworked later when someone has time.  The test exercises the default execution space, and the default host execution space as well (if CUDA is enabled).  Furthermore, this issue blocks Sierra / Trilinos integration.  I fixed a bug; the test alone without the patch fails, thus manifesting the bug; the test after the patch passes, showing that the patch fixes the bug.

mndevec:
Is this a test to be run only when ETI only is turned off?

mhoemmen:
@prwolfe -- @mndevec and @crtrott just explained to me that while this test builds fine in Trilinos, kokkos-kernels has an option that makes this test not build correctly.  For full generality, the test (not necessarily the library!) needs to instantiate the kernels for `Scalar = float`.  @crtrott explained how to do this.

mhoemmen:
@prwolfe As a temporary measure, I could move the test to Tpetra for now, and move it back once we figure out how to make it build with the ETI only option enabled.

crtrott:
OK I will take care of this this week.

mhoemmen:
@crtrott No rush -- if you're too busy, I can move this test to Tpetra temporarily!

prwolfe:
Thanks guys! I meant to look at this today and got sidetracked, so I really appreciate it.

Paul

mhoemmen:
Thanks @crtrott ! :-D

mhoemmen:
@prwolfe FYI, the fix was already in Trilinos.  The issue was that the test, while building correctly for Trilinos' configuration, needed some tweaking for a kokkos-kernels configuration that Trilinos doesn't currently exercise.

PR #121, opened 2017-11-10
Update modules avail for white in test_all_sandia script

ndellingwood:
xl 14.1.0-BETA is no longer available on White for testings. Modified script to test xl 13.1.6-BETA instead.


srajama1:
@crtrott @nmhamster We are changing to test w/ XL 13.1.6-BETA instead of  14.1.0-BETA which is not there. Hopefully this is ok. @

nmhamster:
@srajama1 - please use 13.1.6, despite naming, this is actually the most up to date XL compiler.

Issue #125, opened 2017-11-29
Add GEMM default Kernel


crtrott:
The kernel should be able to handle arbitrary layouts as well as scalar types.
It also should have ok performance.

mhoemmen:
Q: What should users do if the fall-back is host only?

crtrott:
This kernel works on everything. And the point is that it is a fallback for not having the right type of TPL. It does not work with AD types which use internal parallelization though.

mhoemmen:
Stokhos deals with this in its Belos adapters by calling GEMM on the underlying entry type (e.g., `double`).  Tpetra may still want to _compile_ those functions for Stokhos, etc. types, though.  If we get compilation errors, Tpetra can always work around by wrapping and specializing out for unsupported Scalar types.

PR #126, opened 2017-11-29
Issue 125

crtrott:


src/blas/impl/KokkosBlas3_gemm_impl.hpp
```diff
@@ -0,0 +1,506 @@
...
+  typedef LayoutAScratch type;
+};
+
+#ifdef KOKKOS_ENABLE_CUDA
+template<class LayoutA, class LayoutAScratch>
+struct impl_gemm_choose_copy_layout<Kokkos::Cuda,LayoutA,LayoutAScratch> {
+  typedef LayoutA type;
+};
+#endif
+
```
kyungjoo-kim: Just thought. Scratch space layout may or may not be important in implementing an inner kernel. A better packing might be Z mortan ordering for a locality reason.

src/blas/impl/KokkosBlas3_gemm_impl.hpp
```diff
@@ -0,0 +1,506 @@
...
+        const int idx_i = offset_i+i;
+        Kokkos::parallel_for(Kokkos::ThreadVectorRange(team,blockDim_j), [&] (const int j) {
+          const int idx_j = offset_j+j;
+          A_scr(i,j) = idx_i<A.extent(0) && idx_j<A.extent(1) ? A(idx_i,idx_j) : ATV::zero();
+        });
+      });
+    }
+  }
+};
+
```
kyungjoo-kim: A question. Do we also consider that this gemm function is also used for sacado type ? As it uses vector range here, there is no chance that AD type can use vectorization inside. So, I guess that this whole thing is only about built-in data type (with AD type, packing does not make sense and data is not even contiguous).

Can we make this clear in some documentation ?

src/blas/impl/KokkosBlas3_gemm_impl.hpp
```diff
@@ -0,0 +1,506 @@
...
+    #pragma omp simd
+    for(int B_j=0; B_j<blockB1/4; B_j++) {
+#endif
+      ScalarC C_ij0 = 0;
+      ScalarC C_ij1 = 0;
+      ScalarC C_ij2 = 0;
+      ScalarC C_ij3 = 0;
+      for(int j = 0; j < blockA1; j++) {
+        ScalarC A_ij = A(i,j);
+        C_ij0 += A_ij*B(j,B_j);
```
kyungjoo-kim: Why do we need to cast to ``ScalarC`` ? This means that the function does not allow mixed precision ?

src/blas/impl/KokkosBlas3_gemm_impl.hpp
crtrott: You would loose stride-1 access on the innermost loop, and thus screw up vectorization.

src/blas/impl/KokkosBlas3_gemm_impl.hpp
```diff
@@ -0,0 +1,506 @@
...
+
+// Compute a single A block 8 B block, also do an in-place no-additional blocking team GEMM
+template<class TeamHandle, class ViewTypeA, class ViewTypeB, class ViewTypeC>
+KOKKOS_INLINE_FUNCTION
+void impl_team_gemm_block(const TeamHandle& team, const ViewTypeC& C, const ViewTypeA& A, const ViewTypeB& B) {
+  typedef typename ViewTypeC::non_const_value_type ScalarC;
+  const int blockA0 = A.extent(0);
+  const int blockA1 = A.extent(1);
+  const int blockB1 = B.extent(1);
+  Kokkos::parallel_for(Kokkos::TeamThreadRange(team,blockA0), [&] (const int i) {
```
kyungjoo-kim: Tuning capability between team size and blockA0 size. On a GPU, team size can be determined by cuda function considering functor size and shared memory (max or optimal in their context). This means that blockA0 is derived from team size ? or team size is chosen by either the cuda suggestion or blockA0 ?

src/blas/impl/KokkosBlas3_gemm_impl.hpp
crtrott: There is many other places where this doesn't work for those types, since we are using ThreadVectorRange in a number of math kernels (for example SPMV). This is a general concern. If you use a degree of parallelism from Kokkos internally to your data type you need to potentially specialize all KokkosKernels functions.

src/blas/impl/KokkosBlas3_gemm_impl.hpp
crtrott: We do not cast to ScalarC. We accumulate in ScalarC, which I believe is the right thing to do. A and B can still be a different precision Level.

src/blas/impl/KokkosBlas3_gemm_impl.hpp
crtrott: TeamSize is chose based on blockA0, otherwise you either have idle threads, or use more scratch memory per thread.

src/blas/impl/KokkosBlas3_gemm_impl.hpp
```diff
@@ -0,0 +1,506 @@
...
+
+  void run(int team_size, int vector_length, int scr_level) {
+    scratch_level = scr_level;
+    int scratch_memory_size =
+      ViewTypeAScratch::required_allocation_size() +
+      ViewTypeBScratch::required_allocation_size() +
+      ViewTypeCScratch::required_allocation_size();
+
+    Kokkos::TeamPolicy<ExecSpace,Kokkos::LaunchBounds<384,2>> policy(num_blocks_0*num_blocks_1,team_size,vector_length);
+
```
kyungjoo-kim: How can you handle different shapes of gemm ? For instance, this parallel run would not work well when C matrix is small as parallelism comes from breaking C into pieces (I call this FixC algorithm). Maybe later other algorithms (FixA, FixB) would be beneficial but I also think that we are not saving the world. So, current impl should be okay.

src/blas/impl/KokkosBlas3_gemm_impl.hpp
kyungjoo-kim: I mean ``ScalarC A_ij = A(i,j)``

crtrott: Hm yeah, but I am not sure that it matters. If this is converted into fma anyway I would assume all three arguments will be converted to ScalarC. In some sense I feel we may actually want to explicitly cast B as well. What do you think?

kyungjoo-kim: Probably possible case is ScalarA and ScalarB are float and and ScalarC is double. Then, we compute low precision multiplication and addition (?). I agree that it is not important. 

Or other possibility may come from complex arithmetics ? A and B are real matrix and C is complex ?

src/blas/impl/KokkosBlas3_gemm_impl.hpp
crtrott: Yeah if there are two few C blocks this is not gonna be super great. Right now its broken up in chunks of 24x64. One thing we could do as a stop-gap measure is to instantiate also a 8x16 variant or so. That said on say Haswell dual socket we should be good as long as C is at least something like 200x200. On P100 about 300x300.

src/blas/impl/KokkosBlas3_gemm_impl.hpp
crtrott: The complex case is interesting, but not sure whether it is relevant. C, complex and one of A and B being complex while the other one is real, that yes but both A and B real, seems weird. The A and B float case I think it would be better to convert everything to double to use fma. I dont' think you would gain anything from not using fma, in order to do the multiplication in float and then the addition in double. It would probably drive vectorization crazy as well since the vector length of the ops doesn't match up.

src/blas/impl/KokkosBlas3_gemm_impl.hpp
mhoemmen: @kyungjoo-kim See my comment here: https://github.com/kokkos/kokkos-kernels/issues/125#issuecomment-348351565

Stokhos, etc. has its own work-arounds for Belos adapters.  Tpetra still needs to compile correctly, though it can just wrap the GEMM implementation with a wrapper than can be partially specialized not to call those functions.

src/blas/impl/KokkosBlas3_gemm_impl.hpp
mhoemmen: Sierra uses mixed-precision sparse mat-vec, with float matrix and double vectors.  

If I were focusing on correctness and maximum generality, I would actually prefer either not doing the explicit cast, or using `std::common_type`.  I would also consider an option to do intermediate sums in extended precision.  On the other hand, I'm glad that @crtrott is writing this ;-) .

PR #127, opened 2017-12-02
Importing Brian's Distance-2 Graph Coloring

william76:
@mndevec 
I've done some cleanup on the code that you and I have been putting together to pull in the distance-2 coloring code that Brian developed over on his Trilinos fork and wanted to see if we can get this in before the Tutorial next week for the muelu team to have one of the distance-2 algorithms available.

I have some places stubbed out for my code to go and have been digging around looking at the distance-1 codes to see how that one was implemented to base my initial implementation off of.

Please have a look and let me know if you'd like me to change anything?  Probably the biggest change from the last time you saw it is that I changed the enum to use `COLORING_D2_MATRIX_SQUARED` to select the M^2 + D1 coloring method and am using `COLORING_D2_WCMCLEN` to select the code I'll be developing (we'll change that name later). 

The `COLORING_D2_WCMCLEN` method will throw an error if anyone tries to use it citing that it's not implemented yet.


william76:
This PR is out of date... closing it.

PR #130, opened 2017-12-13
SIMD type

kyungjoo-kim:
This PR includes modification for SIMD type. What we want is a single SIMD type implementation that are currently used in a similar way for Sierra, Stokhos and KokkosKernels. 

Task 1: Identify the union of all SIMD functionalities and implement them.
0. SIMD type
  - removes execution space
  - ``Vector<SIMD<T>,vl>`` where ``T`` is scalar type and ``vl`` is vector length and ``SIMD`` implies the instruction type applied (maybe this is redundant but when a vector has non contiguous memory instruction type might be different)
  - ``vl`` is compile time parameter and the number of vector threads can be runtime parameter.
  - specialized with AVX and AVX512 when ``Vector<SIMD<double>,4>`` and ``Vector<SIMD<double>,8>; required AVX instruction set is AVX512F only. 
  - ``Vector<SIMD<complex<double> >, 2>`` and ``Vector<SIMD<complex<double> >, 4>`` are implemented with AVX set. One also can do ``complex<SIMD<double>,4>``.

1. Math functions
  - ``Kokkos_ArithTraits.hpp`` are enriched with scalar math functions.
  - pow, cbrt, sin, cos, tan, asin, acos, atan, sinh, cosh, tanh.
  - implemented for both real and complex. 
  - functions are overloaded with SIMD type: ``KokkosBatched_Vector_SIMD_Math.hpp``

2. Ternary operator: ``KokkosBatched_Vector_SIMD_Misc.hpp``
  - ``return_val conditional_assign(condition, true_val, false_val)`` is added. 
  - Mike wants to keep STK style name ``if_then_else``; we can discuss and change the function name. 

3. reduce operator with binary operator:  ``KokkosBatched_Vector_SIMD_Misc.hpp``
  - vector ``is_all_true, is_any_true, min, max, sum, prod`` are implemented. 

4. relation operator: ``KokkosBatched_Vector_SIMD_Relation.hpp``
  - comparison operators (<, >, <=, >=, ==, !=) returns ``Vector<SIMD<bool>, vl>``

5. logical operator: ``KokkosBatched_Vector_SIMD_Logical.hpp``
  - &&, ||, ! are implemented
  - inputs are integral types and returns ``Vector<SIMD<bool>,vl``

You can consider the following two items as suggestion (not part of this pull request) 
=============================================================

6. View inter-operatibility (this needs further discussion about functionality and Kokkos support). 
At this moment, this implementation simply reflects what I need for KokkosBatched kernels and nothing is determined as permanent APIs.
  - see ``KokkosBatched_Vector_SIMD_View.hpp`` and ``Test_Batched_VectorView.hpp``
  - assignment between SIMD view and scalar view
```
View<Vector<SIMD<double>,4>**> Av;
View<double**,PackDim<0> > As; 
As = Av; Av = As; As_flat = Av; 
```
  - this requires a new layout support from Kokkos. 
  
7. Local private types (this is also a rough idea to support hierarchical parallelism in a library).
  - see ``Test_Batched_VectorLocalPrivateType.hpp`` 
  - we want a single implementation for host and cuda architectures. 
  - however, host architecture has implicit vectorization and cuda exploits explicit vectorization.
  - this means that we can hide vectorization within SIMD type in host but on GPUs, we probably need to use bracket operator. 
  - this test example shows multiple implementation of a simple matrix matrix multiplication to find out a best practice for GPU APIs.


kyungjoo-kim:
Thank. It should go to develop.

src/Kokkos_ArithTraits.hpp
```diff
@@ -1292,18 +1533,99 @@ class ArithTraits< ::Kokkos::complex<float> > {
   static KOKKOS_FORCEINLINE_FUNCTION val_type conj (const val_type x) {
     return ::Kokkos::conj (x);
   }
-  // static KOKKOS_FORCEINLINE_FUNCTION val_type pow (const val_type x, const val_type y) {
-  //   return ::pow (x, y);
-  // }
+  static KOKKOS_FORCEINLINE_FUNCTION val_type pow (const val_type x, const val_type y) {
```
mhoemmen: I feel a bit uncomfortable with us getting into the business of transcendental function implementation.  For example, I see terms like (exp(?) - 1) in the denominator below.  What can we promise about that?  Should it always be positive?  What actually needs pow and log?  Can we skip some of these?

src/batched/KokkosBatched_Vector_SIMD_Arith.hpp
```diff
@@ -0,0 +1,725 @@
...
+
+/// \author Kyungjoo Kim (kyukim@sandia.gov)
+
+#include "Kokkos_Complex.hpp"
+
+namespace KokkosBatched {
+  namespace Experimental {
+
+    //#define SIMD_ARITH_RETURN_TYPE(A) typename std::enable_if<std::is_same<Kokkos::Impl::ActiveExecutionMemorySpace,Kokkos::HostSpace>::value,A >::type
+#define SIMD_ARITH_RETURN_TYPE(T,l) Vector< SIMD< T >, l >
```
mhoemmen: Please namespace macros with e.g., `KOKKOS_KERNELS_`.

src/batched/KokkosBatched_Vector_SIMD_Logical.hpp
```diff
@@ -0,0 +1,135 @@
...
+#define __KOKKOSBATCHED_VECTOR_SIMD_LOGICAL_HPP__
+
+/// \author Kyungjoo Kim (kyukim@sandia.gov)
+
+#include "Kokkos_Complex.hpp"
+
+namespace KokkosBatched {
+  namespace Experimental {
+
+#define SIMD_LOGICAL_RETURN_BOOL_TYPE(T0,T1,l) typename std::enable_if<std::is_integral< T0 >::value && std::is_integral< T1 >::value,const Vector<SIMD<bool>,l> >::type
```
mhoemmen: Please namespace macros with e.g., `KOKKOS_KERNELS_`.

mndevec:
Does this need to go before the master promotion this week?

src/Kokkos_ArithTraits.hpp
mhoemmen: I'm OK with this, if we separate the new transcendental functions into a separate commit that comes after the SIMD functionality.  I do worry about getting into this business, though I realize that the other SIMD libraries do it.

kyungjoo-kim: Most math functions of ``Kokkos::complex`` naively implement mathematical definition. It uses ``mag_type`` arithmetic operations and verified wtih the results computed from ``std::complex``.  I may not care for all edge cases for now. 

Probably not many people actually need complex math functions. We can skip those hand-made functions for complex values. 

Any objection from others @mrtupek @etphipps @crtrott ?

crtrott:
No we will not put this into the master promotion. It is fragile enough as it is now. We have multiple issues to resolve, even after fixing a number of things already.

src/batched/KokkosBatched_Vector_SIMD_Arith.hpp
kyungjoo-kim: I got it.

src/batched/KokkosBatched_Vector_SIMD_Logical.hpp
kyungjoo-kim: will change all.

kyungjoo-kim:
@mhoemmen The reason I do not ``Vector<double,4> is because my template knowledge is not good enough to do so. I will figure out a way to do that. 

@mndevec @crtrott This cannot go to master promotion this time. Trilinos tridiagonal factorization uses KokkosKernels batch APIs and syntax would be changed and the code is not yet ready to adopt the changes.

mhoemmen:
@kyungjoo-kim wrote:
> The reason I do not ``Vector<double,4> is because my template knowledge is not good enough to do so. I will figure out a way to do that.

See Tpetra/Redesign2017/alias-demo-2.cpp for an example.  Thanks!

@micahahoward @tcfisher 

Regarding View interoperability (6): 
  - It would be super handy to be able to assign between SIMD and non-SIMD Views.
  - It would be nice if there were some function mapping from the SIMD View, to the layout that the corresponding non-SIMD View needs in order to do the assignment.  Otherwise, how do I deduce this, especially since there was a proposal to have non-SIMD Vector types with possibly noncontiguous layouts?
  - Why not use Kokkos' existing Tiled layout infrastructure to handle the layout for the non-SIMD Views?

kyungjoo-kim:
@mhoemmen 

non-contiguous layout is a bit hypothetical. I mentioned that because Sacado already uses dynamic FAD type which may have non contiguous data layout internally.

Regarding to the tile layout, that is what I first asked when SIMD view comes up. Tile layout is more specialized with 2D tile (not 1D tile) and it seems not to be maintained well. 
I prefer to use 1D tile layout instead of developing another SIMD layout although I am not sure if there is a backward compatibity issue.

src/Kokkos_ArithTraits.hpp
kyungjoo-kim: Okay. Let's postpone this later. Complex functions are not immediately used in any application. I commented them out.

PR #131, opened 2017-12-14
Fix compilation with complex<float> by adding some static casts to scalar constants

mhoemmen:
(Oops, I did this as a commit on my fork's develop branch.)

Author:    Jonas Thies <jonas.thies@dlr.de>
Editor:    Mark Hoemmen <mhoemme@sandia.gov>

Editor's note: Both kokkos and kokkos-kernels have their own
repositories.  They are periodically snapshotted into the Trilinos
repository.  Jonas submitted his original patch to Trilinos.  In order
to turn this into pull requests for kokkos and kokkos-kernels, I had
to split the original patch into two patches, and edit paths in each.
I also changed the original title slightly.  Otherwise, I made no
other changes to the original patch.  Thanks Jonas! :-D


mhoemmen:
Actually never mind, let me fix my develop branch first.

PR #132, opened 2017-12-14
Fix compilation with complex<float> by adding some static casts

mhoemmen:
Original title: "Kokkos: Fix compilation with complex<float>
by adding some static casts to scalar constants"

Author:    Jonas Thies <jonas.thies@dlr.de>
Editor:    Mark Hoemmen <mhoemme@sandia.gov>

Editor's note: Both kokkos and kokkos-kernels have their own
repositories.  They are periodically snapshotted into the Trilinos
repository.  Jonas submitted his original patch to Trilinos.  In order
to turn this into pull requests for kokkos and kokkos-kernels, I had
to split the original patch into two patches, and edit paths in each.
I also changed the original title slightly.  Otherwise, I made no
other changes to the original patch.  Thanks Jonas! :-D

### DIFF
--- a/src/batched/KokkosBatched_InnerTrsm_Serial_Impl.hpp
+++ b/src/batched/KokkosBatched_InnerTrsm_Serial_Impl.hpp
@@ -357,11 +357,11 @@ namespace KokkosBatched {
       //   a_44 = A[4*_as0+4*_as1];
 
       const ValueType           
-	inv_a_00 = 1.0/A[0*_as0+0*_as1],
-	inv_a_11 = 1.0/A[1*_as0+1*_as1],
-	inv_a_22 = 1.0/A[2*_as0+2*_as1],
-	inv_a_33 = 1.0/A[3*_as0+3*_as1],
-	inv_a_44 = 1.0/A[4*_as0+4*_as1];
+	inv_a_00 = static_cast<ValueType>(1.0)/A[0*_as0+0*_as1],
+	inv_a_11 = static_cast<ValueType>(1.0)/A[1*_as0+1*_as1],
+	inv_a_22 = static_cast<ValueType>(1.0)/A[2*_as0+2*_as1],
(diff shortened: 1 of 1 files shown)


kyungjoo-kim:
Question. The code is written assuming that the input value type is either scalar (real, complex<real>) and vector (vector<real>, vector<complex<real> >). Does ``static_cast<ValueType>(1.0)`` behave in the same way as ``ValueType(1.0)`` where ValueType is vector type?

mhoemmen:
@kyungjoo-kim Yes.

http://en.cppreference.com/w/cpp/language/static_cast

> If there is an implicit conversion sequence from expression to `new_type`, or if overload resolution for a direct initialization of an object or reference of type `new_type` from expression would find at least one viable function, then `static_cast<new_type>(expression)` returns the imaginary variable `Temp` initialized as if by `new_type Temp(expression);`, which may involve implicit conversions, a call to the constructor of `new_type` or a call to a user-defined conversion operator.

kyungjoo-kim:
I just saw your response. Do we have a smart way to tell that 1.0 is real type ? real/complex artithmetic is cheaper than complex/complex. the arithmetic operator complex/complex (with imaginery is zero) may need to be distinct in the complex operations. Just thought. Good to go. ~

crtrott:
why not use traits::one ?

kyungjoo-kim:
traits::one still returns a complex<real> type.

crtrott:
I think the issue is that we don't have a real/complex operator defined. That would probably also fix it ...

kyungjoo-kim:
We have definition for real/complex.  e.g.,
[https://github.com/kokkos/kokkos/blob/62e760fa7d3b9e65ac170c8ca1ef5c1b39c8e818/core/src/Kokkos_Complex.hpp#L357](url)

I just think that this issue is not performance critical.

mhoemmen:
@jthies @crtrott @kyungjoo-kim Thanks! :-D

Issue #148, opened 2018-02-09
Document & check CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS=ON in CUDA build


mhoemmen:
kokkos-kernels requires setting the CMake option `CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS=ON` with CUDA builds through TriBITS.  See e.g., https://github.com/trilinos/Trilinos/issues/2221 and https://github.com/trilinos/Trilinos/issues/2115.  kokkos-kernels should document this requirement, and its CMake logic should raise an error at configure time if this option is not set.

mhoemmen:
@rrdrake @prwolfe This could pose issues for Sierra's CUDA builds, due to Sierra's use of the Bjam build system.

bartlettroscoe:
I think this issue is now resolved.  See trilinos/Trilinos#2115.

mhoemmen:
@bartlettroscoe "Document" means "in kokkos-kernels' documentation."  "Check" means "in the CMake logic, such that if users don't specify the flag, the configuration process stops."

ndellingwood:
@mhoemmen @srajama1 added PR #149 with updated CMakeLists.txt to check if `CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS=ON` is set when Cuda is enabled; if not a message with FATAL_ERROR occurs at configuration indicating this should have been set.

mhoemmen:
@ndellingwood I posted my review there; thanks!

PR #149, opened 2018-02-13
CMakeLists update for issue 148

ndellingwood:
Address request in issue #148 to check for
CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS=ON when Cuda is enabled. 
If it is not on, error out at configuration.


ndellingwood:
Tested that this 1) errors out at configuration with Cuda8 if `CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS` is not set on and 2) that it configures properly when the variable is set.

CMakeLists.txt
```diff
@@ -75,6 +75,10 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
     LIST(APPEND DEVICE_LIST "<Cuda,CudaUVMSpace>")
   ENDIF()
 
+  IF(NOT CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS)
```
mhoemmen:
```
IF(NOT (DEFINED(CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS)) OR (NOT CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS))
```
etc.

srajama1:
Is there a problem with always having this ON when CUDA is enabled ? I don't know why we would error out, force the user to set it and reconfigure again. If it is needed can we just set it ON for CUDA.

CMakeLists.txt
mhoemmen: Not sure if `CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS` is defined by default.  (CMake options have three possible states: `NOT DEFINED`, `FALSE`, and `TRUE`.)  The above will check for that.

CMakeLists.txt
```diff
@@ -75,6 +75,10 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
...
+    MESSAGE( FATAL_ERROR "Kokkos-Kernels configuration failed: CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS is required be set to ON when compiling with Cuda." )
```
mhoemmen: Incorrect grammar; try this:
```
The CMake option CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS is either undefined or OFF.  Please set CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS:BOOL=ON when building with CUDA.
```

CMakeLists.txt
ndellingwood: @mhoemmen I didn't realize that, thanks for the catch!

ndellingwood:
Updated PR with suggested changes

ndellingwood:
@bartlettroscoe I added a check in kokkos-kernels for `CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS` when Cuda is enabled; should this check be moved within Trilinos and, if Cuda and kokkos-kernels are enabled, turn `CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS=ON` by default so users don't have to reconfigure like @srajama1 mentioned?

CMakeLists.txt
ndellingwood: @mhoemmen made the change, thanks!

bartlettroscoe:
> @bartlettroscoe I added a check in kokkos-kernels for CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS when Cuda is enabled; should this check be moved within Trilinos and, if Cuda and kokkos-kernels are enabled, turn CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS=ON by default so users don't have to reconfigure like @srajama1 mentioned?

This file gets processed as part of the Trilinos CMake configure as well so it should be active there as well.  In fact, any TriBITS configure of this package will have this check.

mhoemmen:
Thanks @ndellingwood ! :D

ndellingwood:
Thanks @mhoemmen!
1. I don't believe the Cuda build with complex OFF requires this, I can update so the message and check is more specific. 
2. I have not tested against all supported host compilers, though from the Tribits documentation the possible issues involving use of `CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS` seem to stem from combination of OS + CMake version + compiler. I'm not sure how to adequately address this, as testing of host compilers from sems modules on a Linux OS may give false confidence that a compiler is compatible, whereas another combo of CMake + OS may fail.

mhoemmen:
@ndellingwood wrote:
> I don't believe the Cuda build with complex OFF requires this, I can update so the message and check is more specific.

I'm not too worried about this bit, as long as the build settings work in both complex ON and OFF cases, but it's still helpful to know, so thanks :-) .

> I'm not sure how to adequately address this, as testing of host compilers from sems modules on a Linux OS may give false confidence that a compiler is compatible, whereas another combo of CMake + OS may fail.

If we don't test it, it doesn't exist ;-) .  I would say, let's discover this rather than trying to guess.  If somebody reports on a new platform that it doesn't work, we can change the message then.

ndellingwood:
@crtrott requested the `FATAL_ERROR` be changed to `WARNING` in `CMakeLists.txt`
Updated the PR with improved message and more restricted check (Cuda and complex double enabled) as suggested, and changed replaced with `WARNING`. 
Since the `FATAL_ERROR` was replaced by a `WARNING`  no one will be forced to use this option which potentially may not work with some combo of host compiler + OS + CMake version. 

@srajama1 since `CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS` will not be enabled by default (discussion and links in issue #176), I think this PR is ready.

ndellingwood:
@srajama1 README updated with comment for usage `CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS` when building with Trilinos.

CMakeLists.txt
```diff
@@ -75,6 +75,10 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
     LIST(APPEND DEVICE_LIST "<Cuda,CudaUVMSpace>")
   ENDIF()
 
+  IF( Trilinos_ENABLE_COMPLEX_DOUBLE AND ((NOT DEFINED CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS) OR (NOT CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS)) )
```
mhoemmen: This is fine, though it may be possible to turn OFF `KokkosKernels_INST_COMPLEX_DOUBLE` even if `Trilinos_ENABLE_COMPLEX_DOUBLE` is ON.

ndellingwood: @mhoemmen Thanks for the catch! I'm guessing most users use `Trilinos_ENABLE_COMPLEX_DOUBLE` so I think I should leave that and add a check for  KokkosKernels_INST_COMPLEX_DOUBLE also, does this sound correct?

mhoemmen: @ndellingwood What about stand-alone builds?  Does `Trilinos_ENABLE_COMPLEX_DOUBLE` even exist in that case?

ndellingwood: @mhoemmen I'm not sure that kokkos-kernels has standalone CMake support, I think for standalone builds the Makefile system should be used but I may be wrong.

PR #153, opened 2018-02-15
fixing sign error in distance-2 graph coloring

mndevec:


mndevec:
@ndellingwood , I was write about to push this before I leave work Today. 
This should avoid static casting.

PR #159, opened 2018-02-28
atomic_fetch_add: explicit casts needed for '1' increments

ndellingwood:
Required when compiling with serial-only backed to avoid compilation
errors due to unmatched types.

decltype used to determine the type based on the input to
atomic_fetch_add; std::remove_reference needed because expressions may
return a reference type.

### DIFF
--- a/src/common/KokkosKernels_HashmapAccumulator.hpp
+++ b/src/common/KokkosKernels_HashmapAccumulator.hpp
@@ -737,7 +737,7 @@ struct HashmapAccumulator{
       (*used_size_) += num_writes;
     });
      */
-    size_type my_write_index = Kokkos::atomic_fetch_add(used_size_, 1);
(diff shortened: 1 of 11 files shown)


mndevec:
@ndellingwood 
I though this was going to be fixed in kokkos side. Has that decision changed?

ndellingwood:
@mndevec the explicit casting for '1's in `atomic_fetch_add` is necessary after all following modification to the serial implementation for atomics in Kokkos. 

I used `decltype` in this PR to get the type to remove possible errors in guesswork of what the type should be. 

The functionality of this PR needed for the kokkos and kokkos-kernels promotion.

ndellingwood:
@mndevec I had thought this was going to be changed in kokkos also, but seems not after all. The performance fixes are in the kokkos develop branch for serial atomics though.

ndellingwood:
@mndevec I tested this PR locally with the Serial-only backend enabled, unit tests build and pass.

PR #160, opened 2018-03-01
Batched blas enhancement on gpus

kyungjoo-kim:
This PR is to complete Q2 delivery items before kokkos/kokkoskernels get merged int Trilinos. Modifications are only limited in kokkoskernels. Since ifpack2 uses batch operations, I will submit another PR in Trilinos to guard this modifications during promoting. 

Added features:
- remove AVX tags and unify with SIMD only
- reduce the number of subviewing and use assign_data
- performance test on gpus
- relation, logical, math operators enhancement

I now submit this PR as it compiles all and pass all unit tests. A block crs on a gpu has a runtime error and I am fixing it now. 

A few questions:
- How can I interface MKL and CUBLAS ? In my performance test, I compare perforamance and accuracy against those vendor libraries. Is there a way that I can figure out that with a macro ? 
- Kokkos deprecated codes; I tried to remove deprecated APIs. One problem is that I still use a "relaxed" operator(). I remembered that a special accessor function is going to be provided with removing flexibility in operator(). Do we have that ?


mhoemmen:
@kyungjoo-kim wrote:
> How can I interface MKL and CUBLAS ?

What do you mean?  Do you mean, "How can I make MKL and cuBLAS work together?", "How do I call MKL or cuBLAS?", or something else?

kyungjoo-kim:
@mhoemmen Oh sorry... What I wrote sounds weird. I am putting a performance test on the kokkoskernels and I want to compare mkl and cublas. So, I was looking for a macro to guard mkl and cublas code. In the configuration (makefile), --with-tpls=mkl or --with-tpls=cublas defines KOKKOSKERNELS_ENABLE_TPL_MKL AND KOKKOSKERNELS_ENABLE_TPL_CUBLAS in KokkosKernels_config.h. 

This was what I looked for.

kyungjoo-kim:
@ndellingwood @crtrott This PR is good to go and tested.

mhoemmen:
@kyungjoo-kim ah I see, no worries then :-)

PR #161, opened 2018-03-01
Fix makefile issue

crtrott:


crtrott:
Using admin rights to merge immediatly we need integration testing going.

PR #193, opened 2018-03-25
Fix out-of-bounds and timeout with bounds checking

ibaned:
This is a backport of the changes proposed in trilinos/Trilinos#2452 .

There was an off-by-one error when constructing subviews to supply as begin/end arrays. Also, the `dgemm_double` test is very slow with bounds checking on, so until it is modified such that it runs faster, I've added code to disable it when bounds checking is on.


mndevec:
thanks for catching and fixing these @ibaned !

ibaned:
@mndevec please don't merge yet, I have a slight change from @bartlettroscoe

ibaned:
Okay, I just pushed that extra change.

PR #197, opened 2018-03-28
D2 Color: Fix memory error on GPUs

william76:
@mndevec 

This pull request addresses the memory error fix that we worked on today for the Distance-2 Graph Coloring code.

### DIFF
--- a/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
+++ b/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
@@ -849,12 +849,17 @@ private:
       nnz_lno_t vid      = _vertexList(vid_);
       color_t   my_color = _colors(vid);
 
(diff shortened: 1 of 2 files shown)


src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
```diff
@@ -849,13 +849,18 @@ class GraphColorD2
       nnz_lno_t vid      = _vertexList(vid_);
       color_t   my_color = _colors(vid);
 
+      // If my_color is 0 then we should bail here b/c vertex was never colored.
+      if(my_color==0) return;
```
mndevec: This case should never be hit. Even when my_color is 0, I think  the correct thing is to place it in the recolorlist. 
Simply, we need to remove this line.

Issue #150, opened 2018-02-14
New Distance-2 Coloring memory access issues with Cuda Backend


mndevec:
The new distance-2 coloring runs well with OpenMP backend. However, there are some memory issues with Cuda backend that needs to be fixed.

william76:
@mndevec 
I don't know how to get Kokkos-Kernels to build on the testbed machines currently, can you send me some information and/or scripts that can get me going on one of the systems?

Also, "there are some memory access issues that need to be fixed" is really vague... It may become obvious once I can actually run on a machine with GPUs, but any details you can provide on what you're seeing or any pointers to what to look for would be tremendously helpful.

mhoemmen:
@trilinos/tpetra

mndevec:
@william76 
Below are the content of the files I use for compileKernels.sh for white and bowman. I did not look into the problems on GPUs, but you should see it when you run the unit test.

On d2 graph coloring unit test, there are codes as below that disables d2 test on gpus. If you comment out these lines you can replicate the error. 

```c++
  KokkosKernels::Impl::ExecSpaceType my_exec_space = KokkosKernels::Impl::kk_get_exec_space_type<typename device::execution_space>();
  if (my_exec_space == KokkosKernels::Impl::Exec_CUDA){
    num_algorithms = 1; //for cuda dont run d2 for now.
  }
```
The error I am getting is as below:
```
./KokkosKernels_UnitTest_Cuda --gtest_filter=*color*
Note: Google Test filter = *color*
[==========] Running 4 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 4 tests from cuda
[ RUN      ] cuda.graph_graph_color_double_int_int_TestExecSpace
[       OK ] cuda.graph_graph_color_double_int_int_TestExecSpace (536 ms)
[ RUN      ] cuda.graph_graph_color_double_int_size_t_TestExecSpace
[       OK ] cuda.graph_graph_color_double_int_size_t_TestExecSpace (543 ms)
[ RUN      ] cuda.graph_graph_color_d2_double_int_int_TestExecSpace
unknown file: Failure
C++ exception with description "cudaDeviceSynchronize() error( cudaErrorIllegalAddress): an illegal memory access was encountered /ascldap/users/mndevec/work/kokkos/core/src/Cuda/Kokkos_Cuda_Impl.cpp:119
Traceback functionality not available
" thrown in the test body.
[...]

PR #198, opened 2018-03-28
D2 Graph Coloring: Asymmetric (DAG) Transpose Fix

william76:
@mndevec 

This PR addresses the transpose issue we discussed in our meeting.  Per our conversation, I changed the inner loop (neighbor of neighbors) to use the transpose of the adjacency matrix (i.e., reverse the edges in the DAG).

There are 3 places in the code that this change was made:
1. [functorFindConflicts_Atomic][1]
2. [functorGreedyColor][2]
3. [resolveConflicts][3] 
    - We didn't discuss (3) in our conversation but I changed it to match the pattern we're using in (1) and (2) for consistency.  

**Note**: This PR should be looked AFTER PR #197 since I branched this from the mem-fix branch... or just merge this PR and ignore #197 since this one contains the changes in that PR as well.  I split them in case you want to grab the mem error bug fix quickly but wanted to look more closely at the transpose operation.

@csiefer2 FYI since you're an interested party...

[1]: https://github.com/william76/kokkos-kernels/blob/6c5b51787343f551b7167acc0165bd93c7295314/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp#L911
[2]: https://github.com/william76/kokkos-kernels/blob/6c5b51787343f551b7167acc0165bd93c7295314/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp#L821
[3]: https://github.com/william76/kokkos-kernels/blob/6c5b51787343f551b7167acc0165bd93c7295314/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp#L654

### DIFF
--- a/perf_test/graph/KokkosGraph_color_d2.cpp
+++ b/perf_test/graph/KokkosGraph_color_d2.cpp
@@ -331,7 +331,6 @@ void run_multi_mem_experiment(Parameters params)
   slow_graph_t a_slow_crsgraph, /*b_slow_crsgraph,*/ c_slow_crsgraph;
   fast_graph_t a_fast_crsgraph, /*b_fast_crsgraph,*/ c_fast_crsgraph;
 
-
   //read a and b matrices and store them on slow or fast memory.
   if (params.a_mem_space == 1)
   {
@@ -348,7 +347,6 @@ void run_multi_mem_experiment(Parameters params)
     a_slow_crsgraph = a_slow_crsmat.graph;
     a_slow_crsgraph.num_cols = a_slow_crsmat.numCols();
   }
-
 
   if (params.a_mem_space == 1)
   {
@@ -539,8 +537,8 @@ int main (int argc, char ** argv)
   std::cout << "Sizeof(kk_lno_t) : " << sizeof(kk_lno_t) << std::endl 
             << "Sizeof(size_type): " << sizeof(kk_size_type) << std::endl;
 
-#if defined( KOKKOS_HAVE_OPENMP )
 
+#if defined( KOKKOS_HAVE_OPENMP )
   if (params.use_openmp) 
   {
     Kokkos::OpenMP::initialize( params.use_openmp );
@@ -555,6 +553,7 @@ int main (int argc, char ** argv)
     Kokkos::OpenMP::finalize();
   }
 #endif
+
 
 #if defined( KOKKOS_ENABLE_CUDA )
   if (params.use_cuda) 
@@ -573,6 +572,7 @@ int main (int argc, char ** argv)
     Kokkos::HostSpace::execution_space::finalize();
   }
 #endif
+
 
 #if defined( KOKKOS_HAVE_SERIAL )
   if (params.use_serial) 
--- a/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
+++ b/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
@@ -849,12 +849,17 @@ private:
       nnz_lno_t vid      = _vertexList(vid_);
       color_t   my_color = _colors(vid);
 
+      // If my_color is 0 then we should bail here b/c vertex was never colored.
+      if(my_color==0) return;
+
       size_type vid_1adj     = _idx(vid);
       size_type vid_1adj_end = _idx(vid+1);
 
       for(; vid_1adj < vid_1adj_end; vid_1adj++)
       {
         nnz_lno_t vid_1idx = _adj(vid_1adj);
+
+        bool break_out = false;
 
         for(size_type vid_2adj=_idx(vid_1idx); vid_2adj < _idx(vid_1idx+1); vid_2adj++)
         { 
@@ -869,9 +874,11 @@ private:
             const nnz_lno_t k = Kokkos::atomic_fetch_add( &_recolorListLength(), atomic_incr_type(1));
             _recolorList(k) = vid;
             numConflicts += 1;
+            break_out = true;
             break;  // Can exit if vertex gets marked as a conflict.
           }
         }
+        if(break_out) break;
       }
     }
   }; // struct functorFindConflicts_Atomic (end)
--- a/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
+++ b/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
@@ -59,7 +59,7 @@
 
 // #define EBCOLORING_HIGHER_QUALITY        //suggested
 //
-namespace KokkosGraph { 
+namespace KokkosGraph {
 
 namespace Impl {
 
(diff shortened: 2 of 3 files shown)


mndevec:
Will, let's make only single pull request, as two of them with the same content is confusing. I closed the other PR.

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
```diff
@@ -849,16 +896,21 @@ class GraphColorD2
       nnz_lno_t vid      = _vertexList(vid_);
       color_t   my_color = _colors(vid);
 
+      // If my_color is 0 then we should bail here b/c vertex was never colored.
+      if(my_color==0) return;
```
mndevec: This check should be removed. 
my_color should never be 0, even if it was it should be placed to recolorList.

perf_test/graph/KokkosGraph_color_d2.cpp
```diff
@@ -539,8 +537,8 @@ int main (int argc, char ** argv)
   std::cout << "Sizeof(kk_lno_t) : " << sizeof(kk_lno_t) << std::endl 
             << "Sizeof(size_type): " << sizeof(kk_size_type) << std::endl;
 
-#if defined( KOKKOS_HAVE_OPENMP )
```
crtrott: Don't do this explicit OpenMP initialization. We are deprecating this. Call some form of Kokkos::initialize() instead.

mndevec: @william76 @crtrott 

Is there an example of running OMP execution space when Cuda is enabled?

william76: @crtrott
Happy to change it, but need to know what to change it to... 

@mndevec if we have to change it here, we'll also need to change the distance-1 coloring code too since this shares a lot with that driver, right?

william76:
@mndevec I committed the fix you requested into the branch -- should have updated the PR.

perf_test/graph/KokkosGraph_color_d2.cpp
mndevec: @william76  : yes, this needs to be changed for all perf_test. 
If you can fix it in this, I would follow your example to fix it for the rest of the tests.

mndevec:
@william76 : actually, could you also enable the unit tests for cuda? 

Can you remove the if check for below file in this pull request ?
https://github.com/kokkos/kokkos-kernels/blob/master/unit_test/graph/Test_Graph_graph_color_d2.hpp#L159

perf_test/graph/KokkosGraph_color_d2.cpp
william76: @mndevec ok... so being new to `Kokkos`, is there a place to look to find what variant of `Kokkos::initialize()` that I should go for?  I'm thinking there's probably one that goes something like:  `Kokkos::initialize(argc, argv)`  ?

Is there an API available that does the same thing as `KOKKOS_HAVE_OPENMP` ?

william76:
@mndevec Sure, I pulled out the if-check you referenced

perf_test/graph/KokkosGraph_color_d2.cpp
crtrott: You don't need KOKKOS_HAVE_OPENMP (which btw. is now KOKKOS_ENABLE_OPENMP).
For initialization syntax look here:
https://github.com/kokkos/kokkos/wiki/Initialization

william76: @mndevec
@crtrott

I'm still confused on this... the main() in [KokkosGraph_color_d2.cpp][1] has several sections that look like:

```
#if defined( KOKKOS_HAVE_OPENMP )
  if (params.use_openmp)
  {
    Kokkos::OpenMP::initialize( params.use_openmp );
    Kokkos::OpenMP::print_configuration(std::cout);
#ifdef KOKKOSKERNELS_MULTI_MEM
    KokkosKernels::Experiment::run_multi_mem_experiment
        <kk_size_type, kk_lno_t, Kokkos::OpenMP, Kokkos::OpenMP::memory_space, Kokkos::HostSpace>(params);
#else
    KokkosKernels::Experiment::run_multi_mem_experiment
        <kk_size_type, kk_lno_t, Kokkos::OpenMP, Kokkos::OpenMP::memory_space, Kokkos::OpenMP::memory_space>(params);
#endif
    Kokkos::OpenMP::finalize();
  }
#endif
```

so, I can put in a `Kokkos::initialize(argc, argv)` and `Kokkos::finalize()` at the end of the function but I'm not sure about the modifications needed to the driver.

I imagine `Kokkos::OpenMP::initialize()` can go away, but there's still stuff like `Kokkos::OpenMP::print_configuration()`, `Kokkos::OpenMP::memory_space`, and the OpenMP parameters that are getting passed around inside the `#if defined(KOKKOS_HAVE_OPENMP)` guard.  What do I do with those? Will they need to change? Are there analogs for them if they do?  Sadly, Kokkos doxygen pages aren't really helpful and I didn't write this file so it's not completely clear what the modifications are.  Are there any examples in Kokkos that show what the pattern should look like?

If the delta is small to fix this issue, then I think we can put it in with this PR... otherwise, I think this is outside of the scope of the purpose of this particular PR, which is to fix the CUDA bug in the D2 coloring kernel, and we should merge this PR if it's fixing the bug in the D2 coloring algorithm.  I think we should put in a separate issue ticket for the restructuring of this code and address the specific changes there.

[1]: https://github.com/kokkos/kokkos-kernels/blob/6e8e97a977564673bdadc15085d27a6a56b98114/perf_test/graph/KokkosGraph_color_d2.cpp#L542

william76: @crtrott your requests are beyond the scope of this Pull Request.  I have created an issue, #200, to track the change request and broaden the scope of that to all the tests in `perf_test/graph`.  

Assuming the updates in this PR sufficiently address the bugs in the D2 Graph Coloring implementation, do you mind merging this PR in and we can work on the other stuff off of #200?

william76:
@mndevec I don't have access to merge... are you able to click the button?

mndevec:
@william76 I don't have access either. 
@srajama1 , @crtrott  could you merge this pull request. I will handle the Christian's comment's on a separate PR.

mndevec:
#150

srajama1:
@mndevec : Went with your recommendation to merge, but #150 says it is failing, hopefully you tested this and develop is not broken. If it please revert this commit.

william76:
@srajama1 I think the comments on #150 are addressed now... it was working on my tests.  

@mndevec can you double check to make sure?  

Provided something else didn't creep in, this PR should close #150 I think.

mndevec:
@william76 , yes this should address #150. 
I will add PR "in develop" tag, which will be closed once it is promoted to master.

Issue #199, opened 2018-03-29
Generated CPP files moving includes inside the ifdef's


mndevec:
As suggested by @ibaned : many of the generated cpp files are not compiled, yet they have includes that are not protected by the ifdef macros.

For example: 
https://github.com/kokkos/kokkos-kernels/blob/master/src/impl/generated_specializations_cpp/abs/KokkosBlas1_abs_eti_spec_inst_Kokkos_complex_double__LayoutLeft_Cuda_CudaSpace.cpp

Moving "#ifdef" macros outside so that they wrap "#include ..."  might reduce the compile time further.

srajama1:
Can we change the generator too ?

crtrott:
Yeah we should change the generator. I look into it.

ndellingwood:
PR #203

PR #202, opened 2018-04-05
Moving includes inside the ifdefs in generated cpp files.

mndevec:
Related to #199.
Compilation on white with single lno, size_type and scalar_type and two execution spaces.
--previously took 3m 55s
--after changes suggested by @ibaned reduced to 3m 35 s.


mndevec:
Closing the request. I will fix the conflicts and do another PR.

PR #207, opened 2018-04-09
test_all_sandia: updates for apollo

ndellingwood:
1. change default arch from Kepler37 to Volta70
2. change order that modules load due to error with hwloc and intel
   loading hwloc prior to compiler fixed the issue; did not see
   this with other compilers.

### DIFF
--- a/scripts/test_all_sandia
+++ b/scripts/test_all_sandia
@@ -279,7 +279,7 @@ elif [ "$MACHINE" = "apollo" ]; then
 
   SKIP_HWLOC=True
 
-  BASE_MODULE_LIST="sems-env,kokkos-env,sems-<COMPILER_NAME>/<COMPILER_VERSION>,kokkos-hwloc/1.10.1/base"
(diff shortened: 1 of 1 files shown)


ndellingwood:
@crtrott for some reason with intel/17.0.1, the existing order in test_all_sandia for loading compiler then hwloc caused issues:

```
[ndellin@apollo kokkos-kernels]$ module load sems-intel/17.0.1 kokkos-hwloc/1.10.1/base
	ERROR loading module kokkos-hwloc/1.10.1/base
	Module points to path  /projects/sems/install/rhel6-x86_64/kokkos/tpl/hwloc/1.10.1/intel/17.0.1/base  which does not exist
	You may need to load the appropriate compilers before loading this module
```

Changing the order so that hwloc is loaded first fixed the issue. I did not see this when experimenting with gcc and nvcc.

ndellingwood:
@crtrott @kyungjoo-kim @srajama1 could one of you review/approve this PR? Otherwise we'll have to keep manually changing the order of the hwloc and sems compiler modules in order to test on Apollo without the module error above.

crtrott:
These changes were part of the team stuff pull request.

Issue #184, opened 2018-03-16
Implement BlockCRS in Kokkoskernels


srajama1:
We need block CRS data structure in Kokkoskernels

ndellingwood:
Initial implementation on [my fork](https://github.com/ndellingwood/kokkos-kernels/tree/feature-184)

@crtrott

ndellingwood:
PR #208 submitted.

PR #208, opened 2018-04-10
BlockCrsMatrix

ndellingwood:
Implementation of BlockCrsMatrix.
Address feature request #184

### DIFF
--- a/src/sparse/KokkosSparse_BlockCrsMatrix.hpp
+++ b/src/sparse/KokkosSparse_BlockCrsMatrix.hpp
@@ -701,11 +701,9 @@ public:
     const ordinal_type length = row_view.length; // num blocks in block-row rowi
     const ordinal_type block_size = this->blockDim();
 
-    //ordinal_type hint = 0; // Guess for offset of current column index in row
     ordinal_type numValid = 0; // number of valid local column indices
 
     for (ordinal_type i = 0; i < ncol; ++i) {
-
       // Find offset into values for block-row rowi and colidx cols[i]
       // cols[i] is the index to match
       // blk_offset is the offset for block colidx from bptr[rowi] to bptr[rowi + 1] (not global offset)
@@ -723,55 +721,11 @@ public:
             }
             else {
               local_row_values[lcol] += vals[ offset_into_vals + lrow*block_size + lcol];
-              //values_[ offset_into_values + lrow*values_row_stride + lcol] += vals[ offset_into_vals + lrow*block_size + lcol];
             }
           }
         }
         ++numValid;
       }
-
-#if 0
-      // FIXME pseudo-code thought-gathering details - remove
-      for ( ordinal_type blk_offset = 0; blk_offset < length; ++blk_offset ) {
-        ordinal_type idx = row_view.colidx_[blk_offset];
-        if ( idx == cols[i] ) {
-          //found, matched
-          ordinal_type offset_into_vals = i*block_size*block_size; //stride == 1 assumed between elements
-          ordinal_type offset_into_values = blk_offset*block_size;
-          ordinal_type values_row_stride = block_size*length; // stride to start of next row
-
-          for ( ordinal_type lrow = 0; lrow < block_size; ++lrow ) {
-            auto local_row_values = row_view.local_row_in_block(blk_offset, lrow);
-            for ( ordinal_type lcol = 0; lcol < block_size; ++lcol ) {
-              local_row_values[lcol] += vals[ offset_into_vals + lrow*block_size + lcol];
-              //values_[ offset_into_values + lrow*values_row_stride + lcol] += vals[ offset_into_vals + lrow*block_size + lcol];
-            }
-          }
-        }
-      }
-
-      // FIXME Previous implementation - remove
-      // NOTE (mfh 19 Sep 2017) This assumes that row_view stores
-      // column indices contiguously.  It does, but one could imagine
-      // changing that at some point.
-      const ordinal_type offset =
-        findRelOffset (&(row_view.colidx(0)), length, cols[i], hint, is_sorted);
-      if (offset != length) {
-        if (force_atomic) {
-          Kokkos::atomic_add (&(row_view.value(offset)), vals[i]);
-        }
-        else {
-          row_view.value(offset) += vals[i];
-        }
-        ++numValid;
-        // If the hint is out of range, findRelOffset will ignore it.
-        // Thus, while it's harmless to have a hint out of range, it
-        // may slow down searches for subsequent valid input column
-        // indices.
-        hint = offset + 1;
-      }
-#endif
-
     } // end for ncol
     return numValid;
   }
@@ -804,11 +758,9 @@ public:
     const ordinal_type length = row_view.length;
     const ordinal_type block_size = this->blockDim();
 
-    //ordinal_type hint = 0; // Guess for offset of current column index in row
     ordinal_type numValid = 0; // number of valid local column indices
 
     for (ordinal_type i = 0; i < ncol; ++i) {
-
       // Find offset into values for block-row rowi and colidx cols[i]
       // cols[i] is the index to match
       // blk_offset is the offset for block colidx from bptr[rowi] to bptr[rowi + 1] (not global offset)
@@ -826,34 +778,11 @@ public:
             }
             else {
               local_row_values[lcol] = vals[ offset_into_vals + lrow*block_size + lcol];
-              //values_[ offset_into_values + lrow*values_row_stride + lcol] = vals[ offset_into_vals + lrow*block_size + lcol];
             }
           }
         }
-      }
-
-#if 0
-      // NOTE (mfh 19 Sep 2017) This assumes that row_view stores
-      // column indices contiguously.  It does, but one could imagine
-      // changing that at some point.
-      const ordinal_type offset =
-        findRelOffset (&(row_view.colidx(0)), length, cols[i], hint, is_sorted);
-      if (offset != length) {
-        if (force_atomic) {
-          Kokkos::atomic_assign (&(row_view.value(offset)), vals[i]);
-        }
-        else {
-          row_view.value(offset) = vals[i];
-        }
         ++numValid;
-        // If the hint is out of range, findRelOffset will ignore it.
-        // Thus, while it's harmless to have a hint out of range, it
-        // may slow down searches for subsequent valid input column
-        // indices.
-        hint = offset + 1;
       }
-#endif
-
     } // end for ncol
     return numValid;
   }
--- a/src/sparse/KokkosSparse_BlockCrsMatrix.hpp
+++ b/src/sparse/KokkosSparse_BlockCrsMatrix.hpp
@@ -202,12 +202,12 @@ public:
 
 
   /// \brief Return offset into colidx_ for the requested block idx
-  ///        If none found, return std::numeric_limits::max
+  ///        If none found, return Kokkos::Details::ArithTraits::max
   /// \param idx_to_match [in] local block idx within block-row
   /// \param is_sorted [in] defaulted to false; no usage at this time
   KOKKOS_INLINE_FUNCTION
   ordinal_type findRelBlockOffset ( const ordinal_type idx_to_match, bool is_sorted = false ) const {
-    ordinal_type offset = std::numeric_limits< ordinal_type >::max();
+    ordinal_type offset = Kokkos::Details::ArithTraits< ordinal_type >::max();
     for ( ordinal_type blk_offset = 0; blk_offset < length; ++blk_offset ) {
       ordinal_type idx = colidx_[blk_offset];
       if ( idx == idx_to_match ) 
@@ -352,12 +352,12 @@ public:
 
 
   /// \brief Return offset into colidx_ for the requested block idx
-  ///        If none found, return std::numeric_limits::max
+  ///        If none found, return Kokkos::Details::ArithTraits::max
   /// \param idx_to_match [in] local block idx within block-row
   /// \param is_sorted [in] defaulted to false; no usage at this time
   KOKKOS_INLINE_FUNCTION
   ordinal_type findRelBlockOffset ( const ordinal_type &idx_to_match, bool is_sorted = false ) const {
-    typename std::remove_cv<ordinal_type>::type offset = std::numeric_limits< ordinal_type >::max();
+    typename std::remove_cv<ordinal_type>::type offset = Kokkos::Details::ArithTraits< ordinal_type >::max();
     for ( typename std::remove_cv<ordinal_type>::type blk_offset = 0; blk_offset < length; ++blk_offset ) {
       ordinal_type idx = colidx_[blk_offset];
       if ( idx == idx_to_match ) 
@@ -707,7 +707,7 @@ public:
       // blk_offset is the offset for block colidx from bptr[rowi] to bptr[rowi + 1] (not global offset)
       // colidx_ and values_ are already offset to the beginning of blockrow rowi
       auto blk_offset = row_view.findRelBlockOffset(cols[i], is_sorted);
-      if ( blk_offset != std::numeric_limits<ordinal_type>::max() ) {
+      if ( blk_offset != Kokkos::Details::ArithTraits<ordinal_type>::max() ) {
         ordinal_type offset_into_vals = i*block_size*block_size; //stride == 1 assumed between elements
         for ( ordinal_type lrow = 0; lrow < block_size; ++lrow ) {
           auto local_row_values = row_view.local_row_in_block(blk_offset, lrow); // pointer to start of specified local row within this block
@@ -762,7 +762,7 @@ public:
       // blk_offset is the offset for block colidx from bptr[rowi] to bptr[rowi + 1] (not global offset)
       // colidx_ and values_ are already offset to the beginning of blockrow rowi
       auto blk_offset = row_view.findRelBlockOffset(cols[i], is_sorted);
-      if ( blk_offset != std::numeric_limits<ordinal_type>::max() ) {
+      if ( blk_offset != Kokkos::Details::ArithTraits<ordinal_type>::max() ) {
         ordinal_type offset_into_vals = i*block_size*block_size; //stride == 1 assumed between elements
         for ( ordinal_type lrow = 0; lrow < block_size; ++lrow ) {
           auto local_row_values = row_view.local_row_in_block(blk_offset, lrow); // pointer to start of specified local row within this block
--- a/src/sparse/KokkosSparse_BlockCrsMatrix.hpp
+++ b/src/sparse/KokkosSparse_BlockCrsMatrix.hpp
@@ -358,8 +358,9 @@ public:
   /// \param is_sorted [in] defaulted to false; no usage at this time
   KOKKOS_INLINE_FUNCTION
   ordinal_type findRelBlockOffset ( const ordinal_type &idx_to_match, bool is_sorted = false ) const {
-    typename std::remove_cv<ordinal_type>::type offset = Kokkos::Details::ArithTraits< ordinal_type >::max();
-    for ( typename std::remove_cv<ordinal_type>::type blk_offset = 0; blk_offset < length; ++blk_offset ) {
+    typedef typename std::remove_cv<ordinal_type>::type non_const_ordinal_type;
+    non_const_ordinal_type offset = Kokkos::Details::ArithTraits< non_const_ordinal_type >::max();
+    for ( non_const_ordinal_type blk_offset = 0; blk_offset < length; ++blk_offset ) {
       ordinal_type idx = colidx_[blk_offset];
       if ( idx == idx_to_match ) 
       { 
--- a/src/sparse/KokkosSparse_BlockCrsMatrix.hpp
+++ b/src/sparse/KokkosSparse_BlockCrsMatrix.hpp
@@ -1,0 +1,982 @@
+/*
+//@HEADER
+// ************************************************************************
+//
+//               KokkosKernels 0.9: Linear Algebra and Graph Kernels
+//                 Copyright 2017 Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the Corporation nor the names of the
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY SANDIA CORPORATION "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL SANDIA CORPORATION OR THE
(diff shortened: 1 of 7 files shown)


src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+    return values_+(i*length*blockDim_) ;
+  }
+
+  // Return a pointer offset to local row i of block K of values_ array; 
+  //   user responsible for indexing into this pointer correctly
+  KOKKOS_INLINE_FUNCTION
+  value_type* local_row_in_block (const ordinal_type& K, const ordinal_type& i) const {
+    return (values_+(K*blockDim_ + i*length*blockDim_)) ;
+  }
+
```
kyungjoo-kim: Minor thing. We probably need to comment that all pointer based interfaces are not guaranteed for backward compatibility as this type of interface is more used in kernels for performance not for ordinary users.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+
+  KOKKOS_INLINE_FUNCTION
+  ordinal_type findRelBlockOffset ( const ordinal_type idx_to_match, bool is_sorted = false ) const {
+    ordinal_type offset = std::numeric_limits< ordinal_type >::max();
+    for ( ordinal_type blk_offset = 0; blk_offset < length; ++blk_offset ) {
+      ordinal_type idx = colidx_[blk_offset];
+      if ( idx == idx_to_match ) { offset = blk_offset; } // return relative offset
+    }
```
kyungjoo-kim: Maybe we need to early return or break when it find out the matched index ?

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+    // not once per matrix.  We need to protect this somehow.
+    cusparseCreate (&cusparse_handle);
+
+    // This is a per-matrix attribute.  It encapsulates things like
+    // whether the matrix is lower or upper triangular, etc.  Ditto
+    // for other TPLs like MKL.
+    cusparseCreateMatDescr (&cusparse_descr);
+#endif // KOKKOS_USE_CUSPARSE
+  }
+
```
kyungjoo-kim: We may need to check possibility of interfacing to cusparse with our format. As it is probably not the standard format, it may not be inter-operatible. In that case, we can just remove unncessary code guarded by KOKKOS_USE_CUSPARSE.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+  typedef OrdinalType ordinal_type;
+  typedef MemoryTraits memory_traits;
+  /// \brief Type of each entry of the "row map."
+  ///
+  /// The "row map" corresponds to the \c ptr array of row offsets in
+  /// compressed sparse row (CSR) storage.
+  typedef SizeType size_type;
+
+  //! Type of a host-memory mirror of the sparse matrix.
+  typedef BlockCrsMatrix<ScalarType, OrdinalType, host_mirror_space, MemoryTraits> HostMirror;
```
mhoemmen: Is `host_mirror_space` a memory space or a Kokkos::Device specialization? If it's a memory space, then you'll be getting its host execution space, which may not be what you want.  For example, if the device memory space is `CudaUVMSpace`, then the host memory space will also be `CudaUVMSpace`, whose execution space is `Cuda` (!).

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+  //! Type of the graph structure of the sparse matrix.
+  typedef Kokkos::StaticCrsGraph<OrdinalType, Kokkos::LayoutLeft, execution_space, SizeType> StaticCrsGraphType;
+  //! Type of column indices in the sparse matrix.
+  typedef typename StaticCrsGraphType::entries_type index_type;
+  //! Const version of the type of column indices in the sparse matrix.
```
mhoemmen: Why do you need all these typedefs?  Users can get const versions of things from the Kokkos::View typedefs directly.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+  /// \brief Type of each entry of the "row map."
+  ///
+  /// The "row map" corresponds to the \c ptr array of row offsets in
+  /// compressed sparse row (CSR) storage.
+  typedef SizeType size_type;
+
+  //! Type of a host-memory mirror of the sparse matrix.
+  typedef BlockCrsMatrix<ScalarType, OrdinalType, host_mirror_space, MemoryTraits> HostMirror;
+  //! Type of the graph structure of the sparse matrix.
+  typedef Kokkos::StaticCrsGraph<OrdinalType, Kokkos::LayoutLeft, execution_space, SizeType> StaticCrsGraphType;
```
mhoemmen: We should take this opportunity to make this typedef have consistent formatting with the other typedefs.  You could always keep this inconsistently formatted typedef around as an extra backwards compatibility measure with `KokkosSparse::CrsMatrix`, but that one really should change too.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+  //! Nonconst version of the type of row offsets in the sparse matrix.
+  typedef typename row_map_type::non_const_value_type non_const_size_type;
+  //! Kokkos Array type of the entries (values) in the sparse matrix.
+  typedef Kokkos::View<value_type*, Kokkos::LayoutRight, device_type, MemoryTraits> values_type;
+  //! Const version of the type of the entries in the sparse matrix.
+  typedef typename values_type::const_value_type const_value_type;
+  //! Nonconst version of the type of the entries in the sparse matrix.
+  typedef typename values_type::non_const_value_type non_const_value_type;
+
+#ifdef KOKKOS_USE_CUSPARSE
```
mhoemmen: Is this just a copy-and-paste from `KokkosSparse::CrsMatrix` or do you really intend to use the cuSPARSE stuff here?

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+             OrdinalType ncols,
+             size_type annz,
+             ScalarType* val,
+             OrdinalType* rows,
+             OrdinalType* cols,
+             OrdinalType blockdim,
+             bool pad = false)
+  {
+    (void) pad;
+    import (label, nrows, ncols, annz, val, rows, cols, blockdim);
```
mhoemmen: This is an unfortunate name because it collides with both Epetra and Tpetra in confusing ways.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+    }
+    // nnz returns graph.entries.extent(0) i.e. ptr[ nrows + 1 ] nnz entry
+    // input annz is nnz of values, not comparable with block ptr 'nnz' i.e. numBlocks
+    if (blockDim_ <= 0) {
+      std::ostringstream os;
+      os << "Input argument blockDim = " << blockDim 
+         << " is not larger than 0.";
+      throw std::invalid_argument (os.str ());
+    }
+
```
mhoemmen: See above comment.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+             const OrdinalType& ncols,
+             const values_type& vals,
+             const StaticCrsGraphType& graph_,
+             const OrdinalType& blockDim) :
+    graph (graph_),
+    values (vals),
+    numCols_ (ncols),
+    blockDim_ (blockDim)
+  {
+#ifdef KOKKOS_USE_CUSPARSE
```
mhoemmen: See above comment.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+    blockDim_ = blockDim;
+    numCols_ = crs_mtx.numCols() / blockDim;
+    values = crs_mtx.values;
+
+    OrdinalType nbrows = crs_mtx.numRows()/blockDim; // actual number of block rows; add 1 for ptr length
+
+    // block_rows will accumulate the number of blocks per row - this is NOT the row_map with cum sum!!
+    std::vector<OrdinalType> block_rows( nbrows, 0 );
+
+    typename crs_graph_row_map_type::HostMirror h_crs_row_map = Kokkos::create_mirror_view(crs_mtx.graph.row_map);
```
mhoemmen: Might you want to share the graph with the input object, if it's possible?

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+      OrdinalType offset_into_colidx_start = offset_into_blkcolidx_start*blockDim*blockDim;
+
+      for ( OrdinalType lidx = 0; lidx < blks_in_row; ++lidx ) {
+        h_entries( offset_into_blkcolidx_start+lidx ) = h_crs_entries( offset_into_colidx_start + blockDim*lidx ) / blockDim;
+      }
+    }
+
+    Kokkos::deep_copy (graph.entries, h_entries);
+
+#ifdef KOKKOS_USE_CUSPARSE
```
mhoemmen: See above comments on cuSPARSE handles.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+    cusparse_handle (crs_mtx.cusparse_handle),
+    cusparse_descr (crs_mtx.cusparse_descr),
+#endif // KOKKOS_USE_CUSPARSE
+  }
+
+
+  /// Declaration for import - this member function is not inlined
+  void
```
mhoemmen: It's confusing that this function is called "import"; see above comment.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+          }
+        }
+        ++numValid;
+      }
+    } // end for ncol
+    return numValid;
+  }
+
+  //! Attempt to assign the input matrix to \c *this.
+  // Are the CUDA sparse handles needed to be copied here??
```
mhoemmen: Nothing uses the cuSPARSE handles; please get rid of them.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+  /// should instead assign to 'auto'.
+  ///
+  KOKKOS_INLINE_FUNCTION
+  SparseBlockRowView<BlockCrsMatrix> block_row (const ordinal_type i) const {
+
+    const size_type start = graph.row_map(i); // total num blocks prior to this block-row
+    const ordinal_type count = static_cast<ordinal_type> (graph.row_map(i+1) - start); // num blocks in this row
+
+    if (count == 0) {
+      return SparseBlockRowView<BlockCrsMatrix> (NULL, NULL, 1, 0);
```
mhoemmen: Please use `nullptr` instead of `NULL`; thanks!

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+//----------------------------------------------------------------------------
+//----------------------------------------------------------------------------
+
+// Assumes rows is the row_map pointer for the BlockCrsMatrix graph (i.e. cum sum of number of blocks)
+//         cols is the entries pointer for the BlockCrsMatrix graph (colidx for block-row blocks)
+//         annz is the total number of non-zeros in the CrsMatrix (blockDim*blockDim*numBlocks)
+template< typename ScalarType , typename OrdinalType, class Device, class MemoryTraits, typename SizeType >
+void
+BlockCrsMatrix<ScalarType , OrdinalType, Device, MemoryTraits, SizeType >::
+import (const std::string &label,
```
mhoemmen: See above comments on the name of this method.

I don't actually think this method should exist; it should at least be a hidden implementation detail.  Users don't need this.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,940 @@
...
+        OrdinalType* cols,
+        const OrdinalType blockDim)
+{
+  std::string str = label;
+  values = values_type (str.append (".values"), annz);
+
+  numCols_ = ncols;
+  blockDim_ = blockDim;
+
+  // FIXME (09 Aug 2013) CrsArray only takes std::vector for now.
```
mhoemmen: Let's fix this.

mhoemmen:
@micahahoward and @ambrad may have some suggestions too.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: I don't see any usage or testing of this method for `CrsMatrix` in KokkosKernels
@crtrott is it safe for me to remove this from `CrsMatrix` and `BlockCrsMatrix`?

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: Yes, responded to comment below to see if this can be removed (or at least renamed and hidden).

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: Yes, responded to comment below to see if this can be removed (or at least renamed and hidden).

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: Done. I made the commenting more consistent with other parts of the code also.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: Good catch, thanks! Done.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: I removed all cusparse interface handles etc. per your and @mhoemmen's request, thanks!

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: I kept the typedefs from `KokkosSparse::CrsMatrix`. I will check to see which are not in use and clean up both files in the next wave.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: Yes, I used `KokkosSparse::CrsMatrix` as the starting point for the code and tried to keep things looking similar for familiarity. I removed all the cuSparse stuff in recent commit.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: Done :)

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: Good catch thanks! Done.

ndellingwood:
@mhoemmen @kyungjoo-kim thanks for the comments! I addressed most of them. Some of the changes that should also be done in `KokkosSparse::CrsMatrix` will come in the next commit.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: The use for the `import` method here and what is should play well with are not clear to me (this was copied along from the `KokkosSparse::CrsMatrix` and with functionality modified for the block structure of the graph). Can we discuss this more? I'd be happy to fix it :)

mhoemmen: I would say, get rid of it.  If nothing tests it, then it's broken :-) .

mhoemmen:
Thanks @ndellingwood !

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: I renamed the function to `ctor_impl` to prevent problems and confusion, will get the actual impl performance tested and cleaned up based on those results.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: I added an extra typedef consistent with the one in Kokkos: `staticcrsgraph_type`, but left this one for backward compatibility like you mentioned in case someone is using it outside of Trilinos.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: I checked in Kokkos and think this should work fine - `HostMirror` in `Kokkos::ViewTraits` takes the space (here, the `Device` template parameter) as parameter and checks that the `execution_space can` access `HostSpace` and if the `memory_space` can access `HostSpace`; depending on results, an appropriate `Device` specialization or `HostSpace` is returned.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,957 @@
...
+#include "Kokkos_Core.hpp"
+#include "Kokkos_StaticCrsGraph.hpp"
+#include <sstream>
+#include <stdexcept>
+#include <type_traits>
+#include "KokkosSparse_CrsMatrix.hpp"
+
+namespace KokkosSparse {
+
+namespace Experimental {
```
srajama1: Why is this Experimental ?  I am not saying it shouldn't be, just checking there is a proper reason.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
```diff
@@ -0,0 +1,957 @@
...
+    row_lengths[i] = rows[i + 1] - rows[i];
+  }
+
+  str = label;
+  graph = Kokkos::create_staticcrsgraph<staticcrsgraph_type> (str.append (".graph"), row_lengths);
+  typename values_type::HostMirror h_values = Kokkos::create_mirror_view (values);
+  typename index_type::HostMirror h_entries = Kokkos::create_mirror_view (graph.entries);
+
+  // FIXME (mfh 21 Jun 2013) This needs to be a parallel copy.
+  // Furthermore, why are the arrays copied twice? -- once here, to a
```
srajama1: Let us avoid moving FIXME from one to another. Let us fix it or if it is low priority just make it a comment.

srajama1:
@ndellingwood : Let us do the high priority changes requested and get this PR merged.

src/sparse/KokkosSparse_BlockCrsMatrix.hpp
ndellingwood: @mhoemmen @srajama1 I have a commit ready to push to change the internals for this `ctor_impl` routine and wanted to comment on the changes here in case I'm overlooking something.

Replace the whole impl with the following:
```
  numCols_ = ncols;
  blockDim_ = blockDim;
  typename values_type::HostMirror unman_val( val, annz );
  typename row_map_type::HostMirror unman_rows( rows, nrows+1);
  typename index_type::HostMirror unman_cols( cols, ncols );

  // Create temporary Views for row_map and entries because the StaticCrsGraph ctor requires View inputs
  values_type tmp_row_map("tmp_row_map", nrows+1);
  values_type tmp_entries("tmp_entries", ncols);

  // deep_copy uses Kokkos::MDRangePolicy which will perform the copy much more efficiently than simply doing so serially
  Kokkos::deep_copy( val, unman_val );
  Kokkos::deep_copy( tmp_row_map, unman_rows );
  Kokkos::deep_copy( tmp_entries, unman_cols );

  // Construct graph using the temp entries and row_map Views
  graph = staticcrsgraph_type( tmp_entries, tmp_row_map );
```

This will be more efficient by comparing the previous impl to the code above - 
  // Previous ctor_impl does the following:
  // 1. Compute row counts from the `rows` pointer
  // 2. Create a `StaticCrsGraph` which basically reconstructs the existing `rows` pointer (stored in `HostMirror` of the `row_map` member) by serially iterating over the row counts in host space, followed by a `deep_copy` to the `row_map` member; also allocate the `entries` View
  // 3. Create temporary `HostMirror` views for `values` and `entries` for copying
  // 4. Serially copy on the host the `val` and `cols` pointers to `values` and `entries` (resp.)
  // 5. `deep_copy` results to corresponding member Views

  // Instead, reduce unnecessary copying and computation:
  // 1. Wrap the `val`, `rows`, `cols` in unmanaged `HostMirror` Views
  // 2. Create temporary Views (managed) for `entries` and `row_map` - these will be assigned to the `BlockCrsMatrix` `graph` member via its constructor (passed by reference)
  // 3. `deep_copy` the above unmanaged HostMirror Views to corresponding Views
  // 4. Assign `row_map` and `entries` to `graph` member; when `ctor_impl` completes the ref count of these managed Views will drop to 1 rather than 2. 

This removes the unnecessary computation of the row counts only to recompute what's stored in the already-existing rows pointer entries to store in the host row_map View, and removes the serial copies from the crs pointers to corresponding host views; replaced by unmanaged view construction to wrap each pointer and the existing `deep_copy` calls remain.

srajama1: Looks good to me

ndellingwood:
@srajama1 pushed the changes for `ctor_impl` will need to run spot-check which takes awhile.

srajama1:
Apparently you/I cannot merge unless @mhoemmen approves as he requested changes.

ndellingwood:
@mhoemmen can you take a glance at the most recent comments/changes I left and let me know if it accept or need more changes?

mhoemmen:
Thanks @ndellingwood !

ndellingwood:
Updated PR, rerunning spot-check

ndellingwood:
Working on a cuda bug...

ndellingwood:
Most recent push should fix the cuda issue (need to rerun spot-check), the other compiler/backend combos had passed. Will post results after a rerun and it will be ready to merge then.

ndellingwood:
Got the unit tests and code fixed and passes a spot-check on kokkos-dev:

```
Going to test compilers:  gcc/5.3.0 intel/16.0.1 clang/3.9.0 cuda/8.0.44
Testing compiler gcc/5.3.0
Testing compiler intel/16.0.1
  Starting job gcc-5.3.0-Serial-hwloc-release
Testing compiler clang/3.9.0
  Starting job intel-16.0.1-OpenMP-hwloc-release
  PASSED gcc-5.3.0-Serial-hwloc-release
Testing compiler cuda/8.0.44
  Starting job clang-3.9.0-Pthread_Serial-hwloc-release
  PASSED intel-16.0.1-OpenMP-hwloc-release
  PASSED clang-3.9.0-Pthread_Serial-hwloc-release
PASSED TESTS
clang-3.9.0-Pthread_Serial-hwloc-release build_time=783 run_time=1106
clang-3.9.0-Pthread_Serial-release build_time=784 run_time=2413
cuda-8.0.44-Cuda_OpenMP-release build_time=1073 run_time=1080
gcc-5.3.0-Serial-hwloc-release build_time=784 run_time=1111
gcc-5.3.0-Serial-release build_time=797 run_time=1089
intel-16.0.1-OpenMP-hwloc-release build_time=398 run_time=360
intel-16.0.1-OpenMP-release build_time=991 run_time=346
FAILED TESTS
```

@srajama1 this is ready to merge :)

srajama1:
@ndellingwood : Should I click the green button ?

ndellingwood:
@srajama1 Yes, it's ready to go :)

Issue #214, opened 2018-04-23
Add team-based scal, mult, update, nrm2


vqd8a:
Adding scal(team_handle,...), mult(team_handle,...), update(team_handle,...), nrm2(team_handle,...)

Issue #209, opened 2018-04-13
Add team based abs


vqd8a:
Adding abs(team_handle,...)

mhoemmen:
@vqd8a Did you mean to write "dot, axpby, and gemv" instead of "abs" in the title?

vqd8a:
@mhoemmen No, I meant "abs". @crtrott did for dot, axpby, gemv. I will do for other blas functions.

mhoemmen:
@vqd8a I asked that question because the word "abs" does not appear anywhere in the pull request :-) .  The pull request only appears to affect dot, axpby, and gemv.

vqd8a:
@mhoemmen I see your point :-). I don't know why dot, axpby and gemv appear in the PR. Let me try creating a new PR

mhoemmen:
@vqd8a Thanks!

mhoemmen:
@vqd8a Thanks!  Another option is just to rename the PR.

PR #215, opened 2018-04-23
Adding team-based BLAS functions

vqd8a:
Addressing issues #209, #214


src/blas/KokkosBlas1_team_axpby.hpp
```diff
@@ -0,0 +1,69 @@
...
+#ifndef KOKKOSBLAS1_TEAM_AXPBY_HPP_
+#define KOKKOSBLAS1_TEAM_AXPBY_HPP_
+
+#include<KokkosBlas1_team_axpby_spec.hpp>
+
+namespace KokkosBlas {
+namespace Experimental {
+
+template<class TeamType, class XVector,class YVector>
+void KOKKOS_INLINE_FUNCTION axpby (const TeamType& team,
```
mhoemmen: Sad that the coefficients have to have the same type as the vector entries.  Ah well.

src/blas/impl/KokkosBlas1_team_axpby_spec.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+template<class TeamType, class XVector, class YVector>
+struct TeamAXPBY<TeamType, XVector, YVector, false>
+{
+  static KOKKOS_INLINE_FUNCTION void team_axpby (const TeamType& team,
+      const typename XVector::non_const_value_type& a, const XVector& x,
+      const typename YVector::non_const_value_type& b, const YVector& y) {
+    const int N = x.extent(0);
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(team,N), [&] (const int& i) {
+      y(i) = b*y(i) + a*x(i);
```
mhoemmen: This breaks BLAS semantics:
  1. If `b` is zero, the code should overwrite `y(i)` without multiplying.  This matters if `y(i)` is `Inf` or `NaN`.
  2. If `a` is zero, the code should ignore `x(i)` without multiplying.  This matters if `x(i)` is `Inf` or `NaN`.

If we intend to do this, then we should fix the behavior of the node-global version to be consistent with this.

src/blas/impl/KokkosBlas1_team_dot_spec.hpp
```diff
@@ -0,0 +1,87 @@
...
+};
+
+template<class TeamType, class XV, class YV>
+struct TeamDot<TeamType, XV, YV, false> {
+  typedef typename Kokkos::Details::InnerProductSpaceTraits<typename XV::non_const_value_type>::dot_type dot_type;
+  static KOKKOS_INLINE_FUNCTION dot_type team_dot (const TeamType& team, const XV& X, const YV& Y) {
+    dot_type result;
+    int N = X.extent(0);
+    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team,N), [&] (const int& i, dot_type& val) {
+      val += X(i)*Y(i);
```
mhoemmen: If you're using InnerProductSpaceTraits (which you should), then you should use `InnerProductSpaceTraits::dot` to compute this product, not `operator*`.

src/blas/impl/KokkosBlas1_team_mult_spec.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+template<class TeamType, class YVector, class AVector, class XVector>
+struct TeamMult<TeamType, YVector, AVector, XVector, false>
+{
+  static KOKKOS_INLINE_FUNCTION void team_mult (const TeamType& team,
+      const typename YVector::non_const_value_type& gamma, const YVector& y,
+      const typename AVector::non_const_value_type& alpha, const AVector& a, const XVector& x) {
+    const int N = x.extent(0);
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(team,N), [&] (const int& i) {
+      y(i) = gamma*y(i) + alpha*a(i)*x(i);
```
mhoemmen: See above note on BLAS semantics.

src/blas/impl/KokkosBlas1_team_scal_spec.hpp
```diff
@@ -0,0 +1,83 @@
...
+  static KOKKOS_INLINE_FUNCTION void team_scal (const TeamType& team, const RV& R, const typename XV::non_const_value_type& a, const XV& X);
+};
+
+template<class TeamType, class RV, class XV>
+struct TeamScal<TeamType, RV, XV, false>
+{
+  static KOKKOS_INLINE_FUNCTION void team_scal (const TeamType& team, const RV& R, const typename XV::non_const_value_type& a, const XV& X) {
+    const int N = X.extent(0);
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(team,N), [&] (const int& i) {
+      R(i) = a*X(i);
```
mhoemmen: See above note on BLAS semantics.  (If `a` is zero, this should fill `R` with zeros, regardless of the contents of `X`.)

src/blas/impl/KokkosBlas1_team_update_spec.hpp
```diff
@@ -0,0 +1,89 @@
...
+template<class TeamType, class XVector, class YVector, class ZVector>
+struct TeamUpdate<TeamType, XVector, YVector, ZVector, false>
+{
+  static KOKKOS_INLINE_FUNCTION void team_update (const TeamType& team,
+      const typename XVector::non_const_value_type& alpha, const XVector& x,
+      const typename YVector::non_const_value_type& beta,  const YVector& y, 
+      const typename ZVector::non_const_value_type& gamma, const ZVector& z) {
+    const int N = x.extent(0);
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(team,N), [&] (const int& i) {
+      z(i) = gamma*z(i) + alpha*x(i) + beta*y(i);
```
mhoemmen: See above note on BLAS semantics.

src/blas/impl/KokkosBlas2_team_gemv_spec.hpp
```diff
@@ -0,0 +1,142 @@
...
+      const typename YVector::non_const_value_type& beta, const YVector& y) {
+    const int N = A.extent(0);
+    const int M = A.extent(1);
+
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(team,N), [&] (const int& i) {
+      dot_type Ax_i;
+      Kokkos::parallel_reduce(Kokkos::ThreadVectorRange(team,M), [&] (const int& j, dot_type& val ) {
+        val += A(i,j) * x(j);
+      },Ax_i);
+      y(i) = beta * y(i) + alpha * Ax_i;
```
mhoemmen: See above note on BLAS semantics.

src/blas/impl/KokkosBlas2_team_gemv_spec.hpp
```diff
@@ -0,0 +1,142 @@
...
+      const XVector& x,
+      const typename YVector::non_const_value_type& beta, const YVector& y) {
+    const int N = A.extent(1);
+    const int M = A.extent(0);
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(team,N), [&] (const int& i) {
+      dot_type Ax_i;
+      Kokkos::parallel_reduce(Kokkos::ThreadVectorRange(team,M), [&] (const int& j, dot_type& val ) {
+        val += A(j,i) * x(j);
+      },Ax_i);
+      y(i) = beta * y(i) + alpha * Ax_i;
```
mhoemmen: See above note on BLAS semantics.

src/blas/impl/KokkosBlas2_team_gemv_spec.hpp
```diff
@@ -0,0 +1,142 @@
...
+      const XVector& x,
+      const typename YVector::non_const_value_type& beta, const YVector& y) {
+    const int N = A.extent(1);
+    const int M = A.extent(0);
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(team,N), [&] (const int& i) {
+      dot_type Ax_i;
+      Kokkos::parallel_reduce(Kokkos::ThreadVectorRange(team,M), [&] (const int& j, dot_type& val ) {
+        val += Kokkos::ArithTraits<value_type>::conj(A(j,i)) * x(j);
+      },Ax_i);
+      y(i) = beta * y(i) + alpha * Ax_i;
```
mhoemmen: See above note on BLAS semantics.

unit_test/blas/Test_Blas1_team_abs.hpp
```diff
@@ -0,0 +1,292 @@
...
+    Kokkos::fill_random(b_y,rand_pool,ScalarB(10));
+
+    Kokkos::deep_copy(b_org_y,b_y);
+
+    Kokkos::deep_copy(h_b_x,b_x);
+    Kokkos::deep_copy(h_b_y,b_y);
+
+    ScalarA expected_result = 0;
+    for(int i=0;i<N;i++)
+      expected_result += AT::abs(h_x(i)) * AT::abs(h_x(i));
```
mhoemmen: This would be a good place to compute a Wilkinson-style running error bound, rather than using an ad-hoc bound (see below).

unit_test/blas/Test_Blas1_team_abs.hpp
```diff
@@ -0,0 +1,292 @@
...
+
+    //KokkosBlas::abs(y,x); 
+    Kokkos::parallel_for( policy, KOKKOS_LAMBDA ( const team_member &teamMember ) {
+       const int teamId = teamMember.league_rank();
+       KokkosBlas::Experimental::abs(teamMember, Kokkos::subview(y,std::make_pair(teamId*team_data_siz,(teamId < M-1)?(teamId+1)*team_data_siz:N)), Kokkos::subview(x,std::make_pair(teamId*team_data_siz,(teamId < M-1)?(teamId+1)*team_data_siz:N)));
+    } );
+
+    ScalarB nonconst_nonconst_result = KokkosBlas::dot(y,y);
+    EXPECT_NEAR_KK( nonconst_nonconst_result, expected_result, eps*expected_result);
```
mhoemmen: See above note.

unit_test/blas/Test_Blas1_team_abs.hpp
```diff
@@ -0,0 +1,292 @@
...
+    Kokkos::deep_copy(b_y,b_org_y);
+
+    //KokkosBlas::abs(y,c_x);
+    Kokkos::parallel_for( policy, KOKKOS_LAMBDA ( const team_member &teamMember ) {
+       const int teamId = teamMember.league_rank();
+       KokkosBlas::Experimental::abs(teamMember, Kokkos::subview(y,std::make_pair(teamId*team_data_siz,(teamId < M-1)?(teamId+1)*team_data_siz:N)), Kokkos::subview(c_x,std::make_pair(teamId*team_data_siz,(teamId < M-1)?(teamId+1)*team_data_siz:N)));
+    } );
+
+    ScalarB const_nonconst_result = KokkosBlas::dot(y,y);
+    EXPECT_NEAR_KK( const_nonconst_result, expected_result, eps*expected_result);
```
mhoemmen: See above note.

src/blas/impl/KokkosBlas1_team_dot_spec.hpp
crtrott: When is that not the same for dot? I understand why it wouldn't be the same for NRM2 but for dot?

crtrott:
Hm interesting question regarding the blast semantics. At the top level we handle that logic pretty far outward and just do ETI for everything. Here we would need to inject all the different code variants into your kernel, which could have some bad performance effects. I definitely want to keep BLAS semantics on the global functions but for the nested ones I just don't know whether the possible performance issues are worth it. Any thoughts?

src/blas/impl/KokkosBlas1_team_dot_spec.hpp
mhoemmen: @crtrott It's for some Stokhos scalar types, where Eric wants to do Krylov methods with plain old `double` as the dot product / norm results, even though the vectors have Stokhos types in them.

This is also a reason why one might want `axpby` with coefficients of a different type than the vectors' entries.

mhoemmen:
@crtrott wrote:
> I definitely want to keep BLAS semantics on the global functions but for the nested ones I just don't know whether the possible performance issues are worth it. Any thoughts?

Tpetra can always implement BLAS semantics on its own, so I'm not personally worried so much about the global functions.  However, you do have to worry about users following current developments in the batched linear algebra standardization effort.  In addition, it depends on how we intend to use these new functions.  I'm not sure what multifrontal / supernodal sparse factorizations expect, for example.

vqd8a:
@crtrott Any further guidance?

srajama1:
@vqd8a : The BLAS semantics are important for node level calls. For the team level ones you are working on, the semantics are not clear, especially with the performance hit Christian is talking about. In fact, this is a point of discussion for the batched BLAS committee. Let me check with them next time we talk. We don't expect this semantics for our sparse solvers, so what you have done is good for us.

The use case of coefficients different than vector entries is also a node-level use case. I have never heard this for a team level use case though it could extend. I am willing to go with what you have and then iterate for the other use case.  Can you create a wiki page describing these comments so users know what they are getting ?

vqd8a:
Thanks, @srajama1 @mhoemmen.
@srajama1 I am going to create a wiki page for these. BTW, I am not quite clear about "The use case of coefficients different than vector entries is also a node-level use case". Could you please explain more?

etphipp:
For some Stokhos scalar types, the results of dot-products are different from the scalar type, hence the use of InnerProductSpaceTraits as Mark described.  However Stokhos has historically overloaded these functions for its scalar types anyways, so it probably won't really matter (unless this breaks those overloads where they are no longer overriding the right function signature).

etphipp:
And after looking at the code more closely, you really should follow Mark's suggestion.  It's a 1-line change and makes the code consistent.

vqd8a:
Thanks. @etphipp

srajama1:
@vqd8a : I was wondering if the InnerProductSpaceTraits are really needed because of the overloading @etphipp mentioned. He says it is needed, so let us do that.

vqd8a:
Changed to using InnerProductSpaceTraits::dot to compute the product in KokkosBlas1_team_dot_spec.hpp

srajama1:
As @mhoemmen and @etphipp wanted this change I will wait for them to review. I can click the green button once either of them approve.

vqd8a:
{71:81}src/blas/impl/KokkosBlas1_team_mult_spec.hpp

vqd8a:
https://github.com/kokkos/kokkos-kernels/blob/1c68159b325a4a8a2135fa73606374202521c1d7/src/blas/impl/KokkosBlas1_team_mult_spec.hpp#L77-L79

Should we also use InnerProductSpaceTraits::dot here?

https://github.com/kokkos/kokkos-kernels/blob/1c68159b325a4a8a2135fa73606374202521c1d7/src/blas/impl/KokkosBlas2_team_gemv_spec.hpp#L88-L90
https://github.com/kokkos/kokkos-kernels/blob/1c68159b325a4a8a2135fa73606374202521c1d7/src/blas/impl/KokkosBlas2_team_gemv_spec.hpp#L109-L111
https://github.com/kokkos/kokkos-kernels/blob/1c68159b325a4a8a2135fa73606374202521c1d7/src/blas/impl/KokkosBlas2_team_gemv_spec.hpp#L131-L133
and here?

etphipp:
No, just dot.

vqd8a:
@etphipp I am not quite clear why.

etphipp:
Inner products and norms are treated differently than general multiplies, which are what those functions do.

vqd8a:
but in the KokkosBlas1_team_mult_spec.hpp and KokkosBlas2_team_gemv_spec.hpp, we still need to compute products, right? Correct me if I am wrong. Thanks.

etphipp:
Yes, but these functions mathematically represent different operations, so you don't use the IPT there.

vqd8a:
Oh, I see. Thanks. @etphipp

crtrott:
Running tests now.

crtrott:
Mark you can open a separate issue for that. I only touched that file to fix some compile issues I had.

mhoemmen:
@crtrott wrote:
> Mark you can open a separate issue for that. I only touched that file to fix some compile issues I had.

Done: https://github.com/kokkos/kokkos-kernels/issues/225

crtrott:
```
Running on machine: apollo
Going to test compilers:  gcc/4.8.4 gcc/5.3.0 intel/16.0.1 clang/3.9.0 cuda/9.1
Testing compiler gcc/4.8.4
...
PASSED TESTS
clang-3.9.0-Pthread_Serial-release build_time=211 run_time=1508
gcc-4.8.4-Pthread-release build_time=123 run_time=911
FAILED TESTS
cuda-9.1-Cuda_OpenMP-release (test failed)
gcc-4.8.4-OpenMP-release (test failed)
gcc-5.3.0-Serial-release (test failed)
intel-16.0.1-OpenMP-release (test failed)
```

crtrott:
```
[crtrott@apollo KokkosKernelsBuild]$ grep "error:" nohup.PR215e
[crtrott@apollo KokkosKernelsBuild]$ grep "FAIL" nohup.PR215e
  FAILED gcc-4.8.4-OpenMP-release
[  FAILED  ] openmp.batched_vector_arithmatic_simd_dcomplex2 (137 ms)
[  FAILED  ] 1 test, listed below:
[  FAILED  ] openmp.batched_vector_arithmatic_simd_dcomplex2
 1 FAILED TEST
  FAILED gcc-5.3.0-Serial-release
[  FAILED  ] serial.batched_vector_arithmatic_simd_dcomplex2 (117 ms)
[  FAILED  ] 1 test, listed below:
[  FAILED  ] serial.batched_vector_arithmatic_simd_dcomplex2
 1 FAILED TEST
  FAILED intel-16.0.1-OpenMP-release
[  FAILED  ] openmp.batched_scalar_serial_trsv_l_nt_n_dcomplex_dcomplex (10 ms)
[  FAILED  ] openmp.batched_scalar_serial_trsv_l_nt_n_dcomplex_double (10 ms)
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] openmp.batched_scalar_serial_trsv_l_nt_n_dcomplex_dcomplex
[  FAILED  ] openmp.batched_scalar_serial_trsv_l_nt_n_dcomplex_double
 2 FAILED TESTS
  FAILED cuda-9.1-Cuda_OpenMP-release
[  FAILED  ] openmp.batched_vector_arithmatic_simd_dcomplex2 (126 ms)
[  FAILED  ] 1 test, listed below:
[  FAILED  ] openmp.batched_vector_arithmatic_simd_dcomplex2
 1 FAILED TEST
FAILED TESTS
```

crtrott:
Looks like this is all batched stuff. So I am going to approve this pull request and then we fix the rest.

ndellingwood:
Issue #231 was opened regarding the batched stuff, @kyungjoo-kim is working on it.

Issue #223, opened 2018-05-02
Make KokkosKernels work without KOKKOS_ENABLE_DEPRECATED_CODE


crtrott:

ndellingwood:
See kokkos/kokkos#1277

ndellingwood:
PR #227 is a start on this for search/replace changes needed.

ndellingwood:
PR #228 issued separately to handle the initialize and finalize calls in perf_tests.

PR #227, opened 2018-05-03
Issue 223

ndellingwood:
Work towards disabling deprecated code.

PR #228, opened 2018-05-04
perf_test: Remove exec_space initialize/finalize calls

ndellingwood:
Partially addresses issue #223

### DIFF
--- a/perf_test/graph/KokkosGraph_color.cpp
+++ b/perf_test/graph/KokkosGraph_color.cpp
@@ -436,10 +436,9 @@ int main (int argc, char ** argv){
   const int num_threads = params.use_openmp; // Assumption is that use_openmp variable is provided as number of threads
   const int device_id = 0;
   Kokkos::initialize( Kokkos::InitArguments( num_threads, -1, device_id ) );
-
   Kokkos::print_configuration(std::cout);
-#if defined( KOKKOS_ENABLE_OPENMP )
 
+#if defined( KOKKOS_ENABLE_OPENMP )
   if (params.use_openmp) {
 #ifdef KOKKOSKERNELS_MULTI_MEM
     KokkosKernels::Experiment::run_multi_mem_experiment
--- a/perf_test/graph/KokkosGraph_color.cpp
+++ b/perf_test/graph/KokkosGraph_color.cpp
@@ -437,10 +437,10 @@ int main (int argc, char ** argv){
   const int device_id = 0;
   Kokkos::initialize( Kokkos::InitArguments( num_threads, -1, device_id ) );
 
+  Kokkos::print_configuration(std::cout);
 #if defined( KOKKOS_HAVE_OPENMP )
 
   if (params.use_openmp) {
-    Kokkos::OpenMP::print_configuration(std::cout);
 #ifdef KOKKOSKERNELS_MULTI_MEM
     KokkosKernels::Experiment::run_multi_mem_experiment
     <size_type, idx, Kokkos::OpenMP, Kokkos::OpenMP::memory_space, Kokkos::HostSpace>(
@@ -458,7 +458,6 @@ int main (int argc, char ** argv){
 
 #if defined( KOKKOS_ENABLE_CUDA )
   if (params.use_cuda) {
-    Kokkos::Cuda::print_configuration(std::cout);
 #ifdef KOKKOSKERNELS_MULTI_MEM
     KokkosKernels::Experiment::run_multi_mem_experiment
     <size_type, idx, Kokkos::Cuda, Kokkos::Cuda::memory_space, Kokkos::CudaHostPinnedSpace>(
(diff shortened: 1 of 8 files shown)


ndellingwood:
I hope this doesn't conflict with #228, if so I'll fold them into a single PR. I kept the two separate since there is some independence in the changes, in case something goes wrong with one.

ndellingwood:
Will run spot-check Friday morn to avoid resource conflicts with the nightly Jenkins jobs.

perf_test/graph/KokkosGraph_color.cpp
```diff
@@ -433,12 +433,13 @@ int main (int argc, char ** argv){
...
+  const int num_threads = params.use_openmp; // Assumption is that use_openmp variable is provided as number of threads
+  const int device_id = 0;
+  Kokkos::initialize( Kokkos::InitArguments( num_threads, -1, device_id ) );
 
 #if defined( KOKKOS_HAVE_OPENMP )
 
   if (params.use_openmp) {
-
-    Kokkos::OpenMP::initialize( params.use_openmp );
     Kokkos::OpenMP::print_configuration(std::cout);
```
crtrott: All of this print_configuration stuff should be done via `Kokkos::print_configuration(std::cout)`

perf_test/sparse/KokkosSparse_pcg.cpp
```diff
@@ -307,27 +307,49 @@ int main (int argc, char ** argv){
   }
 
 
+  // Prepare for Kokkos::initialize() call
+  Kokkos::InitArguments init_args; // Construct with default args, change members based on exec space
```
crtrott: all this num_threads etc. setting should happen independent of the backend. You can always parse all the command line options and set those things. Even for CUDA. No need for backend specific code paths here.

perf_test/sparse/KokkosSparse_spgemm.cpp
```diff
@@ -293,11 +291,13 @@ int main (int argc, char ** argv){
...
-#if defined( KOKKOS_HAVE_OPENMP )
 
-  if (params.use_openmp) {
-    Kokkos::OpenMP::initialize( params.use_openmp );
+  const int num_threads = params.use_openmp; // Assumption is that use_openmp variable is provided as number of threads
+  const int device_id = params.use_cuda - 1;
+  Kokkos::initialize( Kokkos::InitArguments( num_threads, -1, device_id ) );
 
+#if defined( KOKKOS_HAVE_OPENMP )
+  if (params.use_openmp) {
```
crtrott: again no need for backend specific calls here.

perf_test/test_mv.cpp
```diff
@@ -263,22 +263,25 @@ int main(int argc, char **argv)
...
 
+ Kokkos::InitArguments args_init;
 
- KokkosCUDA(
-   Kokkos::Cuda::SelectDevice select_device(device);
-   Kokkos::Cuda::initialize( select_device );
- )
+ args_init.device_id = device;
+
+ KokkosCUDA()
```
crtrott: That can go away (the KokkosCUDA())

perf_test/sparse/KokkosSparse_pcg.cpp
ndellingwood: This test is a bit funky, since it is likely run by a script it may be ready to run with all three backends enabled since `exec_space::initialize()` and `exec_space::finalize()` were called within backend-specific if-guarded code. I think what I did would allow one backend's num_threads to overwrite the other's in this case. I'm going to change this so initialize and finalize are called within each backend's if-guards with the num_threads etc. parsing happening there.

perf_test/test_mv.cpp
ndellingwood: Removed.

perf_test/sparse/KokkosSparse_spgemm.cpp
ndellingwood: Removed the backend-specific print-configuration; left the backend if-guards so appropriate exec space and/or memory space can be set. Would like to move these calls to static functions in structs templated on exec space and mem space, but that will require changing whatever scripts are used for testing this and don't want to break them for now.

ndellingwood:
Spot-check on kokkos-dev, deprecated code still enabled:
```
[ndellin@kokkos-dev TestAllSandia-spot]$ ../../scripts/test_all_sandia --spot-check --skip-hwloc
Running on machine: sems
Going to test compilers:  gcc/5.3.0 intel/16.0.1 clang/3.9.0 cuda/8.0.44
Testing compiler gcc/5.3.0
Testing compiler intel/16.0.1
Testing compiler clang/3.9.0
Testing compiler cuda/8.0.44
PASSED TESTS
clang-3.9.0-Pthread_Serial-release build_time=300 run_time=1991
cuda-8.0.44-Cuda_OpenMP-release build_time=768 run_time=1188
gcc-5.3.0-Serial-release build_time=330 run_time=1085
intel-16.0.1-OpenMP-release build_time=480 run_time=383
```

ndellingwood:
Merged relevant changes from here into #227, closing this PR.

PR #229, opened 2018-05-04
BlockCrsMatrix: Fix gcc/4.8.4 shadow error

ndellingwood:
BlockDim ctor input variable caused shadow error due to
blockDim() member function name


ndellingwood:
Caught this trying to run spot-check on Apollo.

Issue #225, opened 2018-05-03
Team-based BLAS unnecessarily redefines real & imag from ArithTraits


mhoemmen:
@vqd8a , your recent changes to #215 defined the `RealPart` and `ImagPart` functions, for taking the real resp. imaginary part of a complex number (either `Kokkos::complex<T>`, or `SIMD<Kokkos::complex<T>>`).  `Kokkos::ArithTraits` already has `real` and `imag` functions.  For example:

https://github.com/kokkos/kokkos-kernels/blob/6e8e97a977564673bdadc15085d27a6a56b98114/src/Kokkos_ArithTraits.hpp#L442

Thus, you don't need these extra `RealPart` and `ImagPart` functions.  If you can instantiate `Kokkos::ArithTraits` for the `SIMD<Kokkos::complex<T>>` type, then you should be able to call the function in ArithTraits, and it should return the right thing.  If it doesn't return the right thing, then we should fix ArithTraits to do the right thing.

Please remove these extra functions, and instead define `real` and `imag` for `SIMD<Kokkos::complex<T>>` in ArithTraits.  I would be happy to review your PR.

@crtrott requested that I file this as a separate issue.

mhoemmen:
Here is an example code showing a possible solution:
```C++
#include <complex>
#include <type_traits>

template<class T> class ArithTraits {};
  
template<>
struct ArithTraits<double> {
  using mag_type = double;   
  static mag_type real (const double x) { return x; }
  static mag_type imag (const double x) { return 0.0; }
};

template<>
struct ArithTraits<std::complex<double>> {
  using mag_type = double;
  static mag_type real (const std::complex<double> x) { return std::real (x); }
  static mag_type imag (const std::complex<double> x) { return std::imag (x); }
};

template<class T>
struct SIMD {
  using value_type = T;
};

template<class T>
struct AVX {
  using value_type = T;
};

template<class X, int L> struct Vector {
  using value_type = typename X::value_type;
  value_type data_[L];
};

template<template<typename> typename X, class T, int L>
struct ArithTraits< Vector<X<T>, L> > {
  using value_type = Vector<X<T>, L>;
  using inner_value_type = typename X<T>::value_type;
  using inner_mag_type = typename ArithTraits<inner_value_type>::mag_type;
  using mag_type = Vector<X<inner_mag_type>, L>;

  static mag_type real (const value_type z) {
      mag_type w;
      for (int k = 0; k < L; ++k) {
[...]

PR #232, opened 2018-05-07
disable avx set when gcc is used.

kyungjoo-kim:
GCC on Sandybridge does not support avx2. Even if I use the avx set only in the code, it causes unit test failures. Since we are not able to detect an architecture type, we disable for gcc.

PR #236, opened 2018-05-08
Issue 225

vqd8a:
Address #225


vqd8a:
Added arith traits overload for vector complex types in [KokkosBatched_Vector.hpp](https://github.com/kokkos/kokkos-kernels/blob/issue-225/src/batched/KokkosBatched_Vector.hpp) 
Modified [Test_Batched_VectorArithmatic.hpp](https://github.com/kokkos/kokkos-kernels/blob/issue-225/unit_test/batched/Test_Batched_VectorArithmatic.hpp) respectively.
Still leave RealPart() in [KokkosBatched_Vector.hpp](https://github.com/kokkos/kokkos-kernels/blob/issue-225/src/batched/KokkosBatched_Vector.hpp) since I am not sure where it is called.

vqd8a:
Thanks, @kyungjoo-kim.
Couldn't these be taken care by imag() in ArithTraits? Why did we need ImagPart for these cases?
```C++
template<typename T> KOKKOS_FORCEINLINE_FUNCTION typename Kokkos::Details::ArithTraits<T>::mag_type ImagPart(const T &val) { return 0; }
template<typename T> KOKKOS_FORCEINLINE_FUNCTION typename Kokkos::Details::ArithTraits<Kokkos::complex<T> >::mag_type ImagPart(const Kokkos::complex<T> &val) { return val.imag(); }
```
I will change RealPart and ImagPart used in other routines too.

kyungjoo-kim:
if you added traits, it is not necessary. they are just quick solution when I wrote first time.

vqd8a:
Changed RealPart and ImagPart used in other routines to real() and imag().

vqd8a:
Thanks @mhoemmen . The code is fixed.

kyungjoo-kim:
@vqd8a Yhis needs to wait until #232 is merged first.

vqd8a:
Thanks, @kyungjoo-kim . Could you please help answer @mhoemmen 's questions about MagnitudeScalarType and ValueScalarType?

kyungjoo-kim:
@vqd8a They are probably a bi-product when I develop the code. It is a quick solution without thinking too much ;). I don't think that I use it in the code but we can change the code as @mhoemmen suggests.

mhoemmen:
@kyungjoo-kim PURGE THEM :D

vqd8a:
Since ValueScalarType is not used elsewhere, I just removed it.
As @mhoemmen suggested, I also changed MagnitudeScalarType to ArithTraits<T>::mag_type. Other routines that used MagnitudeScalarType were also changed accordingly.

src/batched/KokkosBatched_LU_Decl.hpp
```diff
@@ -16,7 +16,7 @@ namespace KokkosBatched {
       KOKKOS_INLINE_FUNCTION
       static int
       invoke(const AViewType &A,
-             const typename MagnitudeScalarType<typename AViewType::non_const_value_type>::type tiny = 0);
+             const typename Kokkos::Details::ArithTraits<typename AViewType::non_const_value_type>::mag_type tiny = 0);
     };
```
kyungjoo-kim: This is wrong. In the batched routines, input value_type can be two -  double or SIMD<double> type. When SIMD<double> type is used, your modification returns SIMD<mag_type<double> > but the previous routine MagnitudeScalarType returns double type.

vqd8a: Should we keep ```MagnitudeScalarType```?

kyungjoo-kim: You have two options: keep MagnitudeScalarType or using the ValueScalarType
```
typename Kokkos::Details::ArithTraits<typename ValueScalarType<typename AViewType::non_const_value_type>::type>::mag_type
```

vqd8a: I think we should go back to MagnitudeScalarType

vqd8a:
@kyungjoo-kim Please review these changes.

kyungjoo-kim:
BTW, did you check the code compiles and pass all the unit tests ?

vqd8a:
thanks @kyungjoo-kim . Yes, it passed all the unit tests.

kyungjoo-kim:
thank you for fixing the code.

crtrott:
Passed tests other than the clang 6 error in SPGEMM_KK_MEMORY:
```
PASSED TESTS
clang-3.9.0-Pthread_Serial-release build_time=225 run_time=1702
cuda-9.1-Cuda_OpenMP-release build_time=619 run_time=1091
gcc-4.8.4-OpenMP-release build_time=138 run_time=317
gcc-4.8.4-Pthread-release build_time=125 run_time=879
gcc-5.3.0-Serial-release build_time=153 run_time=763
intel-16.0.1-OpenMP-release build_time=403 run_time=374
FAILED TESTS
clang-6.0-Cuda_Pthread-release (test failed)
```

src/batched/KokkosBatched_Vector.hpp
```diff
@@ -138,13 +138,13 @@ namespace Kokkos {
 	
 	
     template<typename T, int l>
-    class ArithTraits<Vector<SIMD<Kokkos::complex<T>>,l> > { 
+    class ArithTraits<KokkosBatched::Experimental::Vector<KokkosBatched::Experimental::SIMD<Kokkos::complex<T>>,l> > {
```
mhoemmen: No love for my template template, alas...

crtrott: Uhm that didn't compile without the change ...

vqd8a: when I looked at the old code RealPart and ImagPart, I saw they were only for SIMD. That's why I did not try @mhoemmen 's solution template template.
Now, I have just tried it and it works. Should I create a new PR for this?

mhoemmen: @vqd8a I don't mean to tell you what to do :-)  If it needs to work for AVX, etc. as well as SIMD, then my approach would give you a generic solution.

vqd8a: @crtrott @kyungjoo-kim should we need this generic solution?

mhoemmen: (You don't _need_ it, but you do need ArithTraits, etc. to work for `AVX<T>` and any other `SimdImplementation<T>` you want to support, as well as `SIMD<T>`.  However that happens is up to you :-) .)

kyungjoo-kim:
right now I unified avx into simd. but I still open the possibility that someone else wants to try this kind of structure with a different tag that can guide to a different instruction set e.g., ARM. I think that the current solution is fine and let's not worry about things not happend yet.

Issue #250, opened 2018-06-04
Removes "is_convertible" in the batch code.


kyungjoo-kim:
It is too strict to put is_convertible when the simd code uses complex together and it causes more unnecessary problems than error filtering.

ndellingwood:
Addressed by @kyungjoo-kim in #251

kyungjoo-kim:
done

PR #251, opened 2018-06-04
KokkosKernels - remove is_convertible #250

kyungjoo-kim:
To assert is_convertible, it needs over-riding for complex variables and overriding std::is_convertible is forbidden by the language. So, we relax the condition and remove is_convertible.

### DIFF
--- a/src/batched/KokkosBatched_Util.hpp
+++ b/src/batched/KokkosBatched_Util.hpp
@@ -3,7 +3,7 @@
 
 /// \author Kyungjoo Kim (kyukim@sandia.gov)
 
-#define __KOKKOSBATCHED_PROMOTION__ 1
+//#define __KOKKOSBATCHED_PROMOTION__ 1
 
 #include <iomanip>
 #include <random>
--- a/src/batched/KokkosBatched_Vector_SIMD.hpp
+++ b/src/batched/KokkosBatched_Vector_SIMD.hpp
@@ -75,7 +75,6 @@ namespace KokkosBatched {
       }
       template<typename ArgValueType>
       KOKKOS_INLINE_FUNCTION Vector(const Vector<SIMD<ArgValueType>,vector_length> &b) {
-        static_assert(std::is_convertible<value_type,ArgValueType>::value, "input type is not convertible");
 #if defined( KOKKOS_ENABLE_PRAGMA_IVDEP )
 #pragma ivdep
 #endif
@@ -171,7 +170,6 @@ namespace KokkosBatched {
 
       template<typename ArgValueType>
       KOKKOS_INLINE_FUNCTION Vector(const Vector<SIMD<ArgValueType>,vector_length> &b) {
-        static_assert(std::is_convertible<value_type,ArgValueType>::value, "input type is not convertible");
 	auto dd = reinterpret_cast<value_type*>(&_data);
 	auto bb = reinterpret_cast<ArgValueType*>(&b._data);
 #if defined( KOKKOS_ENABLE_PRAGMA_IVDEP )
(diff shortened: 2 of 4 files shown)


ndellingwood:
@kyungjoo-kim were the `static_assert`s unnecessary or should they be replaced by another check? For example, some type of meta-function similar to `is_convertible`. 
Also, could you run a spot-check on this or would you like me to just in case something unexpected fails?

ndellingwood:
If this fix is necessary for trilinos/Trilinos#2871 and PR trilinos/Trilinos#2879 these changes should also be saved as a patch that can be applied to kokkos-kernels in Trilinos and issued as a hotfix PR.

kyungjoo-kim:
What I like to do is 
```
static_assert<std::is_convertible<double, SIMD<Kokkos::complex<double>,2> >::value, "double is convertible to simd complex double");
```
c++ language prohibits overriding for user defined types. So putting the assert is not feasible. Lets make the problem simple. So does the solution.

ndellingwood:
@kyungjoo-kim then this is fine. I'll start a spot-check on Apollo with clang/6.0 removed from the list.

kyungjoo-kim:
@ndellingwood  I am running the spotcheck on bowman. I will post output after it finishes.

ndellingwood:
@kyungjoo-kim sounds good, should I run cuda/9.1 for completeness even though changes should not have an impact with Cuda?

kyungjoo-kim:
@ndellingwood No it won'ts need CUDA. We only remove static asserts. If it compiles (without any extra typos), it is sufficient.

ndellingwood:
For reference, if you haven't had to create a patch to Trilinos after committing changes and submitting PR the easiest way to create the patch is probably:
```
cd kokkos-kernels
git checkout issue-250
git diff develop > issue250.patch
```
Copy or move  the patch to `Trilinos/packages/kokkos-kernels` then apply it:

`patch -p1 < issue250.patch`

@kyungjoo-kim if you have time to put in the Trilinos PR, even before your spot-check finishes, the autotester can get started since waiting on the tester will slow down your other work (I can approve the PR there too)

kyungjoo-kim:
@ndellingwood Can I add this patch to the existing PR ?

kyungjoo-kim:
@ndellingwood If you don't mind, could you please run spotcheck ? Today my workstation has very unstable network and I got dropped before it finishes spotcheck.

ndellingwood:
@kyungjoo-kim will do :)
This isn't very useful, but I had already kicked off the cuda/9.1 check before your followup message and things passed:
```
PASSED TESTS
cuda-9.1-Cuda_OpenMP-release build_time=621 run_time=908
```
Other checks in progress now.

ndellingwood:
Apollo spot-check passes remaining builds: 
```
Running on machine: apollo
Going to test compilers:  gcc/4.8.4 gcc/5.3.0 intel/16.0.1 clang/3.9.0
Testing compiler gcc/4.8.4
  Starting job gcc-4.8.4-OpenMP-release
  PASSED gcc-4.8.4-OpenMP-release
Testing compiler gcc/5.3.0
  Starting job gcc-4.8.4-Pthread-release
  PASSED gcc-4.8.4-Pthread-release
Testing compiler intel/16.0.1
Testing compiler clang/3.9.0
PASSED TESTS
clang-3.9.0-Pthread_Serial-release build_time=212 run_time=1677
gcc-4.8.4-OpenMP-release build_time=137 run_time=333
gcc-4.8.4-Pthread-release build_time=124 run_time=1036
gcc-5.3.0-Serial-release build_time=154 run_time=915
intel-16.0.1-OpenMP-release build_time=336 run_time=315
```

mhoemmen:
@ndellingwood wrote:
> were the `static_assert`s unnecessary or should they be replaced by another check? For example, some type of meta-function similar to `is_convertible`.

It shouldn't have been necessary to override `std::is_convertible` in this case; the standard definition uses SFINAE to check whether the assignment `From = To` compiles.

srajama1:
@kyungjoo-kim : Everyting ok to merge ? I can click the green button.

kyungjoo-kim:
click it~

Issue #240, opened 2018-05-16
Add L1 variant to multithreaded Gauss-Seidel


jhux2:
L1 Gauss-Seidel* has proved to be useful as a parallel smoother in multigrid.  I'd like to request that this option be added to the multithreaded Gauss-Seidel kernel.

* "Multigrid Smoothers for Ultraparallel Computing", doi 10.1137/100798806.

jhux2:
@csiefer2 @pwxy

mhoemmen:
@jhux2 Just curious -- is that an option that MueLu normally exercises by default?

jhux2:
@mhoemmen By default, no.  But it's used in the exawind project, and I think in Alegra (@csiefer?).

srajama1:
@jhux2: Any deadlines that I should be aware of ?

csiefer:
Funny jhux2 added me here but GitHub confused my with siefer2 (?)

mhoemmen:
@csiefer I think you are correct; sorry to bother you!  GitHub autocomplete sometimes causes issues; poor jhux got bothered a lot before we realized this ;-) .  @csiefer2 is the person @jhux2 meant to tag.

csiefer2:
At medium-to-large core counts, you really want to use l1...

mhoemmen:
@csiefer2 Nobody does!!! It frustrates me :(

srajama1:
#255 adds these feature.

Issue #221, opened 2018-05-02
Multithreaded Gauss-Seidel does not support damping


mhoemmen:
See https://github.com/trilinos/Trilinos/issues/2665

pwxy:
I tried MTGS and MTSGS with damping, and it ignores the ifpack2 damping parameter (never even uses it)

srajama1:
@pwxy We will prioritize this.

srajama1:
#255 fixes these

PR #255, opened 2018-06-11
MT Gauss-Seidel fixups

cgcgcg:
- Add damping option
- Instead of inverting the diagonal in every iteration, save the inverse of the diagonal
- Allow to pass a precomputed diagonal. 

The last item is necessary for MT GS to work correctly with Ifpack2. The diagonal of a Tpetra matrix might not be given by pairs (i, i) of local IDs (depending on the rowmap and the colmap). Moreover, this allows to use MT GS with the l1 option from Ifpack2.

I also added some routines to dump to MatrixMarket format.

This addresses Issues #221 and #240

### DIFF
--- a/src/sparse/KokkosSparse_gauss_seidel.hpp
+++ b/src/sparse/KokkosSparse_gauss_seidel.hpp
@@ -421,6 +421,7 @@ namespace KokkosSparse{
                                             y_scalar_view_t y_rhs_input_vec,
                                             bool init_zero_x_vector = false,
                                             bool update_y_vector = true,
+                                            typename KernelHandle::nnz_scalar_t omega = Kokkos::Details::ArithTraits<typename KernelHandle::nnz_scalar_t>::one(),
                                             int numIter = 1){
       handle->get_gs_handle()->set_block_size(block_size);
       symmetric_gauss_seidel_apply(handle,num_rows,num_cols,
@@ -431,6 +432,7 @@ namespace KokkosSparse{
                                    y_rhs_input_vec,
                                    init_zero_x_vector,
                                    update_y_vector,
+                                   omega,
                                    numIter);
     }
     template <class KernelHandle,
--- a/src/sparse/KokkosSparse_gauss_seidel.hpp
+++ b/src/sparse/KokkosSparse_gauss_seidel.hpp
@@ -49,689 +49,689 @@
 
 namespace KokkosSparse{
 
-namespace Experimental{
+  namespace Experimental{
 
-  template <typename KernelHandle, typename lno_row_view_t_, typename lno_nnz_view_t_>
-  void gauss_seidel_symbolic(
-      KernelHandle *handle,
-      typename KernelHandle::const_nnz_lno_t num_rows,
-      typename KernelHandle::const_nnz_lno_t num_cols,
-      lno_row_view_t_ row_map,
-      lno_nnz_view_t_ entries,
-	  bool is_graph_symmetric = true){
-
-
-
-	  static_assert (std::is_same<typename KernelHandle::const_size_type,
-			  typename lno_row_view_t_::const_value_type>::value,
-			  "KokkosSparse::gauss_seidel_symbolic: Size type of the matrix should be same as kernelHandle sizetype.");
-
-	  static_assert (std::is_same<typename KernelHandle::const_nnz_lno_t,
-			  typename lno_nnz_view_t_::const_value_type>::value,
-			  "KokkosSparse::gauss_seidel_symbolic: lno type of the matrix should be same as kernelHandle lno_t.");
+    template <typename KernelHandle, typename lno_row_view_t_, typename lno_nnz_view_t_>
+    void gauss_seidel_symbolic(
+                               KernelHandle *handle,
+                               typename KernelHandle::const_nnz_lno_t num_rows,
+                               typename KernelHandle::const_nnz_lno_t num_cols,
+                               lno_row_view_t_ row_map,
+                               lno_nnz_view_t_ entries,
+                               bool is_graph_symmetric = true){
 
 
-	  typedef typename KernelHandle::const_size_type c_size_t;
-	  typedef typename KernelHandle::const_nnz_lno_t c_lno_t;
-	  typedef typename KernelHandle::const_nnz_scalar_t c_scalar_t;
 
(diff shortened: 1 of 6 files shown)


src/common/KokkosKernels_SparseUtils.hpp
```diff
@@ -2133,7 +2133,117 @@ void kk_reduce_numrows_larger_than_threshold(size_t num_elements, view_type view
...
+
+  // Print the Matrix Market banner line.  We assume
+  // nonsymmetric storage ("general").
+
+  out << "%%MatrixMarket matrix array " << dataType << " general" << endl;
+
+  // Write the dimensions of the sparse matrix: (# rows, #
+  // columns, # matrix entries (counting duplicates as
+  // separate entries)).
+  out << numRows << " " << 1 << " " << endl;
```
srajama1: Sorry to nit pick, space in the end of line ?

srajama1: Sorry to nit pick, space in the end of line ?

src/sparse/KokkosSparse_gauss_seidel_handle.hpp
```diff
@@ -247,12 +247,12 @@ class GaussSeidelHandle{
   void set_old_to_new_map(const nnz_lno_persistent_work_view_t &old_to_new_map_) {
     this->old_to_new_map = old_to_new_map_;
   }
-  void set_permuted_diagonals (const scalar_persistent_work_view_t permuted_diagonals_){
-    this->permuted_diagonals = permuted_diagonals_;
+  void set_permuted_inverse_diagonal (const scalar_persistent_work_view_t permuted_inverse_diagonal_){
```
srajama1: Does Ifpack2 already call set_permuted_diagonals ?

src/sparse/KokkosSparse_gauss_seidel.hpp
```diff
@@ -193,7 +193,75 @@ namespace Experimental{
...
+			  "KokkosSparse::gauss_seidel_symbolic: Size type of the matrix should be same as kernelHandle sizetype.");
+
+	  static_assert (std::is_same<typename KernelHandle::const_nnz_lno_t,
+			  typename lno_nnz_view_t_::const_value_type>::value,
+			  "KokkosSparse::gauss_seidel_symbolic: lno type of the matrix should be same as kernelHandle lno_t.");
+
+	  static_assert (std::is_same<typename KernelHandle::const_nnz_scalar_t,
+	  			  typename scalar_nnz_view_t_::const_value_type>::value,
+	  			  "KokkosSparse::gauss_seidel_symbolic: scalar type of the matrix should be same as kernelHandle scalar_t.");
+
```
srajama1: This is mainly pattern match, so I assume this is ok.

src/sparse/KokkosSparse_gauss_seidel_handle.hpp
cgcgcg: No, it doesn't. This method only sets the permuted diagonal, it does not do any permutations.

src/common/KokkosKernels_SparseUtils.hpp
```diff
@@ -2133,7 +2133,117 @@ void kk_reduce_numrows_larger_than_threshold(size_t num_elements, view_type view
...
+    if (KAM::isComplex) {
+      out << KAM::real (val) << " " << KAM::imag (val);
+    } else {
+      out << val;
+    }
+    out << endl;
+    }
+  }
+
+
```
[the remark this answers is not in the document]
cgcgcg: Oh, I didn't see those! Then there is obviously no reason to add this again.

cgcgcg:
@srajama1 I removed the matrix market routines and took care of all the whitespace and indentation issues. This blows up the diff quite significantly, but there are no functional changes. I ran the unit test on white and bowman. Anything else I should be running?

src/sparse/KokkosSparse_gauss_seidel.hpp
```diff
@@ -49,611 +49,689 @@
...
+      Internal_alno_row_view_t_ const_a_r (row_map.data(), row_map.extent(0));
+      Internal_alno_nnz_view_t_ const_a_l (entries.data(), entries.extent(0));
+      Internal_ascalar_nnz_view_t_ const_a_v (values.data(), values.extent(0));
+      Internal_ascalar_nnz_view_t_ const_a_d (given_inverse_diagonal.data(), given_inverse_diagonal.extent(0));
+
+
+      using namespace KokkosSparse::Impl;
+
+      GAUSS_SEIDEL_NUMERIC<const_handle_type, Internal_alno_row_view_t_, Internal_alno_nnz_view_t_, Internal_ascalar_nnz_view_t_>::gauss_seidel_numeric
+        (&tmp_handle, num_rows, num_cols, const_a_r, const_a_l, const_a_v, const_a_d, is_graph_symmetric);
```
cgcgcg: @srajama1 The diagonal is passed here as `const_a_d`.

srajama1: Thanks, missed the basic thing.

srajama1:
Will review this after a meeting.

srajama1:
A unit test in KK will be nice as it gets tested on lot more compilers and platforms than Ifpack2.

cgcgcg:
@srajama1 I changed the unit test so it would run with damping. I'll build and test on Bowman and White..

cgcgcg:
@srajama1 Tests pass on Bowman and White.

cgcgcg:
The more I think about it, the less I am convinced that this is correct. It will not bug out, as it did before, but it probably is still not quite correct. The graph that is used for the coloring is given in terms of local row and column IDs as well, so the coloring might be off. I'll look into it when I have time.

srajama1:
The coloring is on symmetric graph and only on the local part. Assuming there is no unsymmetric ordering of local rows and local columns it should be fine. If row i and column i are not the same vertex then we are coloring what is just in the row adjacency and ignoring any mismatch in column adjacency.

jhux2:
@srajama1 @cgcgcg How much work is involved in getting this merged?  Having MT Gauss-Seidel is important for the EMPIRE L2 milestone.

cgcgcg:
@srajama1 Do you think this can be merged?

srajama1:
Let us fix column map != row map outside KK. We can merge this.

Issue #247, opened 2018-05-29
TPL Support for all BLAS functions using CuBLAS


crtrott:

vqd8a:
@crtrott Could you please send me some examples of using ```singleton``` for defining cuBLAS ```handle```? It would be helpful. Thanks.

ndellingwood:
Cross referencing PR #262

mhoemmen:
@vqd8a When you create the cuBLAS handle (be sure only to create it once -- don't recreate it each time!), set up a Kokkos::finalize hook to free it.

vqd8a:
Thanks @mhoemmen . ```cublasCreate``` and ```cublasDestroy``` are called only once.

srajama1:
@vqd8a : Please create a wiki page with all the functions we support.

vqd8a:
@srajama1 I added a wiki page for these.

PR #262, opened 2018-06-18
TPL Support for BLAS functions (nrm2, axpy, dot, gemm) using CuBLAS (Issue #247)

vqd8a:
Added cuBLAS TPL Support for nrm2, axpy, dot. More to come.


srajama1:
Is the BLAS3 (GEMM especially) done ? We have an use for it, so it would be nice to prioritize that next if you can.

srajama1:
Do we have to do anything in generate_makefile ?

vqd8a:
@srajama1 I made a minor change in ```Makefile.kokkos-kernels``` which adds the source file for handling cuBLAS. 
I have not had an experience in Trilinos integration. Maybe @crtrott can answer.
I haven't done BLAS3 yet. But can work on it now.

kyungjoo-kim:
I remembered that @mhoemmen commented that two versions of cublas should be harnessed for sierra build. @mhoemmen How do we test it ?

mhoemmen:
@kyungjoo-kim Just don't expose any of the cublas headers in any header file, and you should be fine.  The main issue is that FETI uses v2 of the cuBLAS API.  If you include the v1 cuBLAS header, FETI won't build.

mhoemmen:
@kyungjoo-kim It's OK to use both versions of the cuBLAS API in the same executable, but you aren't allowed to include both header files in the same compilation unit.

mhoemmen:
@kyungjoo-kim If you must expose the header file (ask yourself why you really must; you almost certainly do not), then please use v2 of the cuBLAS API.  That shouldn't break FETI.

Makefile.kokkos-kernels
```diff
@@ -386,6 +386,8 @@ ifeq ($(KOKKOSKERNELS_INTERNAL_INST_EXECSPACE_CUDA), 1)
   endif
 endif
 
+KOKKOSKERNELS_INTERNAL_SRC_BLAS += ${KOKKOSKERNELS_PATH}/src/impl/generated_specializations_cpp/KokkosBlas_Cuda_tpl.cpp
```
crtrott: This file should not be in a generated_specializations_cpp. That should only have auto generated files.

src/impl/tpls/KokkosBlas_Cuda_tpl.hpp
```diff
@@ -0,0 +1,24 @@
...
+    printf("cuBLAS context initialized.\n");
+}
+
+CudaBlasSingleton::~CudaBlasSingleton()
+{
+    cublasDestroy(handle); printf("cuBLAS context released.\n");
+}
+
+CudaBlasSingleton & CudaBlasSingleton::singleton()
+{ static CudaBlasSingleton s ; return s ; }
```
crtrott: Do we need to do this via a pointer or a create function or something. cublasCreate needs to be called after Kokkos::initialize().

crtrott:
v1 is deprecated. We should not include that ...

Makefile.kokkos-kernels
vqd8a: @crtrott I am moving this file to ```tpls```

vqd8a:
@srajama1 cuBLAS GEMM support has been added.

src/impl/tpls/KokkosBlas_Cuda_tpl.hpp
vqd8a: @crtrott cublasCreate is already called after Kokkos::initialize()

mhoemmen:
If the handle only ever gets created with the default CUDA stream, then the right way to deal with it is this:

  1. Create it once, lazily, on demand
  2. Add it as a Kokkos::finalize hook

If it's possible to create the handle with other CUDA streams, then users (such as Tpetra and other downstream software) need a way to create, manage, and destroy handles.

srajama1:
I was wondering if the users like Tpetra would like to control things like cublasSetMathMode, cublasSetAtomicMode, cublasSetMatrix/VectorAsync, and cuBLASSetStream. KK can leave most of it to default, but making sure we don't force it. The disadvantage of the handle is of course the users have to worry about it even w/ other libraries like MKL (or with no libraries).

mhoemmen:
> I was wondering if the users like Tpetra would like to control things like cublasSetMathMode, cublasSetAtomicMode, cublasSetMatrix/VectorAsync, and cuBLASSetStream. 

  - `cublasSet{Matrix,Vector}Async` appear to be redundant with respect to `Kokkos::deep_copy`
  - "math mode" only appears relevant for `float` and `half`, not for `double` etc.
  - "atomic mode" would give nondeterministic results so I'd rather not enable it unless somebody really insists

That leaves setting the stream.  I could see users wanting to do that.  MAGMA has a handle tied to a CUDA stream, for example.  Better exposure of distinct CUDA streams would help us overlap data movement and computation.

The main value of hiding the stream is for gradually porting from existing interfaces like `Teuchos::BLAS` to this new interface.  I think that's a good idea.  It will be a while before Trilinos is ready to use multiple CUDA streams, so maybe we don't have to worry about that for now.  Even three-argument `Kokkos::deep_copy` currently ignores its execution space instance argument: https://github.com/kokkos/kokkos/blob/d3a941925cbfb71785d8ea68259123ed52d3f9da/core/src/Kokkos_CopyViews.hpp#L1511

vqd8a:
Thanks @mhoemmen for your suggestions. While waiting for decision, I have just fixed the use of kokkos-kernes macros to detect Scalar types as you suggested.

src/impl/tpls/KokkosBlas1_dot_tpl_spec_avail.hpp
```diff
@@ -93,10 +100,18 @@ Kokkos::View<const SCALAR*, LAYOUT, Kokkos::Device<ExecSpace, MEMSPACE>, \
              Kokkos::MemoryTraits<Kokkos::Unmanaged> >, \
 1,1> { enum : bool { value = true }; };
 
+#if defined (KOKKOSKERNELS_INST_DOUBLE)
 KOKKOSBLAS1_DOT_TPL_SPEC_AVAIL_CUBLAS( double,                  Kokkos::LayoutLeft, Kokkos::CudaSpace)
+#endif
+#if defined (KOKKOSKERNELS_INST_FLOAT)
 KOKKOSBLAS1_DOT_TPL_SPEC_AVAIL_CUBLAS( float,                   Kokkos::LayoutLeft, Kokkos::CudaSpace)
+#endif
+#if defined (KOKKOSKERNELS_INST_KOKKOS_COMPLEX_DOUBLE_)
```
mhoemmen: How come the complex double and complex float macros end with underscores, but the double and float ones don't?

vqd8a: I think we can use either without or with underscore, since in the ```KokkosKernels_config.h```, it uses
```#define KOKKOSKERNELS_INST_COMPLEX_DOUBLE
#if defined(KOKKOSKERNELS_INST_COMPLEX_DOUBLE)
#define KOKKOSKERNELS_INST_KOKKOS_COMPLEX_DOUBLE_
#endif```
 but I just followed the convention in hpp files in ```\src\impl\generated_specializations_hpp```

Comment (mhoemmen) [reply src/impl/tpls/KokkosBlas1_dot_tpl_spec_avail.hpp]: OK, just checking :)

**Comment by srajama1** [comment]
@crtrott : Can you check this ? Need your opinion on two things, exposing (or not exposing) the handle  and the ETI use case mentioned above.

**Comment by vqd8a** [comment]
@srajama1 , I talked to @crtrott last week. @crtrott said not exposing handles could be the option for now. and leave exposing for future. If it is the case, I can run the spot-check when the testbeds are back on.

**Comment by srajama1** [comment]
Agreed. Run the spot check when you can. @crtrott still needs to approve it as it is blocked on his request for changes.

**Comment by crtrott** [comment]
Looks good to me if you can run the spot-check. But also open a follow up: we may want to have the TPL variants independent of the INST macros. You see the INST macros don't say whether something is available, they say wether we do ETI. As long as you don't also define ETI only, all the other calls can still be made. And I think it is fine in that case to have the TPL calls available.

**Comment by vqd8a** [comment]
Spot check on white:
kokkos-kernels/scripts/test_all_sandia cuda --spot-check --with-cuda-options=enable_lambda --with-tpls=cublas

```Running on machine: white
Going to test compilers:  cuda/8.0.44 cuda/9.0.103
Testing compiler cuda/8.0.44
Testing compiler cuda/9.0.103
  Starting job cuda-8.0.44-Cuda_Serial-release
  PASSED cuda-8.0.44-Cuda_Serial-release
  Starting job cuda-9.0.103-Cuda_OpenMP-release
  PASSED cuda-9.0.103-Cuda_OpenMP-release
  Starting job cuda-9.0.103-Cuda_Serial-release
  PASSED cuda-9.0.103-Cuda_Serial-release
PASSED TESTS
cuda-8.0.44-Cuda_OpenMP-release build_time=823 run_time=751
cuda-8.0.44-Cuda_Serial-release build_time=792 run_time=1078
cuda-9.0.103-Cuda_OpenMP-release build_time=862 run_time=761
cuda-9.0.103-Cuda_Serial-release build_time=800 run_time=1087
```

srajama1:
Once you have a KNL spot check we can merge this.

srajama1:
Ignore the silly request for KNL, this shouldn't affect any KNL, so merging.

PR #248, opened 2018-05-31
WIP: Distance2 Coloring Update

william76:
-- Work in Progress --

For Luc
--------

@lucbv - You will probably be interested in these updates.  I've tried to reduce the confusing on the naming of files and some functions (the "is it graph_d2 or d2_graph" issue).  Let me know if you think this would reduce the confusing parts.  This update would also contain the timing updates you're interested in.  The test program prints out two CSV lines at the end, prefixed in the first column like:

```
CSVHDR,Filename,Num Rows,Num Edges,Execution Space,Concurrency,Algorithm,Avg Time,Avg Colors,Avg Num Phases
CSVDATA,Laplace_284_A.bin,22906304,159860192,OpenMP,8,COLORING_D2,0.75893,17,1
```

This lets us run a bunch of experiments and then assemble a CSV file containing the results with two  grep commands:

```
grep "CSVHDR" file1.output > results.csv
grep "CSVDATA" *.output >> results.csv
```

This was the first round of putting *something* in to measure timing.  Right now it's just total execution time, but we should coordinate what we want to break out for timing -- I'm thinking we'll want to time the individual phases of the algorithm (ColorGreedy, FindConflicts, ResolveConflicts) too.

For Siva
--------

@srajama1 - Still work in progress but you and I might want to sync up on what I'm trying to do here.  To gauge impact of renaming some stuff, do you know if anyone other than MueLu cares about the distance-2 stuff yet?  What about the distance-1 coloring?  I'd like to rename the `KokkosGraph_graph_color.hpp` to `KokkosGraph_GraphColor.hpp` to be consistent with its _impl file.  Where there any projects that @mndevec did the D1 coloring for that would need to know if we're changing a file name?  

Comments
-----------

There are a lot of other updates.  Mostly focused on some style refactoring in the Distance-2 coloring kernel code and cleanup.  The comment in  2cb3d5a details the bigger changes involving renaming files and functions.

I've created a new perf_test, `KokkosGraph_color_d2_wcmclen.hpp` for testing Team Policy work on the existing Distance-2 Kernel.  Once the team policy work is completed I expect this file will replace the current d2 coloring test file.

### Commit Comment for 2cb3d5a
Changing name of execution method in Distance-2 coloring classes
to just be execute since MatrixSquared and the Direct method are
in separate classes anyhow.  It's redundant to have the method
that runs the algorithm be:
- `GraphColorD2_MatrixSquared::color_graph_d2_matrix_squared()`
[...]

PR #263, opened 2018-06-18
WIP: Distance2 Graph Coloring

william76:
Distance-2 Graph Coloring Update

*Work in Progress*

This supersedes #248 (closed)

Added different variations of parallelism to the D2 coloring code across the primary loops in the `GreedyColor` functor:

- **COLORING_D2_SERIAL**:
  - This executes [GraphColor::d2_color_graph()][1], the serial-only version.  
  - Requires `Serial` mode enabled when building.
- **COLORING_D2_VB**:
  - Single level parallelism
  - [P] loop over chunks of vertices
  - [S] loop over vertices in chunks
  - [S] loop over vertex neighbors
  - [S] loop over vertex neighbors of neighbors 
- **COLORING_D2_VBTP**: 
  - Two level parallelism
  - [P] loop over chunks of vertices
  - [P] loop over vertices in chunks
  - [S] loop over vertex neighbors
  - [S] loop over vertex neighbors of neighbors 
- **COLORING_D2_VBTP2**: 
  - Two level parallelism
  - [P] loop over chunks of vertices
  - [S] loop over vertices in chunks
  - [P] loop over vertex neighbors
  - [S] loop over vertex neighbors of neighbors 
- **COLORING_D2_VBTP3**: 
  - Two level parallelism
  - [P] loop over chunks of vertices
  - [S] loop over vertices in chunks
  - [S] loop over vertex neighbors
  - [P] loop over vertex neighbors of neighbors 
- **COLORING_D2_VBTPVR1**: 
  - Three level parallelism
  - [P] loop over chunks of vertices
  - [P] loop over vertices in chunks
  - [P] loop over vertex neighbors
  - [S] loop over vertex neighbors of neighbors 
- **COLORING_D2_VBTPVR2**: 
  - Three level parallelism
  - [P] loop over chunks of vertices
  - [P] loop over vertices in chunks
  - [S] loop over vertex neighbors
  - [P] loop over vertex neighbors of neighbors 

These are available to the application: KokkosGraph_color_d2_wcmclen.exe

@srajama1 
@lucbv 
@jhux2 

[1]: https://github.com/william76/kokkos-kernels/blob/Distance2GraphColor/src/graph/impl/KokkosGraph_GraphColor_impl.hpp#L206


jhux2:
@srajama1

william76:
@srajama1 
The forbidden array size is not O(V) in size... it's size is 64 as defined by this (line [60[1]):

```
#define VB_D2_COLORING_FORBIDDEN_SIZE 64
```

Then the `forbidden` arrays are allocated inside the functors like this as in [functorGreedyColorVB][2]:
```
// Use forbidden array to find available color.
// - should be small enough to fit into fast memory (use Kokkos memoryspace?)
bool forbidden[VB_D2_COLORING_FORBIDDEN_SIZE];      // Forbidden Colors
```

This is done in the D2 coloring exactly the same as it was done in [functorGreedyColor][3] in the distance-1 coloring code.


[1]: https://github.com/william76/kokkos-kernels/blob/Distance2GraphColor/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp#L65
[2]: https://github.com/william76/kokkos-kernels/blob/Distance2GraphColor/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp#L739
[3]: https://github.com/william76/kokkos-kernels/blob/Distance2GraphColor/src/graph/impl/KokkosGraph_GraphColor_impl.hpp#L1843

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
```diff
@@ -41,898 +41,1453 @@
...
+    void resolveConflicts(nnz_lno_t _nv,
+                          const_lno_row_view_t xadj_,
+                          adj_view_t adj_,
+                          const_clno_row_view_t t_xadj_,
+                          const_clno_nnz_view_t t_adj_,
+                          color_view_type vertex_colors_,
+                          nnz_lno_temp_work_view_t current_vertexList_,
+                          size_type current_vertexListLength_)
+    {
+        color_t *forbidden = new color_t[_nv];
```
srajama1: My comment was about this line. _nv is #vertices, right ?

srajama1:
@william76 See my comment about the conflict resolution phase. Also, you cannot follow the one kernel in d-1 coloring. That was written as an example. The most robust and performance friendly is the one w/ color sets. For e.g: what if #colors is more than 64 ?

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
william76: Ah, ok, I see what you're looking at now... 

So in [resolveConflicts][2]:
```c++
    template<typename adj_view_t>
    void resolveConflicts(nnz_lno_t _nv,
                          const_lno_row_view_t xadj_,
                          adj_view_t adj_,
                          const_clno_row_view_t t_xadj_,
                          const_clno_nnz_view_t t_adj_,
                          color_view_type vertex_colors_,
                          nnz_lno_temp_work_view_t current_vertexList_,
                          size_type current_vertexListLength_)
    {
        color_t *forbidden = new color_t[_nv];
        nnz_lno_t vid      = 0;
        nnz_lno_t end      = _nv;
```

The function `resolveConflicts` is called after the iterative loop exits up on [line 310][1]:

```c++
// ------------------------------------------
// clean up in serial (resolveConflicts)
// ------------------------------------------
if(numUncolored > 0)
{
    this->resolveConflicts(this->nv, this->xadj, this->adj, this->t_xadj, this->t_adj, colors_out, current_vertexList, current_vertexListLength);
}
```
This is done in serial to clean up any issues that remain after the parallel coloring is done so in that case, the forbidden array does need to be O(V).  

From test runs on the Laplace_285 dataset that I sent you last week, `resolveConflicts` takes about 0.0008 seconds total across 10 runs on my laptop.  This was the "Total Time RC" column on the tables.

[1]: https://github.com/william76/kokkos-kernels/blob/328fede3b7bfc9ca97e132e912fdec736ad3f930/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp#L305-L311
[2]: https://github.com/william76/kokkos-kernels/blob/328fede3b7bfc9ca97e132e912fdec736ad3f930/src/graph/impl/KokkosGraph_Distance2Color_impl.hpp#L538-L551

william76:
@srajama1 
The number of colors aren't hard-wired to 64, it just iteratively looks at 64 colors at a time.  There's an offset.

For the first iteration through:
- We try colors 0..63 first (note: color 0 is special and means the vertex is uncolored).
- We look at the distance-2 neighbors and flip bits in the array from false to true corresponding to the color (i.e., if we find a neighbor with color 17 then we flip `forbidden[17]` to true).
- After we look at all the distance-2 neighbors, we scan through `forbidden[]` and grab the first position that is still false and assign that color to vid.

If there are more than 64 colors, we go to a new iteration to check colors 64 to 127 using an offset.  We have to loop back around through the vertex's neighbors but we update `offset` to 64, so the next pass we're looking to find colors that are in the range 64 to 127.  This can become painful if we have more than 62 colors since we're going to have to loop back through the neighbors but it's a tradeoff to accommodate the memory constraints of the GPU.

For example, in the functor for greedyColor_VB:

```c++
color_t offset = 0;

while(!foundColor)
{
    // initialize
    for(int j = 0; j < VB_D2_COLORING_FORBIDDEN_SIZE; j++) { forbidden[j] = false; }

    // Color 0 is special (no color) so mask it out on the first iteration.
    if(offset == 0)
    {
        forbidden[0] = true;
    }

    // Check neighbors, fill forbidden array
    for(size_type vid_d1_adj = _idx(vid); vid_d1_adj < _idx(vid + 1); vid_d1_adj++)
    {
        nnz_lno_t vid_d1 = _adj(vid_d1_adj);

        for(size_type vid_d2_adj = _t_idx(vid_d1); vid_d2_adj < _t_idx(vid_d1 + 1); vid_d2_adj++)
        {
            nnz_lno_t vid_d2 = _t_adj(vid_d2_adj);

            // Skip distance-2-self-loops
            if(vid_d2 == vid || vid_d2 >= nv)
            {
                continue;
            }

            color_t c = _colors(vid_d2);
            if((c >= offset) && (c - offset < VB_D2_COLORING_FORBIDDEN_SIZE))
            {
                forbidden[c - offset] = true;
            }
        }
    }

    // color vertex i with smallest available color (firstFit)
    for(int c = 0; c < VB_D2_COLORING_FORBIDDEN_SIZE; c++)
    {
        if(!forbidden[c])
        {
            _colors(vid) = offset + c;
            foundColor   = true;
            break;
        }
    }      // for c...
    offset += VB_D2_COLORING_FORBIDDEN_SIZE;
}      // while !foundColor
```

srajama1:
Why is this closed ?

william76:
@srajama1 I closed it to do some cleanup on my branch and was going to create a new WIP PR.

srajama1:
Got it thanks !

Issue #264, opened 2018-06-19
enabling cuda lambda by default


lucbv:
This follows a discussion that I had with @srajama1 and @crtrott last week.
The basic questions was: are we OK with the assumption that lambdas are used in the source code of the library which requires us to compile with the option `enable_lambda`.

With the newer version of CUDA it seems that this is not a problem, so I am proposing to modify `generate_makefile.bash` to reflect that.

Adding @ndellingwood here since we also talked about it today

PR #265, opened 2018-06-19
Require cuda lambda

lucbv:
Small modification to the makefile generator that will automatically enable lambda functions for cuda.
This should take care of #264 I think?

### DIFF
--- a/scripts/generate_makefile.bash
+++ b/scripts/generate_makefile.bash
@@ -6,6 +6,7 @@ KOKKOS_DO_EXAMPLES="1"
 
 KOKKOSKERNELS_OPTIONS="eti-only"
 KOKKOSKERNELS_ENABLE_TPLS=""
+KOKKOS_CUDA_OPT="enable_lambda"
 
 while [[ $# > 0 ]]
 do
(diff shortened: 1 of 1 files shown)


srajama1:
@crtrott @ndellingwood Any objections ?

@nmhamster : Do you foresee any problems in enabling lambdas by default ? I believe lambdas should be supported by the vendors, but if this is going to restrict us in current or future platforms, we can write functors.

ndellingwood:
I think this would be nice and Trilinos uses lambdas pretty heavily, though it's worth pointing out lamdas are still considered experimental with nvcc (hence the flag to enable lambdas) which I believe is why they have not been enabled by default. 
@crtrott does support seem to be stable enough in cuda 8 and 9 that we can start testing with lambdas by default?

lucbv:
@ndellingwood I put this PR at the last minute yesterday but I wanted to get it up so I wouldn't forget about it... I still need to test it.

ndellingwood:
@lucbv thanks for the update, just wanted to check :)

lucbv:
I just saw that I never closed this, I had a discussion with @srajama1 and it seems that we are not ready yet for lambda offload yet so will close this PR.

PR #274, opened 2018-07-24
TPL Support for BLAS functions (nrm1, nrminf, scal, gemv) using CuBLAS (Issue #247)

vqd8a:
Adding cuBLAS TPL Support for nrm1, nrminf, scal, gemv.


mhoemmen:
You had me all excited there but it's just a test commit; ah well ;-)

vqd8a:
@mhoemmen Pls wait :-)

mhoemmen:
@csiefer2 and @jjellio also have an interest in kokkos-kernels having better BLAS support.

src/impl/tpls/KokkosBlas1_nrm1_tpl_spec_decl.hpp
```diff
@@ -198,4 +198,148 @@ KOKKOSBLAS1_CNRM1_TPL_SPEC_DECL_BLAS( Kokkos::LayoutLeft, Kokkos::HostSpace, fal
...
+  \
+  static void nrm1 (RV& R, const XV& X) \
+  { \
+    const size_type numElems = X.extent(0); \
+    if (numElems < static_cast<size_type> (INT_MAX)) { \
+      nrm1_print_specialization<RV,XV>(); \
+      const int N = static_cast<int> (numElems); \
+      constexpr int one = 1; \
+      KokkosBlas::Impl::CudaBlasSingleton & s = KokkosBlas::Impl::CudaBlasSingleton::singleton(); \
+      cublasDasum(s.handle, N, X.data(), one, &R()); \
```
mhoemmen: `R.data()` should also work just fine.

src/impl/tpls/KokkosBlas1_nrminf_tpl_spec_decl.hpp
```diff
@@ -208,4 +208,166 @@ KOKKOSBLAS1_CNRMINF_TPL_SPEC_DECL_BLAS( Kokkos::LayoutLeft, Kokkos::HostSpace, f
...
+  typedef Kokkos::View<float, LAYOUT, Kokkos::HostSpace, \
+                       Kokkos::MemoryTraits<Kokkos::Unmanaged> > RV; \
+  typedef Kokkos::View<const float*, LAYOUT, Kokkos::Device<ExecSpace, MEMSPACE>, \
+                       Kokkos::MemoryTraits<Kokkos::Unmanaged> > XV; \
+  typedef typename XV::size_type size_type; \
+  \
+  static void nrminf (RV& R, const XV& X) \
+  { \
+    const size_type numElems = X.extent(0); \
+    if (numElems == 0) { R() = 0.0f; return; } \
```
mhoemmen: What if R is a CudaSpace View?  You're writing to it on host here.  Instead, you could do this:
```c++
Kokkos::deep_copy (R, 0.0);
```

src/impl/tpls/KokkosBlas1_nrminf_tpl_spec_decl.hpp
```diff
@@ -208,4 +208,166 @@ KOKKOSBLAS1_CNRMINF_TPL_SPEC_DECL_BLAS( Kokkos::LayoutLeft, Kokkos::HostSpace, f
...
+                       Kokkos::MemoryTraits<Kokkos::Unmanaged> > RV; \
+  typedef Kokkos::View<const Kokkos::complex<double>*, LAYOUT, Kokkos::Device<ExecSpace, MEMSPACE>, \
+                       Kokkos::MemoryTraits<Kokkos::Unmanaged> > XV; \
+  typedef typename XV::size_type size_type; \
+  typedef Kokkos::Details::InnerProductSpaceTraits<Kokkos::complex<double>> IPT; \
+  \
+  static void nrminf (RV& R, const XV& X) \
+  { \
+    const size_type numElems = X.extent(0); \
+    if (numElems == 0) { R() = 0.0; return; } \
```
mhoemmen: What if R is a CudaSpace View?  You're writing to it on host here.  Instead, you could do this:
```c++
Kokkos::deep_copy (R, 0.0);
```

src/impl/tpls/KokkosBlas1_nrminf_tpl_spec_decl.hpp
```diff
@@ -208,4 +208,166 @@ KOKKOSBLAS1_CNRMINF_TPL_SPEC_DECL_BLAS( Kokkos::LayoutLeft, Kokkos::HostSpace, f
...
+    if (numElems < static_cast<size_type> (INT_MAX)) { \
+      nrminf_print_specialization<RV,XV>(); \
+      const int N = static_cast<int> (numElems); \
+      constexpr int one = 1; \
+      int idx; \
+      KokkosBlas::Impl::CudaBlasSingleton & s = KokkosBlas::Impl::CudaBlasSingleton::singleton(); \
+      cublasIzamax(s.handle, N, reinterpret_cast<const cuDoubleComplex*>(X.data()), one, &idx); \
+      Kokkos::View<Kokkos::complex<double>, LAYOUT, Kokkos::HostSpace > R_cplx("R_cplx",1); \
```
mhoemmen: This still calls `new`.  Why not make an unmanaged View for that one entry?
```c++
Kokkos::complex<double> R_cplx_val {0.0, 0.0};
Kokkos::View<Kokkos::complex<double>, LAYOUT, Kokkos::HostSpace> > R_cplx (&R_cplx_val);
```
More importantly, you shouldn't pass 1 in, because this is a rank-0 Kokkos::View.  (Disable deprecated code in Kokkos and see what happens.)  Even more importantly, if this View is managed, then you'll be using `Kokkos::HostSpace::execution_space` to initialize this View, which may not be what you want.  Instead, just use an unmanaged rank-0 View.

src/impl/tpls/KokkosBlas2_gemv_tpl_spec_decl.hpp
```diff
@@ -44,9 +44,456 @@
 #ifndef KOKKOSBLAS2_GEMV_TPL_SPEC_DECL_HPP_
 #define KOKKOSBLAS2_GEMV_TPL_SPEC_DECL_HPP_
 
+#ifdef KOKKOSKERNELS_ENABLE_TPL_BLAS
```
mhoemmen: BLAS libraries may use different underscore name mangling conventions, so you shouldn't hard-code the underscore.  See Teuchos for an example of how to handle this.

I am not a fan of all these `extern` declarations being exposed to header files that applications will include.  That will almost certainly collide with applications' `extern` declarations.  A better approach would be to wrap BLAS functions in C-style wrappers, and only expose the wrappers.  This would also help hide cuBLAS, MKL, and other TPL header file includes from kokkos-kernels' header files.

src/impl/tpls/KokkosBlas2_gemv_tpl_spec_decl.hpp
```diff
@@ -44,9 +44,456 @@
...
+        const YViewType& Y) { \
+    \
+    Kokkos::Profiling::pushRegion("KokkosBlas::gemv[TPL_BLAS,double]"); \
+    const int M = static_cast<int> (A.extent(0)); \
+    const int N = static_cast<int> (A.extent(1)); \
+    constexpr int one = 1; \
+    int strides[2]; \
+    bool A_is_lr = std::is_same<Kokkos::LayoutRight,LAYOUTA>::value; \
+    A.stride(strides); \
+    const int LDA = strides[A_is_lr?0:1]; \
```
mhoemmen: LDA must never be zero, even if e.g., the number of rows is zero.  See e.g., http://github.com/trilinos/Trilinos/issues/3235 .

mhoemmen:
The `extern` declarations exposed in a header file, with hard-coded underscore name mangling, will almost certainly cause build failures on some platforms.

src/impl/tpls/KokkosBlas2_gemv_tpl_spec_decl.hpp
```diff
@@ -457,7 +457,7 @@ struct GEMV< \
     \
     bool A_is_lr = std::is_same<Kokkos::LayoutRight,LAYOUTA>::value; \
     A.stride(strides); \
```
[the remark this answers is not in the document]
vqd8a: @mhoemmen I think yes, but aren't two strides are still needed?

src/impl/tpls/KokkosBlas2_gemv_tpl_spec_decl.hpp
vqd8a: Thanks @mhoemmen for your comments. This PR meant for supporting CUBLAS so I just followed the way to call BLAS libraries in other functions. I will fix this in another PR.

src/impl/tpls/KokkosBlas2_gemv_tpl_spec_decl.hpp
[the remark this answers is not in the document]
mhoemmen: Yes -- I was just thinking it would be nicer to get the first two strides by value, rather than using this old "write results to an array" interface.  It doesn't matter, though.

vqd8a:
Spot check on White:
../scripts/test_all_sandia cuda --spot-check --with-cuda-options=enable_lambda --with-tpls=cublas
```
Running on machine: white
Going to test compilers:  cuda/9.0.103
Testing compiler cuda/9.0.103
  Starting job cuda-9.0.103-Cuda_OpenMP-release
  PASSED cuda-9.0.103-Cuda_OpenMP-release
  Starting job cuda-9.0.103-Cuda_Serial-release
  PASSED cuda-9.0.103-Cuda_Serial-release
PASSED TESTS
cuda-9.0.103-Cuda_OpenMP-release build_time=773 run_time=387
cuda-9.0.103-Cuda_Serial-release build_time=750 run_time=393
```

srajama1:
@vqd8a : I merged it, however, can you create a wiki page for cuBLAS support and add what are all the calls we support before we forget it ?

vqd8a:
@srajama1 Sure, I can do that. Thanks.

Issue #276, opened 2018-08-02
CMake modification for TPLs.


kyungjoo-kim:
Currently, makefile has TPL mechanism to define corresponding KOKKOSKERNELS_ENABLE_TPL_BLAS. This setup mechanism is not available in cmake; thus, it does not work in Trilinos.

ndellingwood:
For reference, this issue completed by #277

kyungjoo-kim:
done

PR #277, opened 2018-08-02
KokkosKernels - cmake modifications for TPLs

kyungjoo-kim:
Reference issue #276
  - add BLAS, LAPACK, CUBLAS, CUSPARSE TPL dependence
  - KokkosBlas.hpp includes gemm
  - comment regarding to gtest path
    - when we use override src directory, the relative path to other
      package (kokkos) may not be correct.

This PR is necessary in order to remove Tpetra BLAS functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,6 +286,44 @@ IF (KokkosKernels_INST_OFFSET_SIZE_T)
   LIST(APPEND OFFSET_LIST "size_t")
 ENDIF()
 
+# ==================================================================
+# Enable Third Party Libraries
+# ==================================================================
+
+IF (TPL_ENABLE_BLAS)
+  SET(KOKKOSKERNELS_ENABLE_TPL_BLAS ${TPL_ENABLE_BLAS})
+ENDIF()
+IF (TPL_ENABLE_MKL)
+  SET(KOKKOSKERNELS_ENABLE_TPL_MKL ${TPL_ENABLE_MKL})
+ENDIF()
+
+IF(${Kokkos_ENABLE_Cuda})
+  # CUBLAS is ON by default when CUDA is enabled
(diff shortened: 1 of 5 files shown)


unit_test/CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ INCLUDE_DIRECTORIES(REQUIRED_DURING_INSTALLATION_TESTING ${CMAKE_CURRENT_SOURCE_
 INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR}/../test_common)
 INCLUDE_DIRECTORIES(REQUIRED_DURING_INSTALLATION_TESTING ${CMAKE_CURRENT_SOURCE_DIR}/../test_common)
 
+# Kyungjoo: when kokkoskernels and kokkos are symbolic link and their src directories are overrided
```
mhoemmen: What does it mean for "users to put kokkoskernels and kokkos at the same place"?

CMakeLists.txt
```diff
@@ -286,13 +286,54 @@ IF (KokkosKernels_INST_OFFSET_SIZE_T)
   LIST(APPEND OFFSET_LIST "size_t")
 ENDIF()
 
+# ==================================================================
+# Enable Third Party Libraries
+# ==================================================================
+
+IF (TPL_ENABLE_BLAS)
+  SET(KOKKOSKERNELS_ENABLE_TPL_BLAS ${TPL_ENABLE_BLAS})
```
mhoemmen: Do you intend for these variables to be visible outside of this package?  If so, you'll need to use `GLOBAL_SET` or something like that.

kyungjoo-kim: No these variables are internal only.

unit_test/CMakeLists.txt
mhoemmen: Thanks for explaining!

PR #278, opened 2018-08-07
KokkosKernels - fix the automatic tpl blas enabling logic

kyungjoo-kim:
Currently, tpl blas is automatically enabled when mkl or cublas is enabled.

This is wrong when cublas is only provided without host blas library. 

This fixes the problem.


kyungjoo-kim:
@ndellingwood would you merge this with your superpower ?

PR #279, opened 2018-08-09
Query positive infinity value of floating-point type using Kokkos::ArithTraits

dalg24:
Are `hasInfinity` and `infinity()` something you would consider add to `Kokkos::ArithTraits<T>`?

### DIFF
--- a/src/Kokkos_ArithTraits.hpp
+++ b/src/Kokkos_ArithTraits.hpp
@@ -849,7 +849,7 @@ template<class RealFloatType>
 class ArithTraits<std::complex<RealFloatType> > {
 public:
   //! Kokkos internally replaces std::complex with Kokkos::complex.
-  typedef ::Kokkos::complex<RealFloatType> val_type;
+  typedef std::complex<RealFloatType> val_type;
   typedef RealFloatType mag_type;
 
   static const bool is_specialized = true;
--- a/src/Kokkos_ArithTraits.hpp
+++ b/src/Kokkos_ArithTraits.hpp
@@ -427,6 +427,17 @@ public:
   //! The one value of T; the multiplicative identity.
   static KOKKOS_FORCEINLINE_FUNCTION T one ();
 
+  /// \brief True if this type T is capable of representing the
+  /// positive infinity as a distinct special value, as with
+  /// std::numeric_limits<T>::has_infinity.
+  static constexpr bool hasInfinity = false;
+
+  /// \brief Returns the special value "positive infinity", as
+  /// represented by the floating-point type T. Only meaningful if
+  /// KokkosArithTraits<T>::hasInfinity == true. Provides same
+  /// functionality as std::numeric_limits<T>::infinity().
+  static KOKKOS_FORCEINLINE_FUNCTION T infinity();
+
   /// \brief The minimum possible value of T.
   ///
   /// If T is a real floating-point type, then this is the minimum
@@ -670,6 +681,9 @@ public:
   static const bool is_integer = false;
   static const bool is_exact = false;
   static const bool is_complex = false;
+
+  static constexpr bool hasInfinity = true;
+  static KOKKOS_FORCEINLINE_FUNCTION float infinity() { return HUGE_VALF; }
 
   static KOKKOS_FORCEINLINE_FUNCTION bool isInf (const float x) {
     #ifndef __CUDA_ARCH__
@@ -1017,6 +1031,9 @@ public:
   static const bool is_exact = false;
   static const bool is_complex = false;
 
+  static constexpr bool hasInfinity = true;
+  static KOKKOS_FORCEINLINE_FUNCTION double infinity() { return HUGE_VAL; }
+
   static KOKKOS_FORCEINLINE_FUNCTION bool isInf (const val_type x) {
     #ifndef __CUDA_ARCH__
     using std::isinf;
@@ -1184,6 +1201,9 @@ public:
   static const bool is_integer = false;
   static const bool is_exact = false;
   static const bool is_complex = false;
+
+  static constexpr bool hasInfinity = true;
+  static KOKKOS_FORCEINLINE_FUNCTION long double infinity() { return HUGE_VALL; }
 
   static bool isInf (const val_type& x) {
     #ifndef __CUDA_ARCH__
--- a/src/Kokkos_ArithTraits.hpp
+++ b/src/Kokkos_ArithTraits.hpp
@@ -1948,6 +1948,9 @@ public:
   static const bool is_exact = true;
   static const bool is_complex = false;
 
+  static constexpr bool hasInfinity = false;
+  static KOKKOS_FORCEINLINE_FUNCTION val_type infinity() { return 0; }
+
   static KOKKOS_FORCEINLINE_FUNCTION bool isInf (const val_type ) {
     return false;
   }
@@ -2090,6 +2093,9 @@ public:
   static const bool is_exact = true;
   static const bool is_complex = false;
 
+  static constexpr bool hasInfinity = false;
+  static KOKKOS_FORCEINLINE_FUNCTION val_type infinity() { return 0; }
+
   static KOKKOS_FORCEINLINE_FUNCTION bool isInf (const val_type ) {
     return false;
   }
@@ -2208,6 +2214,9 @@ public:
   static const bool is_integer = true;
   static const bool is_exact = true;
   static const bool is_complex = false;
+
+  static constexpr bool hasInfinity = false;
+  static KOKKOS_FORCEINLINE_FUNCTION val_type infinity() { return 0; }
 
   static KOKKOS_FORCEINLINE_FUNCTION bool isInf (const val_type ) {
     return false;
@@ -2330,6 +2339,9 @@ public:
   static const bool is_integer = true;
   static const bool is_exact = true;
   static const bool is_complex = false;
+
+  static constexpr bool hasInfinity = false;
+  static KOKKOS_FORCEINLINE_FUNCTION val_type infinity() { return 0; }
 
   static KOKKOS_FORCEINLINE_FUNCTION bool isInf (const val_type ) {
     return false;
@@ -2467,6 +2479,9 @@ public:
   static const bool is_exact = true;
   static const bool is_complex = false;
 
+  static constexpr bool hasInfinity = false;
+  static KOKKOS_FORCEINLINE_FUNCTION val_type infinity() { return 0; }
+
   static KOKKOS_FORCEINLINE_FUNCTION bool isInf (const val_type ) {
     return false;
   }
@@ -2594,6 +2609,9 @@ public:
   static const bool is_integer = true;
   static const bool is_exact = true;
   static const bool is_complex = false;
+
+  static constexpr bool hasInfinity = false;
+  static KOKKOS_FORCEINLINE_FUNCTION val_type infinity() { return 0; }
 
   static KOKKOS_FORCEINLINE_FUNCTION bool isInf (const val_type ) {
     return false;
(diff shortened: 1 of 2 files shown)


dalg24:
PR is incomplete.  Specialization for other types are missing and I did not added any test yet.  I want to make sure there is interest in this feature before I spend more time on it.

srajama1:
@dalg24 Yes, we would add it. Thanks for contributing ! Let us know when the PR is ready to be reviewed.

dalg24:
Ready for your review @srajama1

test_common/Test_Common_ArithTraits.hpp
```diff
@@ -371,6 +371,19 @@ class ArithTraitsTesterBase {
       success = 0;
     }
 
+    if (AT::hasInfinity) {
+      if (! AT::isInf (AT::infinity())) {
+        out << "AT::isInf (inf) != true" << endl;
+        success = 0;
+      }
+    } else {
```
dalg24: I'd rather have had a static assertion
```C++
    static_assert(
      AT::hasInfinity && (!AT::is_integer && !AT::is_complex) ||
        !AT::hasInfinity && (AT::is_integer || !AT::is_complex),
      "real floating-point type cannot represent positive infinity value");
```
rather than that `else` branch but it is not possible because `is_complex` and `is_integer` are not `constexpr`.

src/Kokkos_ArithTraits.hpp
```diff
@@ -427,6 +427,21 @@ class ArithTraits {
   //! The one value of T; the multiplicative identity.
   static KOKKOS_FORCEINLINE_FUNCTION T one ();
 
+  /// \brief True if this type T is capable of representing the
+  /// positive infinity as a distinct special value, as with
+  /// std::numeric_limits<T>::has_infinity.
+  static constexpr bool hasInfinity = false;
```
mhoemmen: It would be better for this to be spelled `has_infinity`, for compatibility with `std::numeric_limits`: https://en.cppreference.com/w/cpp/types/numeric_limits

src/Kokkos_ArithTraits.hpp
```diff
@@ -427,6 +427,21 @@ class ArithTraits {
...
+
+  /// \brief Returns the special value "positive infinity", as
+  /// represented by the floating-point type T. Only meaningful if
+  /// KokkosArithTraits<T>::hasInfinity == true. Provides same
+  /// functionality as std::numeric_limits<T>::infinity().
+  ///
+  /// \note Would have liked to mark it as constexpr but then would
```
mhoemmen: Does Kokkos pass along `KOKKOS_CXX_STANDARD` or `KOKKOS_ENABLE_CXX14` to code as macros?  If so, you could select on that macro to make this `constexpr` (but note CUDA issues).

src/Kokkos_ArithTraits.hpp
dalg24: Will change.  I had assumed it was the style to follow when looking at `isInf()`.

src/Kokkos_ArithTraits.hpp
dalg24: What CUDA issues are your referring to?

dalg24: And I just grepped for `KOKKOS_CXX_STANDARD` and `KOKKOS_ENABLE_CXX` and it looks like `KOKKOS_CXX_STANDARD` is not passed along but `KOKKOS_ENABLE_CXX14` is.  I am not sure the complexity added is worth it at the moment unless there is a real need for it in which case we probably should consider make other member constants and member functions of `ArithTraits` constexp as well.

src/Kokkos_ArithTraits.hpp
mhoemmen: @dalg24 Thanks! :D

src/Kokkos_ArithTraits.hpp
dalg24: This shows as outdated.  Not sure if you have seen it @mhoemmen.

dalg24:
Anything else you would like me to change?

test_common/Test_Common_ArithTraits.hpp
mhoemmen: @dalg24 The main thing hindering `constexpr` here was CUDA not permitting `constexpr` values to be marked `__device__`.  I think that has been fixed, though you may have to select on the CUDA version in order to use `constexpr` without warnings, and/or use a special `nvcc` flag.

src/Kokkos_ArithTraits.hpp
```diff
@@ -840,6 +858,11 @@ class ArithTraits<std::complex<RealFloatType> > {
   static const bool is_exact = false;
   static const bool is_complex = true;
 
+  static constexpr bool has_infinity = false;
+  static KOKKOS_FORCEINLINE_FUNCTION std::complex<RealFloatType> infinity() {
```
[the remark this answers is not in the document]
dalg24: > return Kokkos::complex

Good catch.  Will change.

> If has_infinity is false, then you shouldn't have to implement this method.

Not sure about that.  The return value of `std:: numeric_limits<T>::infinity()` is `T()` is it has not been specialized.

> an issue with static constexpr values being used on device -- be sure to test with CUDA

Will look more into this.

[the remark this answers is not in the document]
dalg24: > > return Kokkos::complex
>
> Good catch. Will change.

Actually this is the specialization for `std::complex` so actually the `val_type` typedef is wrong...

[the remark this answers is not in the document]
dalg24: Or am I missing something?

src/Kokkos_ArithTraits.hpp
```diff
@@ -849,7 +849,7 @@ template<class RealFloatType>
 class ArithTraits<std::complex<RealFloatType> > {
 public:
   //! Kokkos internally replaces std::complex with Kokkos::complex.
-  typedef ::Kokkos::complex<RealFloatType> val_type;
+  typedef std::complex<RealFloatType> val_type;
```
dalg24: @mhoemmen was that intentional?  I am confused now because they are later specializations for `Kokkos::complex` in this file.

src/Kokkos_ArithTraits.hpp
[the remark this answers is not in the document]
dalg24: > > an issue with static constexpr values being used on device -- be sure to test with CUDA
> 
> Will look more into this.

Built the tests on a K40, ran `cuda.common_ArithTraits` and it is passing with the `constexpr`.

src/Kokkos_ArithTraits.hpp
mhoemmen: Don't do this.  Read the comment :)  It's intentional that `val_type` is `::Kokkos::complex<RealFloatType>`.  The point is that `val_type` works with CUDA and works as the result of Kokkos reductions.  It needs to be bitwise equivalent (*) to ArithTraits' template parameter.  There are specializations for both std::complex and Kokkos::complex, because users need to make Views of Kokkos::complex in order for their code to work on CUDA.  std::complex's methods lack the `__device__ __host__` qualifiers.

(*) std::complex technically could use a different representation than (real, imag).  However, we haven't had any problems assuming that this is its representation.

dalg24: Fair enough.  I reverted the change.

dalg24:
@mhoemmen 

1. Given that `std::numeric_limits` does not have a specialization for complex numbers and that 
`std::numeric_limits<T>::infinity()` return value is `T()` for non-specialized types, are you comfortable with my returning `val_type()` in complex number types specializations of `ArithTraits`?
2. Do you want me to remove the `constexpr` specifier?

dalg24:
@mhoemmen Done

dalg24:
> I think you can (and should) just omit the infinity method. It's confusing for the infinity method to exist, but to return 0 for types like short.

It depends.  If the intent with the `Kokkos::ArithTraits` class template is to provide the same functionality as `std::numeric_limits<T>`, then IMO you should have them because that's [how they are specified by the standard](https://en.cppreference.com/w/cpp/types/numeric_limits/infinity).  They are only meaningful if the value of `has_infinity` is `true` but the member function is defined regardless.
Note that the tests would not compile if we omit the specialization when `Kokkos::ArithTraits<T>::has_infinity == true` so we would need to find a way to only enable the tests for types that have the specialization.

mhoemmen:
@dalg24 wrote:
> then IMO you should have them because that's how they are specified by the standard.

Good point; see [numeric.limits] (Section 16.3.4 in the current draft n4762) and [numeric.special] (Section 16.3.4.2, "All members shall be provided for all specializations").  The default behavior if `has_infinity` is `false` is for `infinity()` to return `T()`.  In this case, it should return `val_type()`.

dalg24:
Anything else you'd like to change before this can be merged?

src/Kokkos_ArithTraits.hpp
```diff
@@ -1348,6 +1377,9 @@ class ArithTraits<__float128> {
   static const bool is_exact = false;
   static const bool is_complex = false;
 
+  static constexpr bool has_infinity = true;
+  static KOKKOS_FORCEINLINE_FUNCTION __float128 infinity() { return 1.0q / 0.0q; }
```
mhoemmen: Does `__float128` work with CUDA?  I'm pretty sure that it doesn't, but if it does, I would be pleasantly surprised.

dalg24: Nope.  Good catch :)

test_common/Test_Common_ArithTraits.hpp
```diff
@@ -377,9 +377,21 @@ class ArithTraitsTesterBase {
...
-       if ( ! (AT::is_integer || AT::is_complex)) {
-        out << "AT::has_infinity != true" << endl;
+       // We expect that real floating-point and complex numbers types had
+       // has_infinity == true.  Either we have an integer type or the
+       // ArithTraits class has not been specialized for that type.
+       if ( AT::is_integer ) {
+       // Return value of ArithTraits<T>::inifinity is only meaningfull when
+       // has_infinity == true but nevertheless let's check that the return
+       // value is 0 for integer types because that's how it is defined for
+       // std::numeric_limits::infinity() and how we intended to implement it.
```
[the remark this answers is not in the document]
dalg24: There is no specialization of the `std::numeric_traits` class for `std::complex<T>` :)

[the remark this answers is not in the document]
mhoemmen: @dalg24 oh oops never mind then :D 

Mainly I wanted to suggest that you don't actually have to test the value of `infinity()` in this case, merely that the function exists, doesn't throw, and returns some `val_type` value.  I'm OK with this, though :)

test_common/Test_Common_ArithTraits.hpp
```diff
@@ -395,6 +395,11 @@ class ArithTraitsTesterBase {
         success = 0;
        }
     }
+    if ( ! std::is_same< ScalarType, typename std::result_of<decltype(&AT::infinity)()>::type >::value )
```
dalg24: Could use a static assertion instead if you like.

mhoemmen: Hm, do we not test the case where `ScalarType` is `std::complex<R>`?  I guess not :-)

Why don't we just write `decltype(AT::infinity())` instead of the `std::result_of` expression?

dalg24: No good reason.  Will change.

mhoemmen: btw I'm happy with the PR -- I don't mean to make you do more work -- but I'll be happy to see further improvements and will gladly review them :-)

dalg24: Any reason for delaying the merge?

dalg24:
No more changes.  Thanks.

srajama1:
Thanks for the contribution to KK.

PR #286, opened 2018-09-06
KokkosBatched - change cuda get team size according to kokkos changes

kyungjoo-kim:
kokkos uses launch bounds and impl interface is changed. this PR follows the changes.

### DIFF
--- a/perf_test/batched/KokkosBatched_Test_Gemm_Cuda.cpp
+++ b/perf_test/batched/KokkosBatched_Test_Gemm_Cuda.cpp
@@ -364,11 +364,10 @@ namespace KokkosBatched {
             typedef Kokkos::TeamPolicy<DeviceSpaceType,ScheduleType,TeamTagV1> policy_type;
 
             typedef Functor<view_type,AlgoTagType,VectorLength> functor_type;
-            typedef Kokkos::Impl::ParallelFor<functor_type,policy_type,DeviceSpaceType> parallel_for_type;
-            typedef Kokkos::LaunchBounds<> launch_bounds_type;
 
+            // 128 is rough estimates
             const int team_size = 
-              Kokkos::Impl::cuda_get_opt_block_size<parallel_for_type,launch_bounds_type>(functor_type(), VectorLength, 0, 0)/VectorLength;
+              policy_type(N/32, Kokkos::AUTO, VectorLength).team_size_recommended(functor_type(), Kokkos::ParallelForTag());
               
             const policy_type policy(N/team_size, team_size, VectorLength);
             for (int iter=iter_begin;iter<iter_end;++iter) {
@@ -429,10 +428,7 @@ namespace KokkosBatched {
           double tavg = 0, tmin = tmax;
           {
             typedef Kokkos::TeamPolicy<DeviceSpaceType,ScheduleType,TeamTagV2> policy_type;
-
             typedef Functor<view_type,AlgoTagType,VectorLength> functor_type;
-            typedef Kokkos::Impl::ParallelFor<functor_type,policy_type,DeviceSpaceType> parallel_for_type;
-            typedef Kokkos::LaunchBounds<> launch_bounds_type;
 
             const int 
               is_blocked_algo = (std::is_same<AlgoTagType,Algo::Gemm::Blocked>::value), 
@@ -442,8 +438,9 @@ namespace KokkosBatched {
             const int 
               mblk = is_blocked_algo ? (BlkSize/mb + mp) : BlkSize;
 
-            const int max_cuda_blocksize = Kokkos::Impl::cuda_get_max_block_size<parallel_for_type,launch_bounds_type>(functor_type(), VectorLength, 0, 0);
-            const int team_size = std::min(std::max(mblk*mblk,4), max_cuda_blocksize/VectorLength);
+            const int max_team_size = 
+              policy_type(N, Kokkos::AUTO, VectorLength).team_size_max(functor_type(), Kokkos::ParallelForTag());
+            const int team_size = std::min(std::max(mblk*mblk,4), max_team_size);
 
             policy_type policy(N, team_size, VectorLength);
             for (int iter=iter_begin;iter<iter_end;++iter) {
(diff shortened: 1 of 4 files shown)


crtrott:
You don't need the launch bounds. And also I was working on this too :-) But combined with some other changes.

perf_test/batched/KokkosBatched_Test_Gemm_Cuda.cpp
```diff
@@ -365,9 +365,10 @@ namespace KokkosBatched {
 
             typedef Functor<view_type,AlgoTagType,VectorLength> functor_type;
             typedef Kokkos::Impl::ParallelFor<functor_type,policy_type,DeviceSpaceType> parallel_for_type;
-              
+            typedef Kokkos::LaunchBounds<> launch_bounds_type;
+
             const int team_size = 
-              Kokkos::Impl::cuda_get_opt_block_size<parallel_for_type>(functor_type(), VectorLength, 0, 0)/VectorLength;
+              Kokkos::Impl::cuda_get_opt_block_size<parallel_for_type,launch_bounds_type>(functor_type(), VectorLength, 0, 0)/VectorLength;
```
ndellingwood: @kyungjoo-kim changes look fine but I think Kokkos already sets the launch bounds template parameter by default to `Kokkos::LaunchBounds<>` so this shouldn't be necessary unless you want to explicitly control this. Did you run into compilation errors without using this?

kyungjoo-kim: No. kokko does not put a default launch bounds and that is why I make these changes. If kokkos has a plan to do so, this PR should be modified. Anyway at this moment, kokkos does not have a default template parameter.

kyungjoo-kim: at least what I see from the kokkos develop branch. 

https://github.com/kokkos/kokkos/blob/f7065cacbec39923026561407b35cceabf0d0720/core/src/Cuda/Kokkos_Cuda_Internal.hpp#L58

as this under impl space, I  just follow the changes made here.

crtrott: I don't get it. Yeah we are not putting LaunchBounds<> in by default, but omitting them has the same effect as adding LaunchBounds<>.

ndellingwood:
@kyungjoo-kim my mistake, I thought the routine had default template parameter but it does not in the current develop branch of Kokkos:
[cuda_get_opt_block_size routine](https://github.com/kokkos/kokkos/blob/f7065cacbec39923026561407b35cceabf0d0720/core/src/Cuda/Kokkos_Cuda_Internal.hpp#L197-L201)

@crtrott will the stuff you are working on change this behavior, or can I go ahead and approve these changes?

perf_test/batched/KokkosBatched_Test_Gemm_Cuda.cpp
kyungjoo-kim: This is the error occurred while compiling and it requires launchboudns.
```
error: no instance of function template "Kokkos::Impl::cuda_get_max_block_size" matches the argument list 
```

crtrott:
Ah sorry I get it now. My mind is still a bit mushy from coming home late. You shouldn't use those functions at all if you can avoid it ...

My changes looked more like this: 
```
--- a/perf_test/batched/KokkosBatched_Test_Gemm_Cuda.cpp
+++ b/perf_test/batched/KokkosBatched_Test_Gemm_Cuda.cpp
@@ -366,9 +366,9 @@ namespace KokkosBatched {
             typedef Functor<view_type,AlgoTagType,VectorLength> functor_type;
             typedef Kokkos::Impl::ParallelFor<functor_type,policy_type,DeviceSpaceType> parallel_for_type;
               
-            const int team_size = 
-              Kokkos::Impl::cuda_get_opt_block_size<parallel_for_type>(functor_type(), VectorLength, 0, 0)/VectorLength;
-              
+            const int team_size = policy_type(N/128, Kokkos::AUTO, VectorLength).
+                                     team_size_recommended(functor_type(),Kokkos::ParallelForTag());
+
             const policy_type policy(N/team_size, team_size, VectorLength);
             for (int iter=iter_begin;iter<iter_end;++iter) {
               // flush
@@ -440,7 +440,9 @@ namespace KokkosBatched {
             const int 
               mblk = is_blocked_algo ? (BlkSize/mb + mp) : BlkSize;
 
-            const int max_cuda_blocksize = Kokkos::Impl::cuda_get_max_block_size<parallel_for_type>(functor_type(), VectorLength, 0, 0);
+              const int max_cuda_blocksize = policy_type(N/128, Kokkos::AUTO, VectorLength).
+                team_size_max(functor_type(),Kokkos::ParallelForTag());
```

perf_test/batched/KokkosBatched_Test_Gemm_Cuda.cpp
crtrott: See below: you shouldn't really use the implementation functions, you should go through the public interface.

ndellingwood: I didn't realize the `cuda_get_opt_block_size` routine should deduce the `LaunchBounds` template parameter from the return type `CudaGetOptBlockSize<...>::get_block_size(...)` (which is partial specialized on LaunchBounds) - so if a LauchBounds template parameter is not provided to the `cuda_get_opt_block_size` then it should hit either the partial specialization of line 205 or 235, where the partial specialization is for `LaunchBounds<>`.

kyungjoo-kim:
A question. What is the reason that you use N/128 for league size ?

crtrott:
Doesn't really matter too much. But you may want to use a league_size which is in the right ballpark. For example we may increase the recommended team_size if you have very few teams available, so that the total number of threads is big enough.

ndellingwood:
@crtrott @kyungjoo-kim so it sounds like the suggested change is to use `cuda_get_max_block_size` which is not in the public interface:
[cuda_get_max_block_size signature](https://github.com/kokkos/kokkos/blob/f7065cacbec39923026561407b35cceabf0d0720/core/src/Cuda/Kokkos_Cuda_Internal.hpp#L59)
I'm assuming the LaunchBounds template parameter should be deduced from the return type `CudaGetMaxBlockSize<...>::get_block_size(...)` if not provided, correct?

kyungjoo-kim:
Separately from this PR, do we have a plan to improve or develop the heuristics to determine team size ? Currently you have max or optimal team size but I often see that manual setting perform quite better than what is recommended or max size.

srajama1:
Folks, can we document the correct way for launch bounds usage in a wiki ?

kyungjoo-kim:
@crtrott I changed according to your comment. Could you approve this and merge into develop ?

Issue #289, opened 2018-09-11
Failing spgemm unit test with MKL


ndellingwood:
Sample output with either intel/17 or intel/18.2.199 on Bowman with OpenMP backend:

```
[ RUN      ] openmp.sparse_spgemm_double_int_int_TestExecSpace
/home/ndellin/kokkos-kernels/unit_test/sparse/Test_Sparse_spgemm.hpp:366: Failure
Value of: (failed == is_expected_to_fail)
  Actual: false
Expected: true
[  FAILED  ] openmp.sparse_spgemm_double_int_int_TestExecSpace (9309 ms)
[----------] 1 test from openmp (9309 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (9311 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] openmp.sparse_spgemm_double_int_int_TestExecSpace
```

ndellingwood:
Think I have this chased down, looks like there is a macro defined in the config file `HAVE_KOKKOSKERNELS_MKL` that never gets set but is expected in the failing unit test at this line: 
[line 311](https://github.com/kokkos/kokkos-kernels/blob/f1f6a21475b99a2f6be096d75c6bdb0ed5285dab/unit_test/sparse/Test_Sparse_spgemm.hpp#L311)

ndellingwood:
Updating that line to instead check that 
`#if !defined(HAVE_KOKKOSKERNELS_MKL) && !defined(KOKKOSKERNELS_ENABLE_TPL_BLAS)`
seems to fix it, will put in PR shortly. Will probably combine with update requested by #268

srajama1:
I don't think  !defined(KOKKOSKERNELS_ENABLE_TPL_BLAS) is needed for this test.

ndellingwood:
@srajama1  To have the expected setting for `is_expected_to_fail` the `HAVE_KOKKOSKERNELS_MKL` should be defined. Since it is never set in the CMakeLists.txt file I thought adding the check for `!defined(KOKKOSKERNELS_ENABLE_TPL_BLAS)` was the right thing to do. Should the CMakeLists.txt file be changed to define `HAVE_KOKKOSKERNELS_MKL`  when `KOKKOSKERNELS_ENABLE_TPL_BLAS` is defined?

ndellingwood:
[...]

Issue #268, opened 2018-07-09
MKL - guard minor version for MKL v. 18


ndellingwood:

srajama1:
@ndellingwood : It is better to do this before next update.

ndellingwood:
PR #290 addresses this.

PR #290, opened 2018-09-11
MKL fixes: Add extra macro checks

ndellingwood:
Address issue #289 - fix macro use for unit test,
and #268 - update macro check for INTEL 18 to also check that update is 2,
needed due to changes in MKL interface with updated version.

### DIFF
--- a/src/sparse/impl/KokkosSparse_spgemm_mkl2phase_impl.hpp
+++ b/src/sparse/impl/KokkosSparse_spgemm_mkl2phase_impl.hpp
@@ -169,7 +169,7 @@ void mkl2phase_symbolic(
     }
 #endif
 
-#if __INTEL_MKL__ >= 2018
+#if __INTEL_MKL__ == 2018 && __INTEL_MKL_UPDATE__ == 2
     MKL_INT mklm = m, mkln = n;
     double *mynullptr = NULL;
 
@@ -261,6 +261,10 @@ void mkl2phase_symbolic(
     if (SPARSE_STATUS_SUCCESS != mkl_sparse_destroy (C)){
       throw std::runtime_error ("Error at mkl_sparse_destroy C\n");
     }
+#elif __INTEL_MKL__ == 2018 && __INTEL_MKL_UPDATE__ < 2
+    throw std::runtime_error ("MKL version 18 must have update 2 - use intel/18.2.xyz compiler\n");
+#else
+    throw std::runtime_error ("MKL versions > 18 are not yet tested/supported\n");
 #endif
 
   }
@@ -400,7 +404,7 @@ void mkl2phase_symbolic(
       }
 #endif
 
-#if __INTEL_MKL__ >= 2018
+#if __INTEL_MKL__ == 2018 && __INTEL_MKL_UPDATE__ == 2
       value_type *a_ew = const_cast<value_type*>(valuesA.data());
       value_type *b_ew = const_cast<value_type*>(valuesB.data());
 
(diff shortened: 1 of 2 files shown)


ndellingwood:
Tested on Bowman with OpenMP backend, MKL enabled and intel/17.0.2 and intel/18.2.199

src/sparse/impl/KokkosSparse_spgemm_mkl2phase_impl.hpp
```diff
@@ -529,6 +533,10 @@ void mkl2phase_symbolic(
       if (SPARSE_STATUS_SUCCESS != mkl_sparse_destroy (C)){
         throw std::runtime_error ("Error at mkl_sparse_destroy C\n");
       }
+#elif __INTEL_MKL__ == 2018 && __INTEL_MKL_UPDATE__ < 2
+      throw std::runtime_error ("MKL version 18 must have update 2 - use intel/18.2.xyz compiler\n");
```
srajama1: Minor nit-pick - pls remove the word compiler and instead say "Intel MKL version 18.2. or more"

src/sparse/impl/KokkosSparse_spgemm_mkl2phase_impl.hpp
```diff
@@ -169,7 +169,7 @@ void mkl2phase_symbolic(
     }
 #endif
 
-#if __INTEL_MKL__ >= 2018
+#if __INTEL_MKL__ == 2018 && __INTEL_MKL_UPDATE__ == 2
```
srajama1: I believe greater than or equal to 2 should be fine here for now. If something goes wrong in 3 we will come back and fix it.

src/sparse/impl/KokkosSparse_spgemm_mkl2phase_impl.hpp
```diff
@@ -261,6 +261,10 @@ void mkl2phase_symbolic(
     if (SPARSE_STATUS_SUCCESS != mkl_sparse_destroy (C)){
       throw std::runtime_error ("Error at mkl_sparse_destroy C\n");
     }
+#elif __INTEL_MKL__ == 2018 && __INTEL_MKL_UPDATE__ < 2
+    throw std::runtime_error ("MKL version 18 must have update 2 - use intel/18.2.xyz compiler\n");
```
srajama1: See below

src/sparse/impl/KokkosSparse_spgemm_mkl2phase_impl.hpp
```diff
@@ -400,7 +404,7 @@ void mkl2phase_symbolic(
       }
 #endif
 
-#if __INTEL_MKL__ >= 2018
+#if __INTEL_MKL__ == 2018 && __INTEL_MKL_UPDATE__ == 2
```
srajama1: See above >= 2 should be ok.

src/sparse/impl/KokkosSparse_spgemm_mkl2phase_impl.hpp
ndellingwood: Fixed

src/sparse/impl/KokkosSparse_spgemm_mkl2phase_impl.hpp
ndellingwood: @srajama1 I don't mind changing back to >=, but thought it would be helpful for earlier/easier detection of mkl api changes to force an error of intel/18.2.xyz is not used, since the api changed not just between version (17 to 18) but also updates within a version (update 0 and 1 to 2, for example). If we error out with untested versions we can verify early on if the api changed, and based on intel 18 experience this seems quite possible.

ndellingwood:
As a followup we discussed getting a Jenkins build setup with MKL enabled, adding comment here as reminder.

PR #300, opened 2018-09-25
test_all_sandia: update compiler lists

ndellingwood:
Remove shepard and sullivan

### DIFF
--- a/scripts/test_all_sandia
+++ b/scripts/test_all_sandia
@@ -17,37 +17,18 @@ if [[ "$HOSTNAME" =~ (white|ride).* ]]; then
   module load git
 fi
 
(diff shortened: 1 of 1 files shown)


ndellingwood:
Updated the compiler lists to match those of the kokkos script.

ndellingwood:
@crtrott @srajama1 with these updates to compilers on White (to match testing done by Kokkos) the Cuda tests will fail due to `-Werror` flags - gcc/7.2 is pickier about `int-in-bool-context` occurrences. May I remove `-Werror` from the cuda flags, or would you prefer that we test with gcc/6.4 and cuda/9.2?

ndellingwood:
Updated the script to remove `-Werror` from the cuda flags.

Issue #281, opened 2018-08-22
Build failure with clang 3.9.0


lucbv:
I have encounter the following error while compiling with `clang-3.9.0`:
```
/home/lberge/Research/kk_lucbv_builds/clang_3_9_0/kokkos/install/include/Kokkos_View.hpp:1953:7: error: static_assert failed
      "Incompatible View copy construction"
      static_assert( Mapping::is_assignable , "Incompatible View copy construction" );
      ^              ~~~~~~~~~~~~~~~~~~~~~~
/home/lberge/Research/kk_lucbv_builds/clang_3_9_0/install/include/KokkosSparse_spgemm_impl_symbolic.hpp:2317:4: note: in instantiation of function template
      specialization 'Kokkos::View<unsigned int *, Kokkos::HostSpace>::View<const unsigned int *, Kokkos::LayoutRight, Kokkos::Device<Kokkos::OpenMP,
      Kokkos::HostSpace>, Kokkos::MemoryTraits<0> >' requested here
                        row_mapA_,
                        ^
/home/lberge/Research/kk_lucbv_builds/clang_3_9_0/install/include/KokkosSparse_spgemm_impl_compression.hpp:957:41: note: in instantiation of function template
      specialization 'KokkosSparse::Impl::KokkosSPGEMM<KokkosKernels::Experimental::KokkosKernelsHandle<unsigned int, int, int, Kokkos::OpenMP,
      Kokkos::HostSpace, Kokkos::HostSpace>, Kokkos::View<unsigned int *, Kokkos::HostSpace>, Kokkos::View<int *, Kokkos::HostSpace>, Kokkos::View<int *,
      Kokkos::HostSpace>, Kokkos::View<unsigned int *, Kokkos::HostSpace>, Kokkos::View<int *, Kokkos::HostSpace>, Kokkos::View<int *, Kokkos::HostSpace>
      >::getMaxRoughRowNNZ<Kokkos::View<const unsigned int *, Kokkos::LayoutRight, Kokkos::Device<Kokkos::OpenMP, Kokkos::HostSpace>, Kokkos::MemoryTraits<0>
      >, Kokkos::View<const int *, Kokkos::LayoutRight, Kokkos::Device<Kokkos::OpenMP, Kokkos::HostSpace>, Kokkos::MemoryTraits<0> >, Kokkos::View<unsigned
[...]

PR #301, opened 2018-09-25
Fix issue #281: template parameters of functor contained in class is shadowed

lucbv:
Renaming template parameters of KokkosSPGEMM::PredicMaxRowNNZ to avoid shadowing that leads to a bad copy constructor call.
I am not too sure why gcc is allowing this to go through but clang is clearly less appreciative.


lucbv:
@crtrott this is one of the issues we talked about earlier see issue #281 with the changes made the compiler error disappears.

Issue #298, opened 2018-09-20
no more std::make_pair in a functor.


kyungjoo-kim:
```
/ascldap/users/kyukim/Work/lib/kokkoskernels/master/unit_test/blas/Test_Blas1_team_dot.hpp(66):                                                                                                              
warning: calling a constexpr __host__ function("make_pair") from a __host__ __device__ function("operator()") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.    
          detected during instantiation of "int test_team_dot<ScalarA,ScalarB,Device>() [with ScalarA=double, ScalarB=double, Device=Kokkos::Serial]"        
```

ndellingwood:
Addressed by @vqd8a with PR #302

PR #302, opened 2018-09-26
Issue 298 std::make_pair in functors

vqd8a:
#298 Replaced std::make_pair with Kokkos::make_pair in team-based BLAS unit tests.


vqd8a:
@ndellingwood Yes, I tested. They can be compiled and run successfully with Cuda.

ndellingwood:
Great, thanks @vqd8a !

PR #308, opened 2018-10-01
Update trilinos integration scripts (2)

ndellingwood:
Make default Kokkos branch "develop" if none given


ndellingwood:
@srajama1 @kyungjoo-kim @crtrott could one of you take a look at this PR and merge?
Two changes: 
1) If no arguments are passed to the script then kokkos defaults to the master branch while everything else is on the develop branch causing issues.
2) The math lib `m` was added to the `LAPACK_LIBRARIES` based on a suggestion from Si to get past some link errors with `*ggsvd`

PR #312, opened 2018-10-04
Fix macro for destroy_cuSPARSE_Handle

ndellingwood:


ndellingwood:
Minor fix for cusparse spgemm

Issue #307, opened 2018-10-01
KokkosBlas::dot is broken for complex, due to incorrect assumptions about Fortran ABI


mhoemmen:
@kyungjoo-kim @crtrott See comment here: https://github.com/trilinos/Trilinos/issues/3493#issuecomment-426044331

mhoemmen:
See this article on the relevant functions: http://mirror.informatimago.com/next/developer.apple.com/hardware/ve/errata.html

mhoemmen:
https://www.math.utah.edu/software/c-with-fortran.html#function-return-types

"...you should not expect to use Fortran functions that return types such as COMPLEX or COMPLEX*16. Write a SUBROUTINE interface to your Fortran function instead, and then invoke it as a void function from C or C++."

mhoemmen:
My Trilinos PR https://github.com/trilinos/Trilinos/pull/3538 only minimally patches kokkos-kernels to fix this issue.  I leave it to the kokkos-kernels developers to figure out the best longer-term approach.  However, please do fix this before the next develop-to-master promotion.

srajama1:
@kyungjoo-kim Can you take care of this please ?

kyungjoo-kim:
@crtrott In your dot testing, why do you use the following views ? 

```
    typedef Kokkos::View<ScalarA*[2],                                                                                                                        
       typename std::conditional<                                                                                                                            
                std::is_same<typename ViewTypeA::array_layout,Kokkos::LayoutStride>::value,                                                                  
                Kokkos::LayoutRight, Kokkos::LayoutLeft>::type,Device> BaseTypeA;                                                                            
    BaseTypeA b_a("A",N);                                                                                                                                    
    ViewTypeA a = Kokkos::subview(b_a,Kokkos::ALL(),0);                                                                                                      
```
In this way, the view is strided and it never tests any tpls as tpls require unit stride according to your ETI design.

mhoemmen:
@kyungjoo-kim `_DOT` in the BLAS takes strides (`INCX` and `INCY`).

kyungjoo-kim:
@mhoemmen I know that the standard blas interface has the stride interface but kokkoskernels ETI system instanciates for layoutleft (using unification of layout for 1D view layout right and layout left goes into the instanciated code but the strided case uses kokkos native implementation).

kyungjoo-kim:
@crtrott I confused view semantics and thought subview will return stride view. It is all oaky.

[...]

PR #314, opened 2018-10-05
Issue 307 - fix complex tpl local declarations and some other typos in KK

kyungjoo-kim:
related issues
#307 ,  trilinos/Trilinos#3493

In trilinos KK, complex TPL dot is currently disabled. This fixes the problem and enables TPL dot.

### DIFF
--- a/src/impl/tpls/KokkosBlas1_dot_tpl_spec_decl.hpp
+++ b/src/impl/tpls/KokkosBlas1_dot_tpl_spec_decl.hpp
@@ -47,14 +47,14 @@
 // Generic Host side BLAS (could be MKL or whatever)
 #ifdef KOKKOSKERNELS_ENABLE_TPL_BLAS
 
-extern "C" double               ddot_( const int* N, const double* x, const int* x_inc,
-                                       const double* y, const int* y_inc);
-extern "C" float                sdot_( const int* N, const float* x, const int* x_inc,
-                                       const float* y, const int* y_inc);
-extern "C" std::complex<double> zdotu_( const int* N, const std::complex<double>* x, const int* x_inc,
-                                       const std::complex<double>* y, const int* y_inc);
-extern "C" std::complex<float>  cdotu_( const int* N, const std::complex<float>* x, const int* x_inc,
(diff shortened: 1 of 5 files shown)


src/impl/tpls/KokkosBlas1_dot_tpl_spec_decl.hpp
```diff
@@ -47,14 +47,14 @@
...
-                                       const float* y, const int* y_inc);
-extern "C" std::complex<double> zdotu_( const int* N, const std::complex<double>* x, const int* x_inc,
-                                       const std::complex<double>* y, const int* y_inc);
-extern "C" std::complex<float>  cdotu_( const int* N, const std::complex<float>* x, const int* x_inc,
-                                       const std::complex<float>* y, const int* y_inc);
+extern "C" double ddot_ ( const int* N, const double* x, const int* x_inc,
+                          const double* y, const int* y_inc);
+extern "C" float  sdot_ ( const int* N, const float* x, const int* x_inc,
+                          const float* y, const int* y_inc);
+extern "C" void   zdotu_( std::complex<double> *res, const int* N, const std::complex<double>* x, const int* x_inc,
```
mhoemmen: 1. Does this way of returning a complex value work for all possible Fortran ABIs?
2. This interface still assumes that Fortran mangling uses a trailing underscore; this is not true on all platforms.
3. My objection about exposing extern declarations in user-facing header files still stands, though I think that would need to be addressed in a separate PR.

mhoemmen: 1. I'm not convinced but I'm willing to keep an open mind.
2. Both TeuchosNumerics and my old TpetraCore GEMM and GEMV wrappers used Fortran mangling macros.  You could imitate either of them.
3. The BLAS implements 4 Scalar types; they are always the same.  Why not just wrap them in functions whose interface you declare in a particular namespace?  That's what I did with the TpetraCore GEMM and GEMV wrappers.

ndellingwood:
Merging in this PR for now to fix the complex failures and progress the next kokkos promotion, we'll add new ticket regarding the exposed declarations.

PR #321, opened 2018-10-09
test_all_sandia, generate_makefile updates

ndellingwood:
generate_makefile: add option for users to select layout
  --with-layouts

test_all_sandia: add options for setting the scalar, ordinal, offset
types and layout


ndellingwood:
Address issue #315

PR #327, opened 2018-10-15
Make team abs error checking consistent with other PR #325 changes

ndellingwood:
Reimplement use of dot(r,y,y) instead of the subview and manual
dot reductions.

### DIFF
--- a/unit_test/blas/Test_Blas1_team_abs.hpp
+++ b/unit_test/blas/Test_Blas1_team_abs.hpp
@@ -134,7 +134,10 @@ namespace Test {
         expected_result[j] += AT::abs(h_x(i,j)) * AT::abs(h_x(i,j));
     }
 
-    double eps = std::is_same<ScalarA,float>::value?2*1e-5:1e-7;
+//    double eps = std::is_same<ScalarA,float>::value?2*1e-5:1e-7;
+    typename AT::mag_type eps = AT::epsilon()*1000;
+    typename AT::mag_type zero = AT::abs( AT::zero() );
+    typename AT::mag_type one = AT::abs( AT::one() );
 
     Kokkos::View<ScalarB*,Kokkos::HostSpace> r("Dot::Result",K);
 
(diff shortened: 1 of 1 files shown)


ndellingwood:
Cross-reference _R #325

unit_test/blas/Test_Blas1_team_abs.hpp
```diff
@@ -134,7 +134,10 @@ namespace Test {
         expected_result[j] += AT::abs(h_x(i,j)) * AT::abs(h_x(i,j));
     }
 
-    double eps = std::is_same<ScalarA,float>::value?2*1e-5:1e-7;
+//    double eps = std::is_same<ScalarA,float>::value?2*1e-5:1e-7;
+    typename AT::mag_type eps = AT::epsilon()*1000;
```
mhoemmen: For later: Let's do a bit of numerical analysis here to pick the error bound.  1000*eps is a bit large for small matrices.

ndellingwood: @mhoemmen 

> Let's do a bit of numerical analysis here to pick the error bound.

Definitely, just a first pass at getting things running with `float` types. The current error checking in some of these tests blows up the error as problem size grows, so this all needs to be revisited.

crtrott: If you look at Test_Blas3_gemm.hpp lines 140 following I actually did this for GEMM. That test should be fine. 
```c++
      double diff_C_average = diff_C/(N*M);
      // Expected Result: Random Walk in the least significant bit (i.e. ~ sqrt(K)*eps
      // eps scales with the total sum and has a factor in it for the accuracy of the operations ->
      // eps = K * 75 * machine_eps * 7
      double diff_C_expected = 1.0*sqrt(K)*K*75*machine_eps*7;
```
Basically you should be able to model this whole thing as a random walk with step size machine epsilon (this is for the additions). Than you add a factor in for the accuracy of the operations involved in this, and then you need the factors in how you aggregate the  errors.

mhoemmen: This is probably the easiest way.  Another option is to do _a posteriori_ error analysis, using a hand-rolled sequential reimplementation of the algorithm, and a version that replaces sums with absolute values.

ndellingwood:
@mhoemmen @crtrott are you okay if we get this merged in? This was just to clean up some slop I missed from PR #325, we can plan for a separate PR to fix the error checking more properly later. Both ideas you suggested for better checking sound good.

ndellingwood:
kokkos-dev spot-check:

```
[ndellin@kokkos-dev TestAllSandia-spot]$ tail nohup.out 
PASSED TESTS
clang-4.0.1-Pthread_Serial-release build_time=424 run_time=2044
cuda-8.0.44-Cuda_OpenMP-release build_time=837 run_time=1201
gcc-5.3.0-Serial-release build_time=340 run_time=975
gcc-7.3.0-Serial-release build_time=316 run_time=799
intel-17.0.1-OpenMP-release build_time=605 run_time=333
FAILED TESTS
```

mhoemmen:
@ndellingwood I'm OK with this for now.  Getting kokkos-kernels to have good tests is a separate effort.

ndellingwood:
Thanks @mhoemmen ! Could I get a review approval from you, or @crtrott @kyungjoo-kim @srajama1 ?

PR #330, opened 2018-10-19
Fix #329

### LINKED ISSUE #329: Wrong signature for axpy blas functions
In the file `src/impl/tpls/KokkosBlas1_axpby_tpl_spec_decl.hpp` the function axpy is declared as
```
extern "C" void daxpy_( const int* N, const double* alpha,
                                      const double* x, const int* x_inc,
                                      const double* y, const int* y_inc);
```
but it should be:
```
extern "C" void daxpy_( const int* N, const double* alpha,
                                      const double* x, const int* x_inc,
                                      double* y, const int* y_inc);
```
The second last argument is an output argument, so it shall not be const. This applies also to the other axpy functions.

mhoemmen:
Make "y" argument in BLAS `{s,d,c,z}axpy_` `extern` declarations nonconst.
This fixes #329.


mhoemmen:
Thanks @ndellingwood ! :D

ndellingwood:
Spot-check in progress for completeness.

PR #335, opened 2018-10-24
update trilinos integration scripts

ndellingwood:
Add script to support Pascal GPU testing


scripts/trilinos-integration/white_run_jenkins_script_cuda-pascal
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash -el
+ulimit -c 0
+
+KOKKOSKERNELS_BRANCH=$1
+TRILINOS_UPDATE_BRANCH=$2
+TRILINOS_PRISTINE_BRANCH=$3
+KOKKOS_UPDATE_BRANCH=$4
+
```
srajama1: Does this needs a separate file ? I thought this would an i/p parameter like the four parameters here. These are few line changes, right ?

scripts/trilinos-integration/white_run_jenkins_script_cuda-pascal
```diff
@@ -0,0 +1,89 @@
...
+export JENKINS_ARCH="Power8,Pascal60"
+export JENKINS_ARCH_CXX_FLAG="-mcpu=power8 -arch=sm_60 --expt-extended-lambda --std=c++11"
+export JENKINS_ARCH_C_FLAG="-mcpu=power8"
+export BLAS_LIBRARIES="${BLAS_ROOT}/lib/libblas.a;gfortran;gomp"
+export LAPACK_LIBRARIES="${LAPACK_ROOT}/lib/liblapack.a;gfortran;gomp;m"
+
+export JENKINS_DO_TESTS=ON
+export JENKINS_DO_EXAMPLES=ON
+
+export QUEUE=rhel7G
```
kyungjoo-kim: it used to be 7F and now it is 7G. I like it.

scripts/trilinos-integration/white_run_jenkins_script_cuda-pascal
```diff
@@ -0,0 +1,89 @@
...
+  TRILINOS_UPDATE_BRANCH=develop
+fi
+
+if [ -z $TRILINOS_PRISTINE_BRANCH ]
+then
+  TRILINOS_PRISTINE_BRANCH=develop
+fi
+
+module load devpack/20180521/openmpi/2.1.2/gcc/7.2.0/cuda/9.2.88
+module swap openblas/0.2.20/gcc/7.2.0 netlib/3.8.0/gcc/7.2.0
```
kyungjoo-kim: any reason not to use openblas ?

scripts/trilinos-integration/white_run_jenkins_script_cuda-pascal
```diff
@@ -0,0 +1,89 @@
...
+export OMP_NUM_THREADS=8
+export JENKINS_DO_CUDA=ON
+export JENKINS_DO_OPENMP=OFF
+export JENKINS_DO_PTHREAD=OFF
+export JENKINS_DO_SERIAL=ON
+export JENKINS_DO_COMPLEX=OFF
+
+export JENKINS_ARCH="Power8,Pascal60"
+export JENKINS_ARCH_CXX_FLAG="-mcpu=power8 -arch=sm_60 --expt-extended-lambda --std=c++11"
+export JENKINS_ARCH_C_FLAG="-mcpu=power8"
```
kyungjoo-kim: is this manual setup for cflags necessary ? it should be set from cmake script ?

scripts/trilinos-integration/white_run_jenkins_script_cuda-pascal
ndellingwood: We could make it a single file with extra parameters, but this seemed the easier path to me.

scripts/trilinos-integration/white_run_jenkins_script_cuda-pascal
ndellingwood: There is a Trilinos issue where use of OpenBLAS resulted in lots of problems on White/Ride and the solution they decided upon was to use Netlib instead. I just followed the same course as the ATDM testing.

scripts/trilinos-integration/white_run_jenkins_script_cuda-pascal
ndellingwood: I'm not sure how/if the KOKKOS_ARCH info propogates to the cflags, this may be a remnant that can be removed but was in the scripts when Christian updated them after introducing KOKKOS_ARCH so I left it in.

scripts/trilinos-integration/white_run_jenkins_script_cuda-pascal
srajama1: Ok, go ahead

scripts/trilinos-integration/white_run_jenkins_script_cuda-pascal
kyungjoo-kim: I guessed so. However, as this is trilinos integration test script, those environment variable should be caught from trilinos cmake script. If we provide correct environment, then we cannot test trilinos configure is right or not. While integrating trilinos, this problem is not important but it would be good to be addressed in my opinion.

PR #342, opened 2018-11-06
Disable Pthreads in Gtest

william76:
This relates to the issue in Trilinos where serial builds were failing
because they tried to load the pthreads library as requested by gtest
in KokkosKernels.

This PR mirrors changes made in the Trilinos repository so that the next KokkosKernels push doesn't reintroduce the bug.

See: https://github.com/trilinos/Trilinos/issues/3773


ndellingwood:
Thanks @william76!
@crtrott: now that the promotion is finished can we begin merging into develop again?

Issue #347, opened 2018-11-21
Pervasive build failures on geminga (Trilinos' Version of Kokkos-Kernels)


csiefer2:
A bunch of the print_specialization functions are only defined if KOKKOSKERNELS_ENABLE_TPL_BLAS is enabled, yet are used if
KOKKOSKERNELS_ENABLE_TPL_CUBLAS is enabled.

This leads to pervasive build failures.  If both of these really need to be enabled to make the code work, this mismatch should be detected at configure time.

See:
https://testing.sandia.gov/cdash/viewBuildError.php?buildid=4197910

csiefer2:
As per @ndellingwood Kokkos-Kernels will not build if TPL_ENABLE_BLAS is disabled when CUBLAS is enabled.

ndellingwood:
See these lines of code:

[CMakeLists.txt](https://github.com/kokkos/kokkos-kernels/blob/94456cf9099e3dcdd9bfb91a86a09a4f77a246d7/CMakeLists.txt#L293-L307)

ndellingwood:
Suggested fix:

After this [line](https://github.com/kokkos/kokkos-kernels/blob/94456cf9099e3dcdd9bfb91a86a09a4f77a246d7/CMakeLists.txt#L300), enable BLAS if not enabled. That is 

```
IF(${Kokkos_ENABLE_Cuda})
IF (NOT KOKKOSKERNELS_ENABLE_TPL_BLAS)
  SET(KOKKOSKERNELS_ENABLE_TPL_BLAS ON)
  LIST(APPEND TPL_LIST "BLAS")
ENDIF()
...
```

crtrott:
Solved in #348.

PR #348, opened 2018-11-21
CMakeLists.txt Fix for case CUBLAS=ON but BLAS not explicitly ON

ndellingwood:
Address issue #347 reported by @csiefer2


ndellingwood:
Confirmed this works with a Trilinos build with Cuda; kokkos-kernels enabled, no tpls enabled so TPL_ENABLE_BLAS was never set.

PR #350, opened 2018-11-27
Modify team gemm interface

huttered40:
I moved TeamGemm's MemberType template parameter from the class-template level the function-template level. This template parameter can always be deduced and thus should not need to be explicitely specified by the user. This change also allows SerialGemm and TeamGemm to have the same interface. I suspect other class templates like TeamGemv can also be changed.

The use case for this is I want to be able to pass SerialGemm and TeamGemm as policy classes for a Gemm policy. In order for this to be possible, the interfaces must match. With these changes, each class template shares the same three class-level template parameters.

I did the following steps, with no errors:
```
make install-lib -j16
cd unit_test
make -j
```

### DIFF
--- a/src/batched/KokkosBatched_Gemm_Decl.hpp
+++ b/src/batched/KokkosBatched_Gemm_Decl.hpp
@@ -32,12 +32,12 @@ namespace KokkosBatched {
     /// Team Gemm
     ///
 
-    template<typename MemberType,
-             typename ArgTransA,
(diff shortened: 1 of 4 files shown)


ndellingwood:
@kyungjoo-kim can you review this PR?

crtrott:
Kyungjoo is right. The member type tells us the execution space so we need that partial specialization capability here. Sorry Edward :-(

Issue #353, opened 2018-12-05
make clean - errors


ndellingwood:
Calling make clean results in errors like the following (please ignore the cpp14 suffixes, that is unrelated and artifact of previous PR):

```
s994290:Gcc6.4-cpp14 ndellin$ make clean
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C unit_test clean
/Applications/Xcode.app/Contents/Developer/usr/bin/make -f /Users/ndellin/Research/kokkos_items/kokkos-kernels/unit_test/Makefile  KOKKOS_DEVICES=Serial GTEST_PATH=/Users/ndellin/Research/kokkos_items/kokkos-kernels/../kokkos/tpls/gtest KOKKOSKERNELS_OPTIONS=eti-only KOKKOS_PATH=/Users/ndellin/Research/kokkos_items/kokkos-kernels/testing/Gcc6.4-cpp14/kokkos/install KOKKOSKERNELS_PATH=/Users/ndellin/Research/kokkos_items/kokkos-kernels/testing/Gcc6.4-cpp14/install KOKKOSKERNELS_SRC_PATH=/Users/ndellin/Research/kokkos_items/kokkos-kernels clean
/Users/ndellin/Research/kokkos_items/kokkos-kernels/unit_test/Makefile:41: /Users/ndellin/Research/kokkos_items/kokkos-kernels/testing/Gcc6.4-cpp14/install/Makefile.kokkos-kernels: No such file or directory
make[2]: *** No rule to make target `/Users/ndellin/Research/kokkos_items/kokkos-kernels/testing/Gcc6.4-cpp14/install/Makefile.kokkos-kernels'.  Stop.
make[1]: *** [clean] Error 2
make: *** [clean] Error 2
```

PR #354, opened 2018-12-05
src/Makefile, generate_makefile: Fixes for make clean

ndellingwood:
Updated so that calling 'make clean' works without errors.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -93,3 +93,7 @@ build: kokkoskernels-build-lib
 
 install: kokkoskernels-install
  
+clean: kokkoskernels-clean kokkos-clean
+	rm -f *.o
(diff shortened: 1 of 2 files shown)


ndellingwood:
PR to potentially address issue #353

src/Makefile
```diff
@@ -93,3 +93,7 @@ build: kokkoskernels-build-lib
 
 install: kokkoskernels-install
  
+clean: kokkoskernels-clean kokkos-clean
```
kyungjoo-kim: Do we also need to clean kokkos ?

ndellingwood: @kyungjoo-kim maybe not, I had put that in there while I was making sense of the issue and did not retest without once I had a better idea what was going on. Would you like me to test without and update the PR?

ndellingwood: Sounds good, thanks @kyungjoo-kim !

PR #356, opened 2018-12-10
test_all_sandia: update for white

ndellingwood:
Change xl to 16.1.1
Add actual spot-check list


ndellingwood:
xl 16.1.1 has headers set to gcc/7.2.0 allowing for c++14 support, xl 16.1.0 does not. Jenkins scripts were also modified to look for `ibm/16.1` rather than `ibm/16.1.0`

scripts/test_all_sandia
```diff
@@ -243,20 +243,29 @@ elif [ "$MACHINE" = "white" ]; then
...
+               "cuda/9.2.88 $CUDA_MODULE_LIST "Cuda_OpenMP" ${KOKKOS_PATH}/bin/nvcc_wrapper $CUDA_WARNING_FLAGS"
+    )
+  else
+    # Format: (compiler module-list build-list exe-name warning-flag)
+    COMPILERS=("gcc/6.4.0 $BASE_MODULE_LIST $IBM_BUILD_LIST g++ $GCC_WARNING_FLAGS"
+               "gcc/7.2.0 $BASE_MODULE_LIST $IBM_BUILD_LIST g++ $GCC_WARNING_FLAGS"
+               "ibm/16.1.0 $IBM_MODULE_LIST $IBM_BUILD_LIST xlC $IBM_WARNING_FLAGS"
+               "cuda/9.2.88 $CUDA_MODULE_LIST $CUDA_IBM_BUILD_LIST ${KOKKOS_PATH}/bin/nvcc_wrapper $CUDA_WARNING_FLAGS"
+    )
+  fi
```
kyungjoo-kim: Why do we have a if condition for spot_check ? The compiler description seems to be same. Is this for the convenience when we need to temporarily modify compilers  ?

ndellingwood: @kyungjoo-kim The build list is different - instead of building for every option with each compiler, the build list is a reduced subset which will speed up test times. I left the full list for gcc/7.2.0 because it is probably important to get full coverage there.
The xl build will still take some time, I had considered removing it but thought we should probably allow users to optionally do that instead.

kyungjoo-kim: @ndellingwood I got it. Thanks. Good to go.

ndellingwood:
@kyungjoo-kim thanks! Testing the script on white now to be safe.

nmhamster:
Just a reminder - this is a non-standard install although the mechanism to point at different header files is a supported features of XL (system admin only).

ndellingwood:
@nmhamster thanks Si - should we keep XL 16.1.0 in the compiler list as well for testing with/without the non-standard install?

nmhamster:
@ndellingwood - 16.1.0 is still the current "production" release for Sierra platform. It is fully C++11 compliant and so we should keep testing that until we get a new official production compiler change. 16.1.1 is in release candidate status, so nearly ready.

ndellingwood:
@nmhamster thanks! I'll update the PR so both versions are available for testing in the script.

ndellingwood:
Updated this PR so xl versions 16.1.0 and 16.1.1 are available. Similar PR update for kokkos issued as well kokkos/kokkos#1934

PR #373, opened 2019-01-23
tpl dot is wrong....

kyungjoo-kim:
It looks like that complex code is never well tested. Tpl complex dot uses zdotu which performs result += a(i)*b(i). However, it should use zdotc which does result += conj(a(i))*b(i). zdotc is consistent to what is implemented in Kokkos InnerProduct Traits.

### DIFF
--- a/src/impl/tpls/KokkosBlas1_dot_tpl_spec_decl.hpp
+++ b/src/impl/tpls/KokkosBlas1_dot_tpl_spec_decl.hpp
@@ -51,9 +51,9 @@ extern "C" double ddot_ ( const int* N, const double* x, const int* x_inc,
                           const double* y, const int* y_inc);
 extern "C" float  sdot_ ( const int* N, const float* x, const int* x_inc,
                           const float* y, const int* y_inc);
-extern "C" void   zdotu_( std::complex<double> *res, const int* N, const std::complex<double>* x, const int* x_inc,
+extern "C" void   zdotc_( std::complex<double> *res, const int* N, const std::complex<double>* x, const int* x_inc,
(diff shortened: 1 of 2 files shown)


kyungjoo-kim:
I need one more commit to change the test code. We did two mistakes and they cancel out each other and the test so far passed.

kyungjoo-kim:
@lucbv Your failure is probably due to the bug in the complex dot product. With this, it will fix your problem at least.

kyungjoo-kim:
@ndellingwood This can be merged now. I tested with openmp 
```
[----------] 272 tests from openmp (46003 ms total)

[----------] Global test environment tear-down
[==========] 272 tests from 1 test case ran. (46003 ms total)
[  PASSED  ] 272 tests.
[kyukim @blake-lsm1] unit_test > 
```
After this is merged, I will solve another issue related to fortran mangling by creating a c wrapper interface. It will be more clean at the end.

srajama1:
"We did two mistakes and they cancel out each other" - Oh man, can you please elaborate ?

ndellingwood:
> After this is merged, I will solve another issue related to fortran mangling by creating a c wrapper interface. It will be more clean at the end.

Thanks @kyungjoo-kim ! Is the interface work orthogonal to the ABI issue with zdot or will that be included also?

ndellingwood:
@srajama1 the test checked the result for `zdotu` rather than `zdotc` (u = unconjugated, c = conjugated), so the test passed as `zdotu` was being used which is what masked the issue, but `dot` should have been calling `zdotc` and the test checking that result.

srajama1:
@ndellingwood Got it, Thanks !

PR #379, opened 2019-01-29
Use homebrew's coreutils cp if available on osx

william76:
If the `coreutils` package is installed on OSX via Homebrew, there will be a gnu version of cp installed which does respect the `-u` option.
This will reduce the time it takes to recompile KokkosKernels on OSX systems since we won't be updating the timestamp on all the sources every time we recompile.

### DIFF
--- a/Makefile.kokkos-kernels
+++ b/Makefile.kokkos-kernels
@@ -456,6 +456,8 @@ endif
 
 DEPFLAGS = -M
 
+CP = cp
+
 ifeq ($(KOKKOS_OS),CYGWIN)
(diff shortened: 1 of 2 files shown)


william76:
@ndellingwood when you get a chance, can you have a look at this one?

ndellingwood:
@william76 I'll get this in soon, I started a round of integration tests a couple days ago to update Trilinos with fixes for BLAS TPL support with complex variable types (needed for the upcoming addition to the PR testing). Once complete I'll get this merged in, I'm holding off on non-essential merges to keep the testing results clean.

Issue #368, opened 2019-01-11
GEMM uses wrong function for computing shared memory allocation size


etphipp:
Looking at line 450 of KokkosBlas3_gemm_impl.hpp, GEMM uses `View::required_allocation_size()` to compute the shared memory allocation size:
```
int scratch_memory_size =
      ViewTypeAScratch::required_allocation_size() +
      ViewTypeBScratch::required_allocation_size() +
      ViewTypeCScratch::required_allocation_size();
```
I believe this is the wrong function, as it should instead use `View::shmem_size()`.  The difference between the two is `sizeof(value_type)`:
```
  static constexpr size_t required_allocation_size(
                                       const size_t arg_N0 = 0
                                     , const size_t arg_N1 = 0
                                     , const size_t arg_N2 = 0
                                     , const size_t arg_N3 = 0
                                     , const size_t arg_N4 = 0
                                     , const size_t arg_N5 = 0
                                     , const size_t arg_N6 = 0
                                     , const size_t arg_N7 = 0
                                     )
    {
      return map_type::memory_span(
        typename traits::array_layout
          ( arg_N0 , arg_N1 , arg_N2 , arg_N3
          , arg_N4 , arg_N5 , arg_N6 , arg_N7 ) );
    }
```
```
  static inline
  size_t shmem_size( typename traits::array_layout const& arg_layout )
  {
    return map_type::memory_span( arg_layout )+sizeof(typename traits::value_type);
  }
```
presumably so that the allocation for a view out of scratch memory is aligned to the size of the value_type.  As such, the requested size of shared memory may be too small, resulting in an error.  This affects large scalar types such as those in Stokhos.  A corresponding fix was pushed into Trilinos in PR [#4168](https://github.com/trilinos/Trilinos/pull/4168).

ndellingwood:
@etphipp I'll put in a PR matching the 3 changes in trilinos/Trilinos#4168 but can replace it if you have a PR ready to submit.

ndellingwood:
PR #381 issued matching fix by @etphipp and @kliegeois in Trilinos

PR #381, opened 2019-01-30
Fix GEMM scratch memory size

ndellingwood:
Addresses issue #368 reported by @etphipp, matches patch in
trilinos/Trilinos#4168
scratch_memory_size should call shmem_size rather than
required_allocation_size.


srajama1:
Already tested in Trilinos and ATDM. Looks good.

PR #389, opened 2019-02-17
SPMV_MV: reimplementing SpMV for multivectors using hierarchical parallelism

lucbv:
This work includes a new performance test to implementation
All unit-tests and perfromance test are passing.
Updated kokkos-kernels benchmark for comparison between SpMV and SpMV_MV.

### DIFF
--- a/perf_test/sparse/KokkosSparse_spmv_mv.cpp
+++ b/perf_test/sparse/KokkosSparse_spmv_mv.cpp
@@ -99,11 +99,108 @@ int SparseMatrix_generate(const OrdinalType nrows, const OrdinalType ncols,
   return nnz;
 }
 
+template<class AMatrix, class XVector, class YVector, bool conjugate>
+struct SPMV_MV_test_Functor {
+  using execution_space = typename AMatrix::execution_space;
+  using ordinal_type    = typename AMatrix::non_const_ordinal_type;
+  using value_type      = typename AMatrix::non_const_value_type;
+  using team_policy     = typename Kokkos::TeamPolicy<execution_space>;
+  using team_member     = typename team_policy::member_type;
+  using ATV             = Kokkos::Details::ArithTraits<value_type>;
+
+  const value_type alpha;
+  const value_type beta;
+  AMatrix m_A;
+  XVector m_x;
+  YVector m_y;
+
+  const ordinal_type numVecs;
+  const ordinal_type vecRemainder;
+  const ordinal_type rows_per_thread;
+  ordinal_type rows_per_team;
+  int x_strides[2], y_strides[2];
+
+  SPMV_MV_test_Functor (const value_type alpha_,
+                        const AMatrix m_A_,
+                        const XVector m_x_,
+                        const value_type beta_,
+                        const YVector m_y_,
+                        const ordinal_type rows_per_thread_) :
+    alpha(alpha_), m_A(m_A_), m_x(m_x_), beta(beta_), m_y(m_y_),
+    numVecs (m_x_.extent(1)), vecRemainder (m_x_.extent(1) % 16),
+    rows_per_thread(rows_per_thread_) {
+    static_assert (static_cast<int> (XVector::rank) == 2,
+                   "XVector must be a rank 2 View.");
+    static_assert (static_cast<int> (YVector::rank) == 2,
+                   "YVector must be a rank 2 View.");
+
+    m_x.stride(x_strides);
+    m_y.stride(y_strides);
+  } // SPMV_MV_test_Functor ()
+
+  KOKKOS_INLINE_FUNCTION
+  void operator() (const team_member& dev) const
+  {
+    for(ordinal_type loop = 0; loop < rows_per_thread; ++loop) {
+      const ordinal_type iRow = static_cast<ordinal_type>(dev.league_rank()*dev.team_size() + dev.team_rank())*rows_per_thread + loop;
+      if (iRow >= m_A.numRows ()) {
+	return;
+      }
+
+      value_type sum[16] = {};
+
+      ordinal_type remainder = vecRemainder;
+      const KokkosSparse::SparseRowViewConst<AMatrix> row = m_A.rowConst(iRow);
+      for(ordinal_type entryIdx = 0; entryIdx < row.length; ++entryIdx) {
+        const value_type val = alpha*(conjugate ? ATV::conj(row.value(entryIdx)) : row.value(entryIdx));
+        const ordinal_type colIdx = row.colidx(entryIdx);
+        // value_type* KOKKOS_RESTRICT x_ptr = &m_x(colIdx, 0);
+        for(ordinal_type vecIdx = 0; vecIdx < 16; ++vecIdx) {
+          // sum[vecIdx] += *(x_ptr + x_strides[1]*vecIdx)*val;
+	  sum[vecIdx] += m_x(colIdx, vecIdx)*val;
+        }
+      }
+
+      // value_type* KOKKOS_RESTRICT y_ptr = &m_y(iRow, 0);
+      for(ordinal_type vecIdx = 0; vecIdx < 16; ++vecIdx) {
+        // *(y_ptr + y_strides[1]*vecIdx) *= beta;
+        // *(y_ptr + y_strides[1]*vecIdx) += sum[vecIdx];
+	m_y(iRow, vecIdx) = m_y(iRow, vecIdx)*beta;
+	m_y(iRow, vecIdx) += sum[vecIdx];
+      }
+
+    }
+  } // operator()
+
+};
+
+template<class AMatrix, class XVector, class YVector>
+void spmv_mv_test(double alpha, AMatrix A, XVector x1, double beta, YVector y1) {
+  using size_type    = typename AMatrix::size_type;
+  using ordinal_type = typename AMatrix::non_const_ordinal_type;
+
+  const ordinal_type numRows = A.numRows();
+  const ordinal_type NNZPerRow = static_cast<ordinal_type> (A.nnz () / numRows);
+
+  int vector_length = 1;
+  while( (static_cast<ordinal_type> (vector_length*2*3) <= NNZPerRow) && (vector_length < 8) ) vector_length*=2;
+  const int rows_per_thread = KokkosSparse::RowsPerThread<typename AMatrix::execution_space>(NNZPerRow);
+
+  SPMV_MV_test_Functor<AMatrix, XVector, YVector, false> op (alpha, A, x1, beta, y1, rows_per_thread);
+
+  const int team_size = Kokkos::TeamPolicy<typename AMatrix::execution_space>::team_size_recommended(op, vector_length);
+  const int rows_per_team = rows_per_thread*team_size;
+  const size_type worksets = (numRows + rows_per_team - 1) / rows_per_team;
+
+  Kokkos::parallel_for("KokkosSparse::spmv_mv_test<MV,NoTranspose>", Kokkos::TeamPolicy<typename AMatrix::execution_space>
+                       ( worksets , team_size , vector_length ) , op );
+}
+
 template<typename Scalar, typename Layout>
 int test_spmv_mv(int numRows, int numCols, int nnz, const int numVecs,
                  const double alpha, const double beta, const char* filename,
                  const bool binaryfile, int rows_per_thread, int team_size, int vector_length,
-                 int idx_offset, int schedule, int loop, const int verbose) {
+                 int idx_offset, int schedule, int loop, const bool verbose) {
   typedef KokkosSparse::CrsMatrix<Scalar,int,Kokkos::DefaultExecutionSpace,void,int> matrix_type;
   typedef typename matrix_type::non_const_value_type   value_type;
   typedef typename matrix_type::device_type            device_type;
@@ -271,6 +368,29 @@ int test_spmv_mv(int numRows, int numCols, int nnz, const int numVecs,
           2.0*nnz*loop/ave_new_time/1e9, 2.0*nnz/max_new_time/1e9, 2.0*nnz/min_new_time/1e9,
           ave_new_time/loop*1000, max_new_time*1000, min_new_time*1000,
           num_errors);
+
+
+  Kokkos::fence();
+  if(verbose) { printf("Test performance of new spmv_mv implementation\n"); }
+  // Benchmark new impl
+  min_new_time = 1.0e32;
+  max_new_time = 0.0;
+  ave_new_time = 0.0;
+  for(int i = 0; i < loop; i++) {
+    Kokkos::Timer timer;
+    spmv_mv_test(alpha, A, x1, beta, y1);
+    Kokkos::fence();
+    double time = timer.seconds();
+    ave_new_time += time;
+    if(time > max_new_time) max_new_time = time;
+    if(time < min_new_time) min_new_time = time;
+  }
+  printf("%i %i %i %6.2lf ( %6.2lf %6.2lf %6.2lf ) ( %6.3lf %6.3lf %6.3lf ) ( %6.3lf %6.3lf %6.3lf ) %i RESULT\n",nnz, numRows,numCols,problem_size,
+          (matrix_size+vector_readwrite)/ave_new_time*loop/1024, (matrix_size+vector_readwrite)/max_new_time/1024,(matrix_size+vector_readwrite)/min_new_time/1024,
+          2.0*nnz*loop/ave_new_time/1e9, 2.0*nnz/max_new_time/1e9, 2.0*nnz/min_new_time/1e9,
+          ave_new_time/loop*1000, max_new_time*1000, min_new_time*1000,
+          num_errors);
+
   return (int)total_error;
 }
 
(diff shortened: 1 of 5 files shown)


mhoemmen:
1. Just curious -- any particular reason why you use 4 as the max (short) vector length?
2. It would be fun to do LayoutLeft vs. LayoutRight performance comparisons.  Christian's kernel unrolls across columns, specifically to make LayoutLeft less slow.  Belos' design makes it necessary to use LayoutLeft (since Belos conflates MultiVector with a basis of MultiVectors).

lucbv:
@mhoemmen, these are fair questions and this is what I have in mind so far:
1) at this point I still need to do some more tests to find out what length is working well for our problems. One of the outcomes of these tests will be an new benchmark page in the wiki of kokkos-kernels for spmv/spmv_mv.
2) yes, I am curious about this too and I actually talked with Karen and had an email from Heidi looking at what constraints we have around this LayoutLeft/LayoutRight choice for multivectors. If I come up with a definitive answer to that question I'll make sure to let you know.

mhoemmen:
Thanks @lucbv ! :-)

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -502,380 +502,449 @@ struct SPMV_MV_Transpose_Functor {
...
-  template<int UNROLL>
-  KOKKOS_INLINE_FUNCTION void
-  strip_mine (const team_member& dev, const ordinal_type& iRow, const ordinal_type& kk) const
-  {
-    y_value_type sum[UNROLL];
+  const ordinal_type rows_per_team;
+  const ordinal_type numVecs;
+  const ordinal_type vecRemainder;
+
+  struct value16 {
```
mhoemmen: Does kokkos-kernels already have a SIMD type?

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -502,380 +502,449 @@ struct SPMV_MV_Transpose_Functor {
...
+                         0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
+	  Kokkos::parallel_reduce(Kokkos::ThreadVectorRange(dev, row_length),
+				  [&] (const ordinal_type& entryIdx, value16& lsum) {
+
+				    // Extract data from current row
+				    const value_type val = alpha*(conjugate ? ATV::conj(row.value(entryIdx)) : row.value(entryIdx));
+				    const ordinal_type colIdx = row.colidx(entryIdx);
+
+				    // Perform dot product accros lhs columns
+				    lsum.x0  += m_x(colIdx, vecOffset +  0)*val;
```
mhoemmen: This feels like it should be a method on `value16` etc., that takes `lsum`, `m_x`, `colIdx`, and `val`.

perf_test/sparse/KokkosSparse_spmv_mv.cpp
```diff
@@ -0,0 +1,452 @@
...
+   double* val = NULL;
+   int* row = NULL;
+   int* col = NULL;
+   int numRows,numCols,nnz;
+   SparseMatrix_WriteBinaryFormat<double,int>(filename, numRows, numCols, nnz,
+                                              val, row, col, true, idx_offset);
+   return 0;
+ }
+
+ Kokkos::initialize(argc,argv);
```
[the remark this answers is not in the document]
lucbv: @mhoemmen I will wait for someone to write an example and show how it works.
Right now a naive use of `Kokkos::ScopeGuard()` leads to:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Constructing View and initializing data with uninitialized execution space
```

perf_test/sparse/KokkosSparse_spmv_mv.cpp
```diff
@@ -0,0 +1,452 @@
...
+typedef long long int LocalOrdinalType;
+#else
+typedef int LocalOrdinalType;
+#endif
+
+
+template< typename ScalarType , typename OrdinalType>
+int SparseMatrix_generate(OrdinalType nrows, OrdinalType ncols, OrdinalType &nnz, OrdinalType varianz_nel_row, OrdinalType width_row, ScalarType* &values, OrdinalType* &rowPtr, OrdinalType* &colInd)
+{
+  rowPtr = new OrdinalType[nrows+1];
```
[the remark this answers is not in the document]
lucbv: @mhoemmen this would break compatibility with the current matrix reader interface and I do not wish to rewrite them now.
I will keep this in mind for when time can be spared to tackle a refactor of `SparseMatrix_MatrixMarket_read` and `SparseMatrix_ReadBinaryFormat` in order to have them support Kokkos views.

perf_test/sparse/KokkosSparse_spmv_mv.cpp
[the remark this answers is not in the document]
mhoemmen: Example:
```
int main (int argc, char* argv[]) {
  Kokkos::ScopeGuard kokkosScope (argc, argv); // you must name it
  { 
    // ... do Kokkos stuff; don't let Views persist beyond call of kokkosScope's destructor
  }
  return 0;
}
```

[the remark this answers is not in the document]
lucbv: OK, I actually was able to use the scope guard following the example : )

src/sparse/impl/KokkosSparse_spmv_impl.hpp
lucbv: True, I am waiting for more feedback on the simd type before going ahead and using them in this kernel

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -502,380 +502,449 @@ struct SPMV_MV_Transpose_Functor {
...
+      x11 += input.x11;
+      x12 += input.x12;
+      x13 += input.x13;
+      x14 += input.x14;
+      x15 += input.x15;
     }
 
-    const auto row = m_A.rowConst (iRow);
+    KOKKOS_INLINE_FUNCTION
+    value_type get(const ordinal_type vecIdx) {
```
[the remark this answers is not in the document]
lucbv: I am a bit wary of constexpr on Cuda...

srajama1:
@lucbv : Is this PR ready for a final review ?

lucbv:
Almost, I found a bug that seems to come from the Kokkos implementation of view stride when the `scale_type` of the view is `Kokkos::complex<>`.
I found a work around but it took time as the failure only happens on bowman with intel-17.2.174 with O2 and O3 optimization...

I can run the bowman spot-check fine now. I need to fix a couple more things for the spot-check on white.

srajama1:
@lucbv Ok I will wait.

lucbv:
@srajama1, I had to do some minor modification to have the spot-check pass without failures as you can see below. I would like to rerun some performance tests to check whether I need to add pragma due to my changes. I will keep you posted, this should be fairly quick.

Bowman:
PASSED TESTS
intel-16.4.258-Pthread-release build_time=1188 run_time=2186
intel-16.4.258-Pthread_Serial-release build_time=1799 run_time=4533
intel-16.4.258-Serial-release build_time=1144 run_time=2239
intel-17.2.174-OpenMP-release build_time=1612 run_time=817
intel-17.2.174-OpenMP_Serial-release build_time=2107 run_time=3226
intel-17.2.174-Pthread-release build_time=1092 run_time=2151
intel-17.2.174-Pthread_Serial-release build_time=2371 run_time=4417
intel-17.2.174-Serial-release build_time=1098 run_time=2284
intel-18.2.199-OpenMP-release build_time=1353 run_time=822
FAILED TESTS

White:
PASSED TESTS
cuda-10.0.130-Cuda_Serial-release build_time=730 run_time=1012
cuda-9.2.88-Cuda_OpenMP-release build_time=786 run_time=617
gcc-6.4.0-OpenMP_Serial-release build_time=366 run_time=908
gcc-7.2.0-OpenMP-release build_time=280 run_time=239
gcc-7.2.0-OpenMP_Serial-release build_time=449 run_time=925
gcc-7.2.0-Serial-release build_time=179 run_time=632
ibm-16.1.0-Serial-release build_time=770 run_time=858
FAILED TESTS

lucbv:
@srajama1 actually the spot-check on bowman is not yet finished, I'll post final results tomorrow morning together with performance results.

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -985,10 +986,10 @@ struct SPMV_MV_OpenMP_Functor {
 			  const XVector m_x_,
 			  const value_type beta_,
 			  const YVector m_y_,
-			  const ordinal_type rows_per_team_) :
+			  const int rows_per_thread_) :
```
mhoemmen: The right type should be `ordinal_type`, if that's what you're using to store the number of rows.  What if there's only one thread?  Then this number needs to store the number of rows.

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -1001,46 +1002,79 @@ struct SPMV_MV_OpenMP_Functor {
...
       for(ordinal_type vecOffset = 0; vecOffset < numVecs - 15; vecOffset += 16) {
-	strip_mine_large<16>(row, iRow, row_length, vecOffset, m_x_stride, m_y_stride);
+	strip_mine<16>(iRow, vecOffset);
       } // Loop over numVecs
 
-      if(remainder > 7) {
-	strip_mine_large<8>(row, iRow, row_length, numVecs - remainder, m_x_stride, m_y_stride);
-	remainder -= 8;
-      }
+      switch(remainder) {
```
mhoemmen: Do we really need to instantiate all 16 of these?  Is the larger build time worth the performance?

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -502,63 +508,606 @@ struct SPMV_MV_Transpose_Functor {
...
+
+  KOKKOS_INLINE_FUNCTION
+  void operator() (const team_member& dev) const
+  {
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(dev,0,rows_per_team), [&] (const ordinal_type& loop) {
+        const ordinal_type iRow = static_cast<ordinal_type>(dev.league_rank()) * rows_per_team + loop;
+        if (iRow >= m_A.numRows ()) {
+          return;
+        }
+        const KokkosSparse::SparseRowViewConst<AMatrix> row = m_A.rowConst(iRow);
```
mhoemmen: Consider `auto` here.

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -502,63 +508,606 @@ struct SPMV_MV_Transpose_Functor {
...
+        const ordinal_type row_length = static_cast<ordinal_type> (row.length);
+        ordinal_type remainder = vecRemainder;
+
+	for(ordinal_type vecOffset = 0; vecOffset < numVecs - 15; vecOffset += 16) {
+	  value16 sum = {};
+	  Kokkos::parallel_reduce(Kokkos::ThreadVectorRange(dev, row_length),
+				  [&] (const ordinal_type& entryIdx, value16& lsum) {
```
mhoemmen: `ordinal_type` is probably `int`, so it's better to pass by value than by reference.

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -502,63 +508,606 @@ struct SPMV_MV_Transpose_Functor {
...
+
+				    // Extract data from current row
+				    const value_type val = alpha*(conjugate ? ATV::conj(row.value(entryIdx)) : row.value(entryIdx));
+				    const ordinal_type colIdx = row.colidx(entryIdx);
+
+				    // Perform dot product accros lhs columns
```
mhoemmen: "across"

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -502,63 +508,606 @@ struct SPMV_MV_Transpose_Functor {
...
+				    lsum.x9  += m_x(colIdx, vecOffset +  9)*val;
+				    lsum.x10 += m_x(colIdx, vecOffset + 10)*val;
+				    lsum.x11 += m_x(colIdx, vecOffset + 11)*val;
+				    lsum.x12 += m_x(colIdx, vecOffset + 12)*val;
+				    lsum.x13 += m_x(colIdx, vecOffset + 13)*val;
+				    lsum.x14 += m_x(colIdx, vecOffset + 14)*val;
+				    lsum.x15 += m_x(colIdx, vecOffset + 15)*val;
+				  }, sum);
+
+          Kokkos::parallel_for(Kokkos::ThreadVectorRange(dev, 16), [&] (const ordinal_type& vecIdx) {
```
mhoemmen: `ordinal_type` is probably `int`, so it's better to pass by value than by reference.

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -502,63 +508,606 @@ struct SPMV_MV_Transpose_Functor {
...
+			  const YVector m_y_,
+			  const int rows_per_thread_) :
+    alpha (alpha_), m_A (m_A_), m_x (m_x_),
+    beta (beta_), m_y (m_y_),
+    rows_per_thread (static_cast<ordinal_type>(rows_per_thread_)),
+    numVecs (m_x_.extent(1)),
+    vecRemainder (m_x_.extent(1) % 16)
   {
-    y_value_type sum[UNROLL];
+    static_assert (static_cast<int> (XVector::rank) == 2,
```
mhoemmen: `XVector::Rank == 2` (note capitalization) works fine; no cast needed.

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -567,315 +1116,83 @@ struct SPMV_MV_LayoutLeft_Functor {
...
-        for (ordinal_type iEntry = 0;
-             iEntry < row.length;
-             iEntry ++)
-#endif
-      {
-      const A_value_type val = conjugate ?
-        Kokkos::Details::ArithTraits<A_value_type>::conj (row.value(iEntry)) :
-        row.value(iEntry);
-      const ordinal_type ind = row.colidx(iEntry);
+    for(int entryIdx = 0; entryIdx < row.length; ++entryIdx) {
```
mhoemmen: The right loop index type here should be `ordinal_type`, since it is large enough to hold all the entries in a row (assuming no duplicates).

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -567,315 +1116,83 @@ struct SPMV_MV_LayoutLeft_Functor {
 #pragma unroll
 #endif
 #ifdef KOKKOS_ENABLE_PRAGMA_LOOPCOUNT
-#pragma loop count (15)
+#pragma loop count(15)
```
mhoemmen: Do we still need this?  The point of this was to turn off vectorization on KNC, no?

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -908,59 +1225,62 @@ spmv_alpha_beta_mv_no_transpose (const typename YVector::non_const_value_type& a
...
     // cannot be more than the number of columns of the matrix.  Thus,
     // the appropriate type is ordinal_type.
-    const ordinal_type NNZPerRow = static_cast<ordinal_type> (A.nnz () / A.numRows ());
-
-    int vector_length = 1;
-    while( (static_cast<ordinal_type> (vector_length*2*3) <= NNZPerRow) && (vector_length<8) ) vector_length*=2;
+    const ordinal_type numRows = A.numRows();
+    const ordinal_type NNZPerRow = static_cast<ordinal_type> (A.nnz () / numRows);
 
 #ifndef KOKKOS_FAST_COMPILE // This uses templated functions on doalpha and dobeta and will produce 16 kernels
```
mhoemmen: Do we ever test with `KOKKOS_FAST_COMPILE`?  It's a noble idea, but useless if not tested regularly.  We should just make the one set of kernels build in reasonable time.

src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -908,59 +1225,62 @@ spmv_alpha_beta_mv_no_transpose (const typename YVector::non_const_value_type& a
...
+      if(team_size < 0) {
+	Kokkos::parallel_for("KokkosSparse::spmv_mv<MV,NoTranspose>", Kokkos::TeamPolicy< typename AMatrix::execution_space >
+			     ( worksets , Kokkos::AUTO , vector_length ) , op );
+      } else {
+	Kokkos::parallel_for("KokkosSparse::spmv_mv<MV,NoTranspose>", Kokkos::TeamPolicy< typename AMatrix::execution_space >
+			     ( worksets , team_size , vector_length ) , op );
+      }
+    }
 
 #else // KOKKOS_FAST_COMPILE this will only instantiate one Kernel for alpha/beta
```
mhoemmen: See note above.  If we don't test this case, we should purge it.

lucbv:
@srajama1 the last spot check fix resulted in performance degradation so I had to do some more tweaking, now performance is back and the spot check should pass. I will post again when this is all done.

lucbv:
This work has now been superseded by work in PR #828 so I'm closing this

PR #393, opened 2019-03-06
Add doc/kokkos-promotion.txt file

ndellingwood:
Documentation for promotion testing process

### DIFF
--- a/doc/kokkos-promotion.txt
+++ b/doc/kokkos-promotion.txt
@@ -42,6 +42,11 @@ supported compilers. Those machines are:
          nohup ../scripts/test_all_sandia &
          tail -f nohup.out                   # to watch progress
 
+         NOTE: To kill jobs not running in the background (hopefully prevent ghost processes):
+
+             control+z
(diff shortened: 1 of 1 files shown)


ndellingwood:
Only updates kokkos-promotion.txt file, does not touch code so I'm not running spot-check.

doc/kokkos-promotion.txt
```diff
@@ -0,0 +1,240 @@
...
+  1.4. Create a testing directory
+
+         mkdir testing
+         cd testing
+
+  1.5. Run the test_all_sandia script with no options to test all compilers
+
+         nohup ../scripts/test_all_sandia &
+         tail -f nohup.out                   # to watch progress
+
```
kyungjoo-kim: Could you give the instruction to delete jobs when the test is failed ? You once told me but I do not remember it.

ndellingwood: Good point, 
`ctrl+z`
`kill -9 %1`

I'll update and add this.

PR #394, opened 2019-03-06
Update test_all_sandia

ndellingwood:
1. Recognize kokkos-dev machine
2. Update sems-recongnized machines:
   modules tested are common to both rhel6 and rhel7 envs
   remove use of kokkos-env modules

### DIFF
--- a/scripts/test_all_sandia
+++ b/scripts/test_all_sandia
@@ -208,7 +208,7 @@ if [ "$MACHINE" = "sems" ]; then
   source /projects/sems/modulefiles/utils/sems-modules-init.sh
 
   # On unnamed sems machines, assume more restricted rhel7 environment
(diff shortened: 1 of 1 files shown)


ndellingwood:
Still needs to be tested

ndellingwood:
@kyungjoo-kim in either Makefile.kokkos or nvcc_wrapper the default arch is set to sm_35, for Kepler which matches kokkos-dev. Note this is specific to kokkos-dev, not the new kokkos-dev-2 (which will use the sems env settings).

ndellingwood:
@kyungjoo-kim and if one is to test on kokkos-dev-2, then --arch="Volta70" would be needed. kokkos-dev is a special case that also uses modules from kokkos-env, on systems with a sems env I'm not sure what the most common or safest arch would be to default to.

kyungjoo-kim:
Okay. kokkos-dev is fine with the default setting. Thanks.

PR #399, opened 2019-03-18
Fix #397 (Add GEMM CudaUVMSpace specializations)

### LINKED ISSUE #397: GEMM has no CudaUVMSpace specializations for cuBLAS
Kokkos-kernels' GEMM struct, called by `KokkosBlas::gemm`, has no CudaUVMSpace specializations that call cuBLAS.  It has CudaSpace specializations:

https://github.com/kokkos/kokkos-kernels/blob/4ee5f3c6dbd0981f6d8c7a9b2b1763439cb56039/src/impl/tpls/KokkosBlas3_gemm_tpl_spec_avail.hpp#L121

but no CudaUVMSpace specializations.  @vbrunini and I found this out by benchmarking Belos solves with Tpetra objects.  We found that the `X^T * Y` operation (`Tpetra::MultiVector::multiply`) for tall and skinny X and Y was _very_ slow compared to the CPU, even though the GPU solves as a whole were faster than the CPU solves.  Kokkos profiling led us to labels that only the default implementation of GEMM would use.

The logical fix would be to add CudaUVMSpace specializations to that file.  We're trying that now.  This shouldn't add too much code, since it just dispatches to cuBLAS.  Another approach would be to change how `gemm_tpl_spec_avail` works, so that the macros fill in ExecSpace and MemSpace, instead of just MemSpace.  Then, you would set `Cuda, AnonymousSpace` and thus have only one specialization.

mhoemmen:
Add CudaUVMSpace specializations for GEMM.  This should fix performance issues that @vbrunini noticed on CUDA.  Thanks to @kyungjoo-kim and @vbrunini for feedback.

Corresponding Trilinos patch:
https://github.com/trilinos/Trilinos/pull/4649

### DIFF
--- a/src/impl/tpls/KokkosBlas3_gemm_tpl_spec_avail.hpp
+++ b/src/impl/tpls/KokkosBlas3_gemm_tpl_spec_avail.hpp
@@ -119,35 +119,43 @@ struct gemm_tpl_spec_avail< \
 #if defined (KOKKOSKERNELS_INST_DOUBLE) \
  && defined (KOKKOSKERNELS_INST_LAYOUTLEFT)
  KOKKOSBLAS3_GEMM_TPL_SPEC_AVAIL_CUBLAS( double,                  Kokkos::LayoutLeft, Kokkos::LayoutLeft, Kokkos::LayoutLeft, Kokkos::CudaSpace)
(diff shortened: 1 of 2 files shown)


kyungjoo-kim:
I did put a PR #398 for this, too. but I will just close mine or I will include others in that PR.

mhoemmen:
@kyungjoo-kim It's up to you as long as we merge one of them : - )

ndellingwood:
@kyungjoo-kim shall we merge this PR then instead of #398?

kyungjoo-kim:
Yes. These are essentially same.

Issue #411, opened 2019-04-08
Use of deprecated Kokkos code. 


crtrott:

ndellingwood:
See PR #412

PR #412, opened 2019-04-08
Fix two instances of using deprecated code (one of them inadvertenly)

crtrott:


ndellingwood:
Addresses issue #411

ndellingwood:
Changes are good and straight-forward so I'm merging this in, since deprecated code is off by default  kokkos-kernels tests will now continue failing without this update.

Issue #371, opened 2019-01-19
Remove experimental name space for batched blas


srajama1:
Let us remove the experimental namespace from the batched BLAS codes. It is used by applications and I think we are reasonably confident on how the i/f should look like.

srajama1:
We should do this for 3.0 ..

PR #414, opened 2019-04-08
Remove experimental namespace in batched functors.

kyungjoo-kim:
Remove experimental namespace from kokkosbatched functors. #371


kyungjoo-kim:
@ndellingwood If I were you, I would not try to precisely examine this PR as it was a simple sed script work even thought it change 138 files. I tested spotcheck on white and it passes with the version of today kokkos. Could you merge this PR ?

kyungjoo-kim:
it has a temporal promotion macro that distinguish kokkos::batched from kokkos::batched::experimental. so it is safe with trilinos.

srajama1:
@kyungjoo-kim : Thanks, merging !

PR #415, opened 2019-04-17
D2GC: Reduce unit test problem size

william76:
Reduces the size of the 'test' graph for D2GC by 10x and removes the message to stdout when the distance-2 coloring is validated as correct.


william76:
Patching this into [Trilinos too][1]

[1]: https://github.com/trilinos/Trilinos/pull/4940

ndellingwood:
This slipped through and the Trilinos changes were clobbered and required resubmit. Merging to prevent reclobbering.

PR #416, opened 2019-04-18
Add support for HPX backend

msimberg:
Adds support for the HPX backend that was merged to Kokkos develop (https://github.com/kokkos/kokkos/pull/1975).

I've had to do the same exercise of adding `fence`s to make sure the asynchronous version runs correctly. E.g. `dot` has one version which returns the result and fences, and another which uses an out parameter and doesn't fence. These look pretty well defined (in terms of what blocks and what not) but would be good if you could confirm that the fences are actually where they should be.

There are some which fail consistently even with the OpenMP backend, and I haven't attempted to fix them (I couldn't find issues for them. Will open some if needed.). Let me know if these are known issues and if I can ignore them.

These
```
mult_mv_double_int
team_mult_double_mv_int
update_double_int
update_mv_double_int
team_update_double_int
team_update_double_mv_int
batched_scalar_serial_trsv_u_nt_n_double_double
batched_scalar_serial_trsv_u_nt_n_dcomplex_dcomplex
batched_scalar_serial_trsv_u_nt_n_dcomplex_double
```
all fail with something in the style of:
```
/apps/daint/UES/simbergm/src/kokkos-kernels/unit_test/../test_common/KokkosKernels_TestUtils.hpp:87: Failure                                                                                                                                                                     
The difference between double(AT1::abs(val1)) and double(AT2::abs(val2)) is 0.25, which exceeds double(AT3::abs(tol)), where                                                                                                                                                     
double(AT1::abs(val1)) evaluates to 1644610.75,                                                                                                                                                                                                                                  
double(AT2::abs(val2)) evaluates to 1644610.5, and                                                                                                                                                                                                                               
double(AT3::abs(tol)) evaluates to 0.16446105.
```
i.e. close but not quite.

`reciprocal_int` fails with a division by zero.


msimberg:
Ping @crtrott. Just checking what the status is on your side for having this merged.

PR #423, opened 2019-05-14
Abbreviated type names

akhilPotla97:


crtrott:
What about just renaming the directory to eticpp or so? Wouldn't that safe a lot of characters?

PR #427, opened 2019-05-16
KokkosBatched - add specialization for float2, float4 and double4.

kyungjoo-kim:
These are for testing two scalar type approach in ifpack2 running spotcheck on white.


kyungjoo-kim:
@ndellingwood This is the PR that I mentioned just before. The conflicting files are a result from git mv. I changed the location.

kyungjoo-kim:
```
[kyukim@white26 spotcheck]$ . ../spotcheck.sh
Running on machine: white
Going to test compilers:  cuda/9.2.88
Testing compiler cuda/9.2.88
PASSED TESTS
cuda-9.2.88-Cuda_OpenMP-release build_time=885 run_time=384
FAILED TESTS
```

src/batched/KokkosBatched_Vector_SIMD.hpp
```diff
@@ -135,6 +117,87 @@ namespace KokkosBatched {
...
 
+  template<>
+  class Vector<SIMD<float>,2> {
+  public:
+    using type = Vector<SIMD<float>,2>;
+    using value_type = float;
+    using mag_type = float;
+
+    enum : int { vector_length = 2 };
+    typedef float2 data_type;
```
mhoemmen: Could names like `float2` collide with existing types in the global namespace in e.g., CUDA?

kyungjoo-kim: This is intended to override the cuda builtin type (float2, doule2, float4, double4).

PR #432, opened 2019-05-31
KokkosBatched - reduce vector length from 16 to 8.

kyungjoo-kim:
Most important part in memory efficiency is keeping 32 byte cacheline.
By reducing the length, we can increase occupancy for smaller problems
and sparc is the case.


kyungjoo-kim:
@ndellingwood This is a simple change. Would you please merge this and include in the kokkos promotion ?

ndellingwood:
@kyungjoo-kim merged!

PR #409, opened 2019-04-04
Add support for TPL MAGMA GESV

vqd8a:
Address issue #406 . Initital implementation. Work in progress.

PR #436, opened 2019-06-20
Add TPL BLAS support for gesv

vqd8a:
I have just updated the TPL BLAS support for GESV. This is a supplement to #409.


src/blas/KokkosBlas_gesv.hpp
```diff
@@ -93,6 +93,14 @@ gesv (const char pivot[],
       "Valid values include 'N' or 'n' (No pivoting), 'Y' or 'y' (partial pivoting).";
     Kokkos::Impl::throw_runtime_exception (os.str ());
   }
+#ifdef KOKKOSKERNELS_ENABLE_TPL_BLAS
+  if(!((pivot[0] == 'Y') || (pivot[0] == 'y'))) {
+    std::ostringstream os;
+    os << "KokkosBlas::gesv: pivot[0] = '" << pivot[0] << "'. " <<
+      "TPL BLAS does not support no pivoting.";
```
mhoemmen: Why not just dispatch to your hand-rolled implementation for this case?

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
```diff
@@ -44,6 +44,216 @@
...
+      Kokkos::MemoryTraits<Kokkos::Unmanaged> > BViewType; \
+      \
+  static void \
+  gesv (const char pivot[], \
+        const AViewType& A, \
+        const BViewType& B) { \
+    \
+    Kokkos::Profiling::pushRegion("KokkosBlas::gesv[TPL_BLAS,float]"); \
+    gesv_print_specialization<AViewType,BViewType>(); \
+    const bool pivot_t = (pivot[0]=='Y') || (pivot[0]=='y'); \
```
mhoemmen: `*_t` by convention indicates a type, not a `bool`.

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
```diff
@@ -44,6 +44,216 @@
...
+    const int LDA  = (AST == 0) ? 1 : AST; \
+    const int BST  = static_cast<int> (B.stride(1)); \
+    const int LDB  = (BST == 0) ? 1 : BST; \
+    const int NRHS = static_cast<int> (B.extent(1)); \
+    \
+    int info = 0; \
+    \
+    if(pivot_t) { \
+      int *ipiv = NULL; \
+      ipiv = (int*)malloc(N*sizeof(int)); \
```
mhoemmen: This is C++; it's preferred to use `new` / `delete` (or better yet, `unique_ptr<int[]>`).

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
```diff
@@ -44,6 +44,216 @@
...
+      HostBlas<std::complex<double> >::gesv \
```
mhoemmen: I'm not sure how you've done this, but normally all complex types would turn into `Kokkos::complex` by the time they go into a View.

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
vqd8a: Thanks @mhoemmen. They are all changed.

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
vqd8a: @mhoemmen Thanks. ```new```/```delete``` are now used.

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
vqd8a: @mhoemmen Thanks. This is consistent with how other BLAS functions are done.

src/blas/KokkosBlas_gesv.hpp
vqd8a: Thanks @mhoemmen. I totally agree with you. But, I think we do not have the implementation yet (I planned to do it in another PR). I think we only have the batched versions which are good for small size problems. @kyungjoo-kim and @srajama1, please correct me if I am wrong.

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
```diff
@@ -44,6 +44,216 @@
...
+  typedef Kokkos::View<SCALAR**, LAYOUT, Kokkos::Device<ExecSpace, MEM_SPACE>, \
+      Kokkos::MemoryTraits<Kokkos::Unmanaged> > AViewType; \
+  typedef Kokkos::View<SCALAR**, LAYOUT, Kokkos::Device<ExecSpace, MEM_SPACE>, \
+      Kokkos::MemoryTraits<Kokkos::Unmanaged> > BViewType; \
+ \
+  static void \
+  gesv (const char pivot[], \
+        const AViewType& A, \
+        const BViewType& B) { \
+    \
```
kyungjoo-kim: You should get all necessary arrays as arguments instead of allocating inside of the function. When this code is often used, such internal allocation will be harmful.

vqd8a: @kyungjoo-kim Thanks for suggestion. You also mean the user should be responsible for allocating the pivoting array, right? Then, we should have two interfaces one for gesv with pivoting and one for gesv without pivoting.

kyungjoo-kim: no. if the provided pivot view is a trivial object (zero extent), then you can consider it as no pivot version.

mhoemmen: ^^^ what Kyungjoo said -- the BLAS doesn't allocate, so users will be surprised if this function allocates.

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
```diff
@@ -100,11 +100,11 @@ struct GESV< \
     \
     int info = 0; \
     \
-    if(pivot_t) { \
-      int *ipiv = NULL; \
-      ipiv = (int*)malloc(N*sizeof(int)); \
+    if(with_pivot) { \
+      int *ipiv = nullptr; \
+      ipiv = new int[N]; \
```
mhoemmen: Why doesn't the function just take an array of pivots, that is unused (and could be null) if `with_pivot` is false?

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
```diff
@@ -142,11 +142,11 @@ struct GESV< \
     \
     int info = 0; \
     \
-    if(pivot_t) { \
-      int *ipiv = NULL; \
-      ipiv = (int*)malloc(N*sizeof(int)); \
+    if(with_pivot) { \
+      int *ipiv = nullptr; \
```
mhoemmen: See above.

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
vqd8a: What @kyungjoo-kim  said, we even don't need ```with_pivot```. Users just need to provide zero-extent pivot view for no pivot case.

vqd8a: or should we use ```with_pivot``` (false) and zero-extent pivot view?

kyungjoo-kim: It is just my preference. I do not like putting duplicated information (zero length array and with_pivot arguments can contain common information as ``with_pivot = ipiv.extent(0) != 0``). @mhoemmen comment is that it can cause unused variable warning but as we check its length the variable is always used for both cases.

mhoemmen: `with_pivot` is a run-time option, so as long as you use the input pivot array inside that case, there should be no variable warning.

You should be able to dispatch straight to LAPACK in the host case.  `DGESV` [takes a pivot array](http://www.netlib.org/lapack/explore-html/d7/d3b/group__double_g_esolve_ga5ee879032a8365897c3ba91e3dc8d512.html).

vqd8a:
The ```gesv``` interface has been changed. Users is responsible for allocating pivot view and provide as an argument. Zero-extent pivot view is considered as no pivoting case.
The ```gesv``` MAGMA TPL is also changed accordingly.

src/impl/tpls/KokkosBlas_Host_tpl.cpp
```diff
@@ -249,6 +249,23 @@ extern "C" {
...
+                                     float*, int*, int*,
+                                     float*, int*, int* );
+  void F77_BLAS_MANGLE(dgesv,DGESV)( int*, int*,
+                                     double*, int*, int*,
+                                     double*, int*, int* );
+  void F77_BLAS_MANGLE(cgesv,CGESV)( int*, int*,
+                                     std::complex<float>*, int*, int*,
+                                     std::complex<float>*, int*, int* );
+  void F77_BLAS_MANGLE(zgesv,ZGESV)( int*, int*,
+                                     std::complex<double>*, int*, int*,
```
mhoemmen: Sometimes compilers warn that `extern "C"` declarations can't accept `std::complex<T>`.  If that should happen, you could always have these functions take `void*` instead of `std::complex<T>*`.  That is what the BLAS' C binding does.

src/impl/tpls/KokkosBlas_Host_tpl.cpp
```diff
@@ -249,6 +249,23 @@ extern "C" {
                                      const std::complex<double>*,
                                      const std::complex<double>*, int*,
                                      /* */ std::complex<double>*, int* );
+
+  ///
+  /// Gesv
+  ///
+
+  void F77_BLAS_MANGLE(sgesv,SGESV)( int*, int*,
```
mhoemmen: These are not BLAS functions, they are LAPACK functions.  They may live in a different library.  It's plausible that people will want to build and run on platforms that don't have a Fortran compiler, and thus can't build LAPACK.  Thus, I think it would make sense to separate out LAPACK `extern "C"` declarations into a different `.cpp` file.

src/impl/tpls/KokkosBlas_Host_tpl.cpp
```diff
@@ -443,6 +465,15 @@ namespace KokkosBlas {
                      a, &lda,
                      b, &ldb);
     }
+    template<>
+    void 
+    HostBlas<float>::gesv(int n, int rhs,
```
mhoemmen: `xGESV` routines are in LAPACK, not the BLAS.  They should really go in a separate `HostLapack` struct.

src/impl/tpls/KokkosBlas_Host_tpl.cpp
```diff
@@ -812,6 +861,15 @@ namespace KokkosBlas {
                      (const std::complex<double>*)a, &lda,
                      (      std::complex<double>*)b, &ldb);
     }
+    template<>
+    void 
+    HostBlas<std::complex<double> >::gesv(int n, int rhs,
+                                         std::complex<double> *a, int lda, int *ipiv,
```
mhoemmen: I do find it weird that `HostBlas` is templated on `std::complex` but the instantiations use `Kokkos::complex`.

vqd8a:
Thanks @mhoemmen. Your point is well taken. However, we had a discussion whether we should separate ```gesv``` from BLAS functions in https://github.com/kokkos/kokkos-kernels/pull/409 when I started adding MAGMA to KokkosKernels some months ago. I think I agree with what @kyungjoo-kim said, ```gesv``` can be considered as extended BLAS concept and keeping ```gesv``` (maybe, and other LAPACK functions) within ```KokkosBlas``` might be easy for maintenance purpose. Moreover, MKL, OpenBLAS, and even MAGMA can provide both BLAS functionality and LAPACK functionality in the same libraries.
So, for the sake of time, I opt to keep it under BLAS for now. However, I am happy to make a change in the future if we need to.

src/impl/tpls/KokkosBlas_Host_tpl.cpp
vqd8a: Thanks, @mhoemmen . See my comment above.

src/impl/tpls/KokkosBlas_Host_tpl.cpp
vqd8a: Thanks, @mhoemmen . See my comment below.

src/impl/tpls/KokkosBlas_Host_tpl.cpp
vqd8a: Thanks, @mhoemmen . See my comment below.

src/impl/tpls/KokkosBlas_Host_tpl.cpp
vqd8a: Thanks @mhoemmen. I am just afraid if I make change, we also need changes in other BLAS functions. After talking with @kyungjoo-kim , I think I prefer this way, just to make it consistent with other BLAS functions. Moreover, so far we have not seen any binding issue during our spotchecks. If users have any issue with these, I am willing to help them.

src/impl/tpls/KokkosBlas_gesv_tpl_spec_decl.hpp
srajama1: I also recommend making this change @kyungjoo-kim suggested.

srajama1:
@vqd8a I see you have already made the changes suggested by @kyungjoo-kim . I also like the idea of separating LAPACK, but we can do it later. Can you file a github issue so we deal with this ? This requires approval by @kyungjoo-kim and @mhoemmen before I can merge.

vqd8a:
Thanks, @srajama1 @kyungjoo-kim . I will take care of the wiki. 
I am running spotcheck and will post the results when it is done.

mhoemmen:
My review got dismissed and/or turned into comments, so I don't need to approve again.

vqd8a:
Spotcheck passed on White:

```../Kokkos/kokkos-kernels/scripts/test_all_sandia gcc --spot-check --with-tpls=blas

Running on machine: white
Going to test compilers:  gcc/6.4.0 gcc/7.2.0
Testing compiler gcc/6.4.0
Testing compiler gcc/7.2.0
PASSED TESTS
gcc-6.4.0-OpenMP_Serial-release build_time=607 run_time=290
gcc-7.2.0-OpenMP-release build_time=363 run_time=134
gcc-7.2.0-OpenMP_Serial-release build_time=473 run_time=263
gcc-7.2.0-Serial-release build_time=263 run_time=187
FAILED TESTS
```

vqd8a:
Run spotcheck again on apollo with magma

```
nohup ../kokkos-kernels/scripts/test_all_sandia cuda/10.0 --spot-check --with-cuda-options=enable_lambda --with-tpls=magma

Running on machine: apollo
Going to test compilers:  cuda/10.0
Testing compiler cuda/10.0
  Starting job cuda-10.0-Cuda_OpenMP-release
  PASSED cuda-10.0-Cuda_OpenMP-release
PASSED TESTS
cuda-10.0-Cuda_OpenMP-release build_time=848 run_time=227
FAILED TESTS
```

PR #438, opened 2019-06-23
BLAS IAMAX support

vqd8a:
Added Iamax support (KK implementation, TPL BLAS, TPL cuBLAS).


src/blas/KokkosBlas1_iamax.hpp
```diff
@@ -0,0 +1,159 @@
...
+  return result;
+}
+
+/// \brief R(j) = iamax(X(i,j))
+///
+/// Replace each entry in R with the (smallest) index of the element of the maximum magnitude of the
+/// corresponding entry in X.
+///
+/// \tparam RMV 1-D or 2-D Kokkos::View specialization.
+/// \tparam XMV 1-D or 2-D Kokkos::View specialization.  It must have
```
mhoemmen: The documentation here doesn't match the "brief" description.  R can't possibly have the same rank as X.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+    KOKKOS_INLINE_FUNCTION
+    void operator() (const member_type &teamMember) const
+    {
+      const int lid = teamMember.league_rank();// teamId
+      const int tid = teamMember.team_rank(); // threadId
+
+      maxloc_type col_maxloc;
+      Kokkos::parallel_reduce( Kokkos::TeamThreadRange(teamMember, m_x.extent(0)), [&] (const int& i, maxloc_type& thread_lmaxloc) {
```
mhoemmen: We don't need to pass `int` by reference.  `int i` is fine.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
```diff
@@ -0,0 +1,231 @@
...
+    auto X_0 = Kokkos::subview (X, Kokkos::ALL (), 0);
+    typedef decltype (X_0) XV1D;
+    V_Iamax_Invoke<RV0D, XV1D, SizeType> (r_0, X_0);
+  }
+  else {
+    typedef Kokkos::View<typename RV::non_const_value_type *,
+                         typename RV::array_layout,
+                         typename XMV::device_type> RV_D;
+    typedef MV_Iamax_FunctorVector<RV_D, XMV, mag_type, SizeType> functor_type;
+    RV_D r_d("r_d", r.extent(0));
```
mhoemmen: Why do we need to allocate temporary storage here?  Is the issue that `r` must be host storage?

BLAS `IxAMAX` functions only work with 1-D input arrays, so unless we have a use case for the 2-D case, why not just get rid of it and thereby reduce code complexity?

vqd8a: @mhoemmen yes,  ```r``` is in host space. But, actually, we do not have a use case for the 2-D. I just added this for the sake of completeness. Of course, I can remove this 2-D case.

mhoemmen: The 2-D case could be useful for panel factorizations, but in that case, `r` would always be on device.

vqd8a: @mhoemmen So, I need to change the multi-vector template such that ```r``` is on device. I wasn't sure if ```r``` should be on device or on host.

mhoemmen: As long as these functions never allocate Views, I'm happy.  You have a View, so you can check its memory space accessibility.  The 1-D input case can either return a value (on host) or write to a 0-D View (on device).

src/blas/KokkosBlas1_iamax.hpp
vqd8a: Thanks @mhoemmen . It is fixed.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
vqd8a: @mhoemmen It is fixed.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
vqd8a: Thanks @mhoemmen. The code is fixed accordingly. Basically, there is no view allocation.

vqd8a:
The code was modified such that results can be on device for TPL cuBLAS  and it can either return a value on host or write to a 0-D View on device for the 1-D input case.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+    KOKKOS_INLINE_FUNCTION
+    void operator() (const member_type &teamMember) const
+    {
+      const int lid = teamMember.league_rank();// teamId
+      const int tid = teamMember.team_rank(); // threadId
+
+      maxloc_type col_maxloc;
+      Kokkos::parallel_reduce( Kokkos::TeamThreadRange(teamMember, m_x.extent(0)), [&] (const int i, maxloc_type& thread_lmaxloc) {
+        mag_type val = IPT::norm (m_x(i,lid));
```
kyungjoo-kim: @vqd8a What you are doing is that parallel reduce along a vector, which is typically large compared to the number of columns (extent(1)). Logically this can do the job but will be extremely slow comparable to a serial version. You should put the most outer parallel loop for the biggest work chunk.

src/blas/KokkosBlas1_iamax.hpp
```diff
@@ -90,10 +90,10 @@ typename XVector::size_type iamax (const XVector& x)
...
 /// corresponding entry in X.
 ///
-/// \tparam RMV 1-D or 2-D Kokkos::View specialization.
-/// \tparam XMV 1-D or 2-D Kokkos::View specialization.  It must have
-///   the same rank as RMV, and its entries must be assignable to
-///   those of RMV.
+/// \tparam RMV 0-D or 1-D Kokkos::View specialization.
+/// \tparam XMV 1-D or 2-D Kokkos::View specialization.
+///
+/// Special note for TPL cuBLAS:  RMV must be 0-D view and XMV must be 1-D view, and the index returned in RMV is 1-based since cuBLAS uses 1-based indexing for compatibility with Fortran
```
mhoemmen: The index needs to be the same, regardless of the implementation.  If that means subtracting 1 to convert from 1-based to 0-based indexing, then please do so.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
```diff
@@ -140,7 +144,7 @@ struct MV_Iamax_FunctorVector
       const int tid = teamMember.team_rank(); // threadId
 
       maxloc_type col_maxloc;
-      Kokkos::parallel_reduce( Kokkos::TeamThreadRange(teamMember, m_x.extent(0)), [&] (const int& i, maxloc_type& thread_lmaxloc) {
+      Kokkos::parallel_reduce( Kokkos::TeamThreadRange(teamMember, m_x.extent(0)), [&] (const int i, maxloc_type& thread_lmaxloc) {
```
mhoemmen: Please label all kernels; thanks!

unit_test/blas/Test_Blas1_iamax.hpp
```diff
@@ -46,12 +46,49 @@ namespace Test {
...
+#endif
+
+    ASSERT_EQ( nonconst_max_loc, expected_max_loc);
 
+    KokkosBlas::iamax(r,c_a);
+    Kokkos::deep_copy(h_r,r);
+
+#ifdef KOKKOSKERNELS_ENABLE_TPL_CUBLAS
+    if(std::is_same<typename Device::memory_space,Kokkos::CudaSpace>::value)
+      const_max_loc = h_r()-1;
```
mhoemmen: See above.

src/blas/KokkosBlas1_iamax.hpp
vqd8a: @mhoemmen Understand. I already did the subtraction when the cuBLAS functions return a value on host. 
But when the cuBLAS functions return directly the value on device memory, I do not know how to the subtraction efficiently. I do not want to call a kernel with only one thread for just doing the change for a single value so I leave the subtraction for user's kernel.
What would be the good way to do it? Can I just return the value to host and copy it to device if RMV is 0-D view?

mhoemmen: This is a kokkos-kernels design decision.  I would say, if the BLAS returns a 1-based index, then kokkos-kernels should return a 1-based index.  Kokkos users who want a 0-based index could easily implement their own, using `Kokkos::MaxLoc`.  I'm almost certain that a single kernel that kokkos-kernels provide would be faster than calling cuBLAS and then invoking another kernel just to decrement a single value.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
vqd8a: @kyungjoo-kim Thanks. I have fixed the implementation as per your suggestion. Please take a look.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
vqd8a: @mhoemmen I just did.

src/blas/KokkosBlas1_iamax.hpp
vqd8a: @srajama1 What is your opinion? Should we use 0-based index or 1-based index?

srajama1: There is no good answer for this in the decades we have dealt with it. I recommend to do whatever is most efficient for our users, we can jump through hoops to help them.

vqd8a: @srajama1 @mhoemmen For now, I think I would use 0-based index for all cases, except when TPL cuBLAS iamax is used and returns result to a 0-D view on device memory. I would leave users doing the 1-based to 0-based conversion in their kernels if they need to. I added note to specify this case in the ```KokkosBlas1_iamax.hpp```

However, I am open to changing to 0-based indexing for cuBLAS or using 1-based indexing for all cases. Please let me know.

src/blas/KokkosBlas1_iamax.hpp
```diff
@@ -136,7 +136,10 @@ iamax (const RV& R, const XMV& X,
       typename RV::non_const_value_type,
       typename RV::non_const_value_type* >::type,
     typename KokkosKernels::Impl::GetUnifiedLayout<RV>::array_layout,
-    typename RV::device_type,
+    typename Kokkos::Impl::if_c<
```
mhoemmen: Use `std::conditional` (lives in `<type_traits>`).  It works just like `if_c` here.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
```diff
@@ -63,7 +63,7 @@ struct V_Iamax_Functor
   typedef MagType                                                  mag_type;
   typedef typename XV::non_const_value_type                        xvalue_type;
   typedef Kokkos::Details::InnerProductSpaceTraits<xvalue_type>    IPT;
-  typedef typename Kokkos::MaxLoc<mag_type,size_type>::value_type  maxloc_type;
+  typedef typename RV::value_type                                  value_type;
```
mhoemmen: Prefer `using` alias syntax for new code.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
```diff
@@ -86,10 +86,36 @@ struct V_Iamax_Functor
                    "RV must have rank 0 and XV must have rank 1.");
   }
 
-  KOKKOS_INLINE_FUNCTION
-  void operator()( const size_type & i, maxloc_type & lmaxloc ) const {
-	mag_type val = IPT::norm (m_x(i));
-    if( val > lmaxloc.val ) { lmaxloc.val = val; lmaxloc.loc = i; }
+  KOKKOS_INLINE_FUNCTION void
```
mhoemmen: You should be able to use `Kokkos::MaxLoc` to replace `init` and `join` here.

src/blas/KokkosBlas1_iamax.hpp
vqd8a: @mhoemmen I am curious. why don't we just use if_c?

ndellingwood: @vqd8a I'm pretty sure `if_c` is Kokkos' pre-c++11 implementation of `std::conditional`, and some future cleanup in Kokkos will likely remove `if_c` so if nothing else it saves work later making the change; in general it is better to use standard library implementations when feasible.

mhoemmen: 1. Prefer Standard Library features to Kokkos features that do the same thing.
2. The feature `if_c` has that `std::conditional` lacks is the `select` method.

vqd8a: Okay. Thanks @ndellingwood and @mhoemmen for clarifying.

vqd8a: @mhoemmen ```std::conditional``` is used now.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
vqd8a: @mhoemmen ```using``` is used instead of ```typedef``` now.

src/blas/impl/KokkosBlas1_iamax_impl.hpp
vqd8a: @mhoemmen I used ```Kokkos::MaxLoc``` in my previous commit. Since ```iamax``` returns only the index, I found that ```Kokkos::MaxLoc``` works well for the 1-D input case when the index result returned to host. But, when the index result returned to a 0-D view on device, I need an extra copy the ```Kokkos::MaxLoc```'s ```loc``` result from host to device.
A custom reduction with the index as the reduction value can help in this case.

mhoemmen: Sure, that's ok -- thanks for thinking about it, though.

src/blas/KokkosBlas1_iamax.hpp
srajama1: Can we add a note in the wiki, explaining this ?

srajama1:
This won't let me merge unless @mhoemmen @kyungjoo-kim and @ndellingwood approve.

src/blas/KokkosBlas1_iamax.hpp
vqd8a: Thanks, @srajama1  I will take care of the wiki.
@mhoemmen Could you please approve this so that I can run spotcheck and get it merged?

vqd8a:
Spotchecks passed
```
../Kokkos/kokkos-kernels/scripts/test_all_sandia --spot-check --with-cuda-options=enable_lambda

Running on machine: white
Going to test compilers:  gcc/6.4.0 gcc/7.2.0 ibm/16.1.0 cuda/9.2.88 cuda/10.0.130
Testing compiler gcc/6.4.0
Testing compiler gcc/7.2.0
Testing compiler ibm/16.1.0
  Starting job ibm-16.1.0-Serial-release
  PASSED ibm-16.1.0-Serial-release
Testing compiler cuda/9.2.88
Testing compiler cuda/10.0.130
  Starting job cuda-10.0.130-Cuda_Serial-release
  PASSED cuda-10.0.130-Cuda_Serial-release
PASSED TESTS
cuda-10.0.130-Cuda_Serial-release build_time=952 run_time=340
cuda-9.2.88-Cuda_OpenMP-release build_time=1056 run_time=279
gcc-6.4.0-OpenMP_Serial-release build_time=475 run_time=309
gcc-7.2.0-OpenMP-release build_time=359 run_time=126
gcc-7.2.0-OpenMP_Serial-release build_time=582 run_time=365
gcc-7.2.0-Serial-release build_time=216 run_time=229
ibm-16.1.0-Serial-release build_time=1038 run_time=263
FAILED TESTS


../Kokkos/kokkos-kernels/scripts/test_all_sandia cuda --spot-check --with-cuda-options=enable_lambda --with-tpls=cublas

Running on machine: white
Going to test compilers:  cuda/9.2.88 cuda/10.0.130
Testing compiler cuda/9.2.88
Testing compiler cuda/10.0.130
  Starting job cuda-10.0.130-Cuda_Serial-release
  PASSED cuda-10.0.130-Cuda_Serial-release
PASSED TESTS
cuda-10.0.130-Cuda_Serial-release build_time=1006 run_time=341
cuda-9.2.88-Cuda_OpenMP-release build_time=1032 run_time=280
FAILED TESTS


../Kokkos/kokkos-kernels/scripts/test_all_sandia gcc --spot-check --with-tpls=blas

Running on machine: white
Going to test compilers:  gcc/6.4.0 gcc/7.2.0
Testing compiler gcc/6.4.0
Testing compiler gcc/7.2.0
PASSED TESTS
gcc-6.4.0-OpenMP_Serial-release build_time=531 run_time=277
gcc-7.2.0-OpenMP-release build_time=315 run_time=135
gcc-7.2.0-OpenMP_Serial-release build_time=383 run_time=253
gcc-7.2.0-Serial-release build_time=213 run_time=162
FAILED TESTS
```

srajama1:
Merging is blocked because of changes requested. At least in my browser it says changes requested from @mhoemmen

vqd8a:
@mhoemmen As said above, for now, I opt to use 0-based index for all cases, except when TPL cuBLAS iamax is used and returns result to a 0-D view on device memory (1-based index). 
In this 1-based case, users can use the result in their kernels (have to decrement by 1 before using it in the same kernel). As said by @srajama1, I will add a note to the wiki to explain the detail.
In the future, if needed, I am willing to change to using 1-based indexing for all cases.

unit_test/blas/Test_Blas1_iamax.hpp
vqd8a: Please see my comment below.

mhoemmen:
You should be able to dismiss my review, but if you don't know how to do that, I'll do it.

srajama1:
@vqd8a I didn't quite appreciate that this is just for IAMAX. That seems weird and I am worried now. Why is this exception needed ?

vqd8a:
@srajama1 The issue right now is BLAS and cuBLAS ```iamax``` return a 1-based index while KokkosKernels implementation returns 0-based index. I wanted to use 0-based index for all. For BLAS and cuBLAS with on-host result, we can subtract the result by 1 easily. But when the cuBLAS function return directly the result to device memory, I do not know how to do the subtraction efficiently before giving it to users since they can do the subtraction when they need the index result. That is why I just left the subtraction for users. I did not think this is a big problem as long as we document it in the wiki.

As suggested by @mhoemmen, we can just use 1-based index for all cases. Modifying the KokkosKernels implementation is not difficult. I was just afraid that this can complicate the custom reducer a bit in the KokkosKernels implementation with many subtractions. 

However, to avoid confusions, I think I should change to 1-based indexing for for all cases.
Please let me fix the code this way and I will run the spotchecks again.

vqd8a:
Updated the code such that the returned value is 1-based indexing.
Run spotchecks again:
```
../Kokkos/kokkos-kernels/scripts/test_all_sandia --spot-check --with-cuda-options=enable_lambda

Running on machine: white
Going to test compilers:  gcc/6.4.0 gcc/7.2.0 ibm/16.1.0 cuda/9.2.88 cuda/10.0.130
Testing compiler gcc/6.4.0
Testing compiler gcc/7.2.0
Testing compiler ibm/16.1.0
  Starting job ibm-16.1.0-Serial-release
  PASSED ibm-16.1.0-Serial-release
Testing compiler cuda/9.2.88
Testing compiler cuda/10.0.130
  Starting job cuda-10.0.130-Cuda_Serial-release
  PASSED cuda-10.0.130-Cuda_Serial-release
PASSED TESTS
cuda-10.0.130-Cuda_Serial-release build_time=957 run_time=342
cuda-9.2.88-Cuda_OpenMP-release build_time=955 run_time=278
gcc-6.4.0-OpenMP_Serial-release build_time=458 run_time=305
gcc-7.2.0-OpenMP-release build_time=371 run_time=110
gcc-7.2.0-OpenMP_Serial-release build_time=527 run_time=313
gcc-7.2.0-Serial-release build_time=211 run_time=175
ibm-16.1.0-Serial-release build_time=964 run_time=263
FAILED TESTS

../Kokkos/kokkos-kernels/scripts/test_all_sandia cuda --spot-check --with-cuda-options=enable_lambda --with-tpls=cublas

Running on machine: white
Going to test compilers:  cuda/9.2.88 cuda/10.0.130
Testing compiler cuda/9.2.88
Testing compiler cuda/10.0.130
  Starting job cuda-10.0.130-Cuda_Serial-release
  PASSED cuda-10.0.130-Cuda_Serial-release
PASSED TESTS
cuda-10.0.130-Cuda_Serial-release build_time=991 run_time=342
cuda-9.2.88-Cuda_OpenMP-release build_time=1001 run_time=278
FAILED TESTS

../Kokkos/kokkos-kernels/scripts/test_all_sandia gcc --spot-check --with-tpls=blas

Running on machine: white
Going to test compilers:  gcc/6.4.0 gcc/7.2.0
Testing compiler gcc/6.4.0
Testing compiler gcc/7.2.0
PASSED TESTS
gcc-6.4.0-OpenMP_Serial-release build_time=437 run_time=302
gcc-7.2.0-OpenMP-release build_time=335 run_time=113
gcc-7.2.0-OpenMP_Serial-release build_time=341 run_time=256
gcc-7.2.0-Serial-release build_time=196 run_time=183
FAILED TESTS
```

PR #439, opened 2019-06-24
sptrsv: Fix for team_size

ndellingwood:
Detected during 2.9 Trilinos integration testing


ndellingwood:
Quick fix to replace hard-coded team_size with AUTO, needed when deprecated code disabled in some configurations.
Detected during 2.9 testing, this PR was tested in Cuda+RDC configuration on Ride with Kepler GPUs.

PR #470, opened 2019-09-20
KokkosBatched - vector length adjustment for ARM

kyungjoo-kim:
This set a vector length (4 doubles) for ARM architectures. For Ifpack2 block tridiagonal solver, using 256 bit length is better than 128 bit length. For quick test, I already submit a PR for Trilinos and this is a matching PR in KokkosKernels.


kyungjoo-kim:
This does not require spotcheck. None of those architecture can test ARM.

srajama1:
I am pushing these. Thanks ! How do you test these ? Should we add a spot-check or at least a nightly on the platform you are using ? Let us discuss this in today's meeting.

kyungjoo-kim:
I think that we should automate this spot-check as like PR test in Trilinos. I use mayer for testing.

Issue #474, opened 2019-09-24
Wrong team size heuristic used for SPGEMM when Kokkos deprecated=OFF


brian-kelley:
Follow up and closely related to #468. This issue is for discussing a more permanent fix than #472, which fixed all the unit test failures but still chooses very sub-optimal team sizes on CUDA.

PR #472, opened 2019-09-24
With kokkos deprecated off, fix sugg teamsize = 0 (#468)

brian-kelley:
With Kokkos deprecated off, the KokkosKernels suggested team size was sometimes 0 (especially for small matrices, like in some unit tests). A team of size 0 obviously can't do any work, so this commit just makes sure the #teams is at least 1. This was caught because some of the symmetrize utility functions divide by the team size to divide up work, causing div-by-zero/floating point exceptions in the sparse and graph unittests.

This fixes #468.

Bowman spot check:
#######################################################
PASSED TESTS
#######################################################
intel-16.4.258-Pthread-release build_time=1521 run_time=587
intel-16.4.258-Pthread_Serial-release build_time=2472 run_time=1250
intel-16.4.258-Serial-release build_time=1470 run_time=622
intel-17.2.174-OpenMP-release build_time=1987 run_time=393
intel-17.2.174-OpenMP_Serial-release build_time=2841 run_time=1120
intel-17.2.174-Pthread-release build_time=1322 run_time=623
intel-17.2.174-Pthread_Serial-release build_time=2393 run_time=1268
intel-17.2.174-Serial-release build_time=1380 run_time=665
intel-18.2.199-OpenMP-release build_time=1590 run_time=436
intel-18.2.199-OpenMP_Serial-release build_time=2739 run_time=1049
intel-18.2.199-Pthread-release build_time=1230 run_time=635
intel-18.2.199-Pthread_Serial-release build_time=2357 run_time=1218
intel-18.2.199-Serial-release build_time=1210 run_time=671
#######################################################
FAILED TESTS
#######################################################

White spot check:
#######################################################
PASSED TESTS
#######################################################
cuda-10.0.130-Cuda_Serial-release build_time=1169 run_time=370
cuda-9.2.88-Cuda_OpenMP-release build_time=1351 run_time=305
gcc-6.4.0-OpenMP_Serial-release build_time=709 run_time=327
gcc-7.2.0-OpenMP-release build_time=466 run_time=145
gcc-7.2.0-OpenMP_Serial-release build_time=809 run_time=383
gcc-7.2.0-Serial-release build_time=277 run_time=165
ibm-16.1.0-Serial-release build_time=1514 run_time=272
#######################################################
FAILED TESTS
#######################################################

Kokkos-dev spot check:
#######################################################
PASSED TESTS
#######################################################
clang-4.0.1-Pthread_Serial-hwloc-release build_time=728 run_time=381
[...]

PR #476, opened 2019-09-27
Use up-to-date Kokkos features to get team size

brian-kelley:
A permanent fix for #474, replacing the temporary one in #472. When Kokkos deprecated is off and the execution space is CUDA, ``KokkosKernels::Impl::get_suggested_team_size()`` uses ``TeamPolicy::team_size_recommmended()``, which calls ``cudaOccupancyMaxActiveBlocksPerMultiprocessor()``. This should be a good heuristic, and will definitely never produce a team size too high to launch, since the CUDA runtime knows exactly how many registers a kernel uses.

This change doesn't affect SPGEMM, SPMV or Blas. ``KokkosKernels::Impl::get_suggested_team_size()`` is only used in the utility functions ``symmetrize_and_get_lower_diagonal_edge_list()`` and ``symmetrize_graph_symbolic_hashmap()``, as well as ``get_lower_diagonal_edge_list()`` used during Dist1 coloring setup.

Bowman spot check:
PASSED TESTS
intel-16.4.258-Pthread-release build_time=1507 run_time=585
intel-16.4.258-Pthread_Serial-release build_time=2443 run_time=1248
intel-16.4.258-Serial-release build_time=1459 run_time=618
intel-17.2.174-OpenMP-release build_time=1951 run_time=371
intel-17.2.174-OpenMP_Serial-release build_time=2835 run_time=1136
intel-17.2.174-Pthread-release build_time=1305 run_time=630
intel-17.2.174-Pthread_Serial-release build_time=2381 run_time=1269
intel-17.2.174-Serial-release build_time=1388 run_time=660
intel-18.2.199-OpenMP-release build_time=1587 run_time=420
intel-18.2.199-OpenMP_Serial-release build_time=2761 run_time=1008
intel-18.2.199-Pthread-release build_time=1208 run_time=625
intel-18.2.199-Pthread_Serial-release build_time=2363 run_time=1226
intel-18.2.199-Serial-release build_time=1188 run_time=654
FAILED TESTS

White spot check:
PASSED TESTS
cuda-10.0.130-Cuda_Serial-release build_time=1245 run_time=369
cuda-9.2.88-Cuda_OpenMP-release build_time=1197 run_time=305
gcc-6.4.0-OpenMP_Serial-release build_time=635 run_time=366
gcc-7.2.0-OpenMP-release build_time=515 run_time=146
gcc-7.2.0-OpenMP_Serial-release build_time=665 run_time=448
gcc-7.2.0-Serial-release build_time=285 run_time=212
ibm-16.1.0-Serial-release build_time=1398 run_time=271
FAILED TESTS

kokkos-dev spot check:
PASSED TESTS
clang-4.0.1-Pthread_Serial-hwloc-release build_time=739 run_time=367
clang-4.0.1-Pthread_Serial-release build_time=769 run_time=606
cuda-8.0.44-Cuda_OpenMP-release build_time=1265 run_time=339
gcc-5.3.0-Serial-hwloc-release build_time=506 run_time=191
gcc-5.3.0-Serial-release build_time=504 run_time=214
gcc-7.2.0-Serial-hwloc-release build_time=437 run_time=188
gcc-7.2.0-Serial-release build_time=433 run_time=168
intel-17.0.1-OpenMP-hwloc-release build_time=980 run_time=165
intel-17.0.1-OpenMP-release build_time=950 run_time=123
FAILED TESTS

### DIFF
--- a/perf_test/sparse/KokkosSparse_run_spgemm.hpp
+++ b/perf_test/sparse/KokkosSparse_run_spgemm.hpp
@@ -319,8 +319,9 @@ crsMat_t3 run_experiment(
 	  row_mapC = lno_view_t
 			  ("non_const_lnow_row",
 					  m + 1);
-	  entriesC = lno_nnz_view_t ();
-	  valuesC = scalar_view_t ();
+	  entriesC = lno_nnz_view_t ("entriesC (empty)", 0);
+	  valuesC = scalar_view_t ("valuesC (empty)", 0);
+
 
 	  Kokkos::Impl::Timer timer1;
 	  spgemm_symbolic (
--- a/perf_test/sparse/KokkosSparse_run_spgemm.hpp
+++ b/perf_test/sparse/KokkosSparse_run_spgemm.hpp
@@ -319,8 +319,8 @@ crsMat_t3 run_experiment(
 	  row_mapC = lno_view_t
 			  ("non_const_lnow_row",
 					  m + 1);
-	  entriesC = lno_nnz_view_t ("");
-	  valuesC = scalar_view_t ("");
+	  entriesC = lno_nnz_view_t ();
+	  valuesC = scalar_view_t ();
 
 	  Kokkos::Impl::Timer timer1;
 	  spgemm_symbolic (
--- a/src/common/KokkosKernels_Utils.hpp
+++ b/src/common/KokkosKernels_Utils.hpp
@@ -212,49 +212,26 @@ void get_suggested_vector_size(
 
 }
 
-
-template <typename ExecutionSpace>
-void get_suggested_team_size(
-    int max_allowed_team_size,
-    int suggested_vector_size,
-    int &suggested_team_size_){
-
-    suggested_team_size_ = 1;
-
-#if defined( KOKKOS_ENABLE_SERIAL )
-  if (Kokkos::Impl::is_same< Kokkos::Serial , ExecutionSpace >::value){
-    suggested_team_size_ = 1;
+//Get the best team size for the given functor.
+//If it uses shared memory, the amount used must be available through f.team_shmem_size(n),
+//not through the TeamPolicy. If this is how dynamic shared is set, just use AUTO for the team size.
+template<typename team_policy_t, typename Functor, typename ParallelTag = Kokkos::ParallelForTag>
+int get_suggested_team_size(Functor& f, int vector_size)
+{
+#ifdef KOKKOS_ENABLE_CUDA
+  if(std::is_same<typename team_policy_t::traits::execution_space, Kokkos::Cuda>::value)
+  {
+    team_policy_t temp(1, 1, vector_size);
+    return temp.team_size_recommended(f, ParallelTag());
   }
+  else
 #endif
-
-#if defined( KOKKOS_ENABLE_THREADS )
-  if (Kokkos::Impl::is_same< Kokkos::Threads , ExecutionSpace >::value){
-    suggested_team_size_ =  1;
+  {
+    return 1;
   }
-#endif
-
-#if defined( KOKKOS_ENABLE_OPENMP )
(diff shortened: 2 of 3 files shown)


perf_test/sparse/KokkosSparse_run_spgemm.hpp
```diff
@@ -319,8 +319,8 @@ crsMat_t3 run_experiment(
 	  row_mapC = lno_view_t
 			  ("non_const_lnow_row",
 					  m + 1);
-	  entriesC = lno_nnz_view_t ("");
-	  valuesC = scalar_view_t ("");
+	  entriesC = lno_nnz_view_t ();
```
srajama1: Simple request, can we name it instead of removing it ? Will help in debugging later.

src/common/KokkosKernels_Utils.hpp
```diff
@@ -212,49 +212,26 @@ void get_suggested_vector_size(
...
-#if defined( KOKKOS_ENABLE_CUDA )
-  if (Kokkos::Impl::is_same<Kokkos::Cuda, ExecutionSpace >::value){
-    suggested_team_size_ = max_allowed_team_size / suggested_vector_size;
-    if(suggested_team_size_ == 0)
-      suggested_team_size_ = 1;
+//Get the best team size for the given functor.
+//If it uses shared memory, the amount used must be available through f.team_shmem_size(n),
+//not through the TeamPolicy. If this is how dynamic shared is set, just use AUTO for the team size.
+template<typename team_policy_t, typename Functor, typename ParallelTag = Kokkos::ParallelForTag>
+int get_suggested_team_size(Functor& f, int vector_size)
```
srajama1: Do we still use vector_size ?

brian-kelley: Yes, all the kernels that use this first get a preferred vector size (in general, 1 on CPU and 32 on CUDA), and then find a team size so that teamSize * vectorSize = a reasonable CUDA block size. A kernel that doesn't use ThreadVector loops would just pass in 1.

perf_test/sparse/KokkosSparse_run_spgemm.hpp
brian-kelley: Sure, I just added back the labels and allocated them to size 0. After this change I rebuilt everything and ran the SPGEMM driver on my local machine (CUDA/OpenMP), with and without deprecated.

brian-kelley:
@crtrott Does this look good to you? The main thing is just the new definition of get_suggested_team_size for when deprecated is off:
```
template<typename team_policy_t, typename Functor, typename ParallelTag = Kokkos::ParallelForTag>
int get_suggested_team_size(Functor& f, int vector_size)
{
#ifdef KOKKOS_ENABLE_CUDA
  if(std::is_same<typename team_policy_t::traits::execution_space, Kokkos::Cuda>::value)
  {
    team_policy_t temp(1, 1, vector_size);
    return temp.team_size_recommended(f, ParallelTag());
  }
  else
#endif
  {
    return 1;
  }
}
```

srajama1:
@crtrott What do you think of this change ? @ndellingwood merged it, but I would still like your opinion.

crtrott:
That looks good with the exception of it assuming you are using no scratch memory.

brian-kelley:
@srajama1 None of the functors that get passed to ``get_suggested_teamsize`` use shared memory (currently). The deprecated function that this replaces also doesn't take shared into account. But I could easily add an overload for this that takes shared per team, and per thread. Would that be useful?

crtrott:
Might be useful but yeah if you call it only with stuff which doesn't use shared memory this is good.

srajama1:
I think this is good for now. We need to clearly state (comment ?) that this cannot be used with shared memory. Let us write that one separately.

Issue #488, opened 2019-10-31
Feature request : Support multiple rhs in SGS efficiently


srajama1:
Feature request : <edited based on the discussion below> Support multiple rhs in MT-SGS efficiently.  Currently,  it is implemented but it could be better if we pass the multivector to Kokkos Kernels, which accepts and iterates on all the vectors. Based on comment by @mhoemmen serial SGS also does the same.

Customer:  Exawind

@brian-kelley : One more thing when you are at SGS.

@iyamaza found this was not implemented. @lucbv

mhoemmen:
It's implemented, but it's slow (it just loops over columns on the outside).  I'm pretty sure Ichi was talking about the not-multithreaded case.

brian-kelley:
@mhoemmen Are you talking about inside Ifpack2 Relaxation? It looks like Relaxation::MTGaussSeidel() does a loop over vectors. There would be a significant win there from moving that loop all the way inside the KK SGS implementation, from fewer kernel launches and better locality with the matrix accesses.

lucbv:
@brian-kelley You would probably gain time on a GPU by not having to load the matrix in memory multiple times.

mhoemmen:
@iyamazaki Are we talking about MTGaussSeidel or the non-MT Gauss-Seidel?

iyamazaki:
Hi, @mhoemmen. I was looking at MTGaussSeidel, which seems to loop over the vectors in Ifpack2? But, I did not look at what the KokkosKernel does.

srajama1:
I checked the MT-SGS case and I should have been clear in the original posting. It is implemented but it not in an optimal way.

mhoemmen:
Non-MT Gauss-Seidel has the same issue but it's separate code.

brian-kelley:
Within KokkosKernels MT-SGS multiple RHS doesn't exist at all. Within Ifpack2 the inefficient implementation (outermost loop over vectors) exists.

PR #502, opened 2019-11-25
Gauss-Seidel: rank-2 X/Y support

brian-kelley:
This PR adds rank-2 vector support to all variants of parallel Gauss-Seidel (point-coloring, block, cluster-coloring). This should be a significant speedup for Ifpack2 MTSGS with multivectors. Rank-1 support is not affected from the user's perspective.

The ETI for Gauss-Seidel apply() has been changed to just instantiate for a rank-2 apply (both LayoutLeft and LayoutRight, in the normal ETI configuration). As before, the user's X and Y vectors are internally converted to unmanaged views with the appropriate scalar, space and layout. If a rank-1 view is provided, the apply is done internally on an Nx1 multivector.

Test coverage for Gauss-Seidel has been improved - both Rank-1 and Rank-2 X/Y, and all apply types (forward/backward/symmetric) are now tested. The versions with lazy setup (skipSymbolic/skipNumeric) are no longer tested since they meant 4x as many runs were done in testing, and the logic in apply to call symbolic and/or numeric is very simple.

For CGS (cluster), every clustering algorithm (balloon, RCM and no-op) and 3 different cluster sizes are tested. The RangePolicy, TeamPolicy, Block and BigBlock versions of point MTGS are still all tested (assuming CUDA and at least one other CPU exec space are tested), as are the Range/Team versions of CGS.

(CUDA still broken on kokkos-dev)

Bowman:
PASSED TESTS
intel-16.4.258-Pthread-release build_time=1629 run_time=1659
intel-16.4.258-Pthread_Serial-release build_time=2750 run_time=3317
intel-16.4.258-Serial-release build_time=1580 run_time=1643
intel-17.2.174-OpenMP-release build_time=2125 run_time=765
intel-17.2.174-OpenMP_Serial-release build_time=3158 run_time=2867
intel-17.2.174-Pthread-release build_time=1428 run_time=1663
intel-17.2.174-Pthread_Serial-release build_time=2553 run_time=3253
intel-17.2.174-Serial-release build_time=1655 run_time=1618
intel-18.2.199-OpenMP-release build_time=1693 run_time=744
intel-18.2.199-OpenMP_Serial-release build_time=3284 run_time=2347
intel-18.2.199-Pthread-release build_time=1419 run_time=1695
intel-18.2.199-Pthread_Serial-release build_time=2879 run_time=3138
intel-18.2.199-Serial-release build_time=1365 run_time=1636
FAILED TESTS

White:
PASSED TESTS
cuda-10.0.130-Cuda_Serial-release build_time=1171 run_time=1402
cuda-9.2.88-Cuda_OpenMP-release build_time=1159 run_time=1179
gcc-6.4.0-OpenMP_Serial-release build_time=602 run_time=913
gcc-7.2.0-OpenMP-release build_time=395 run_time=389
gcc-7.2.0-OpenMP_Serial-release build_time=594 run_time=1238
gcc-7.2.0-Serial-release build_time=265 run_time=443
gcc-7.4.0-OpenMP-release build_time=400 run_time=387
ibm-16.1.0-Serial-release build_time=1337 run_time=863
FAILED TESTS

### DIFF
--- a/src/common/KokkosKernels_Handle.hpp
+++ b/src/common/KokkosKernels_Handle.hpp
@@ -195,6 +195,7 @@ public:
   typedef typename size_type_persistent_work_view_t::HostMirror size_type_persistent_work_host_view_t; //Host view type
   typedef typename Kokkos::View<nnz_scalar_t *, HandleTempMemorySpace> scalar_temp_work_view_t;
   typedef typename Kokkos::View<nnz_scalar_t *, HandlePersistentMemorySpace> scalar_persistent_work_view_t;
+  typedef typename Kokkos::View<nnz_scalar_t **, Kokkos::LayoutLeft, HandlePersistentMemorySpace> scalar_persistent_work_view2d_t;
   typedef typename Kokkos::View<nnz_lno_t *, HandleTempMemorySpace> nnz_lno_temp_work_view_t;
   typedef typename Kokkos::View<nnz_lno_t *, HandlePersistentMemorySpace> nnz_lno_persistent_work_view_t;
   typedef typename nnz_lno_persistent_work_view_t::HostMirror nnz_lno_persistent_work_host_view_t; //Host view type
--- a/src/sparse/KokkosSparse_gauss_seidel.hpp
+++ b/src/sparse/KokkosSparse_gauss_seidel.hpp
@@ -334,8 +334,21 @@ namespace KokkosSparse{
                      typename x_scalar_view_t::value_type>::value,
                      "KokkosSparse::symmetric_gauss_seidel_apply: scalar type of the x-vector should be same as kernelHandle non-const scalar_t.");
 
+      static_assert (!std::is_same<typename lno_row_view_t_::array_layout, Kokkos::LayoutStride>::value,
+                     "KokkosSparse::symmetric_gauss_seidel_apply: row_map must have a contiguous layout (Left or Right, not Stride)");
+      static_assert (!std::is_same<typename lno_nnz_view_t_::array_layout, Kokkos::LayoutStride>::value,
+                     "KokkosSparse::symmetric_gauss_seidel_apply: entries must have a contiguous layout (Left or Right, not Stride)");
+      static_assert (!std::is_same<typename scalar_nnz_view_t_::array_layout, Kokkos::LayoutStride>::value,
+                     "KokkosSparse::symmetric_gauss_seidel_apply: values must have a contiguous layout (Left or Right, not Stride)");
 
-
+      // Check compatibility of #vectors
+      if(x_lhs_output_vec.extent(1) != y_rhs_input_vec.extent(1))
+      {
+        std::ostringstream os;
+        os << "KokkosSparse::symmetric_gauss_seidel_apply: " <<
+          "X has " << x_lhs_output_vec.extent(1) << "columns, Y has " << y_rhs_input_vec.extent(1) << " columns.";
+        Kokkos::Impl::throw_runtime_exception (os.str ());
+      }
 
       typedef typename KernelHandle::const_size_type c_size_t;
       typedef typename KernelHandle::const_nnz_lno_t c_lno_t;
@@ -346,7 +359,6 @@ namespace KokkosSparse{
       typedef typename KernelHandle::HandlePersistentMemorySpace c_persist_t;
 
       typedef typename  KokkosKernels::Experimental::KokkosKernelsHandle<c_size_t, c_lno_t, c_scalar_t, c_exec_t, c_temp_t, c_persist_t> const_handle_type;
-      //const_handle_type tmp_handle = *handle;
       const_handle_type tmp_handle (*handle);
 
       typedef Kokkos::View<
@@ -367,47 +379,37 @@ namespace KokkosSparse{
         typename scalar_nnz_view_t_::device_type,
         Kokkos::MemoryTraits<Kokkos::Unmanaged> > Internal_ascalar_nnz_view_t_;
 
-
       typedef Kokkos::View<
-        typename y_scalar_view_t::const_value_type*,
+        typename y_scalar_view_t::const_value_type**,
         typename KokkosKernels::Impl::GetUnifiedLayout<y_scalar_view_t>::array_layout,
         typename y_scalar_view_t::device_type,
         Kokkos::MemoryTraits<Kokkos::Unmanaged> > Internal_yscalar_nnz_view_t_;
 
       typedef Kokkos::View<
-        typename x_scalar_view_t::non_const_value_type*,
+        typename x_scalar_view_t::non_const_value_type**,
         typename KokkosKernels::Impl::GetUnifiedLayout<x_scalar_view_t>::array_layout,
         typename x_scalar_view_t::device_type,
         Kokkos::MemoryTraits<Kokkos::Unmanaged> > Internal_xscalar_nnz_view_t_;
-
-      /*
-
-        Internal_alno_row_view_t_ const_a_r  = row_map;
-        Internal_alno_nnz_view_t_ const_a_l  = entries;
-        Internal_ascalar_nnz_view_t_ const_a_v  = values;
-        Internal_xscalar_nnz_view_t_ nonconst_x_v = x_lhs_output_vec;
-        Internal_yscalar_nnz_view_t_ const_y_v = y_rhs_input_vec;
-      */
 
       Internal_alno_row_view_t_ const_a_r (row_map.data(), row_map.extent(0));
       Internal_alno_nnz_view_t_ const_a_l (entries.data(), entries.extent(0));
       Internal_ascalar_nnz_view_t_ const_a_v (values.data(), values.extent(0));
 
-      Internal_xscalar_nnz_view_t_ nonconst_x_v (x_lhs_output_vec.data(), x_lhs_output_vec.extent(0));
-      Internal_yscalar_nnz_view_t_ const_y_v (y_rhs_input_vec.data(), y_rhs_input_vec.extent(0));
+      Internal_xscalar_nnz_view_t_ nonconst_x_v (x_lhs_output_vec.data(), x_lhs_output_vec.extent(0), x_lhs_output_vec.extent(1));
+      Internal_yscalar_nnz_view_t_ const_y_v (y_rhs_input_vec.data(), y_rhs_input_vec.extent(0), y_rhs_input_vec.extent(1));
 
       using namespace KokkosSparse::Impl;
 
       GAUSS_SEIDEL_APPLY<const_handle_type,
                          Internal_alno_row_view_t_, Internal_alno_nnz_view_t_, Internal_ascalar_nnz_view_t_,
                          Internal_xscalar_nnz_view_t_, Internal_yscalar_nnz_view_t_>::gauss_seidel_apply (
-                                                                                                          &tmp_handle, num_rows, num_cols,
-                                                                                                          const_a_r, const_a_l, const_a_v,
-                                                                                                          nonconst_x_v, const_y_v,
-                                                                                                          init_zero_x_vector,
-                                                                                                          update_y_vector,
-                                                                                                          omega,
-                                                                                                          numIter, true, true);
+                             &tmp_handle, num_rows, num_cols,
+                             const_a_r, const_a_l, const_a_v,
+                             nonconst_x_v, const_y_v,
+                             init_zero_x_vector,
+                             update_y_vector,
+                             omega,
+                             numIter, true, true);
 
 
     }
@@ -431,12 +433,22 @@ namespace KokkosSparse{
                                             bool init_zero_x_vector,
                                             bool update_y_vector,
                                             typename KernelHandle::nnz_scalar_t omega,
-                                            int numIter){
+                                            int numIter)
+    {
+      // Check compatibility of dimensions at run time.
+      if(x_lhs_output_vec.extent(1) != y_rhs_input_vec.extent(1))
+      {
+        std::ostringstream os;
+        os << "KokkosSparse::symmetric_block_gauss_seidel_apply: Dimensions of X and Y do not match: " <<
+          "X has " << x_lhs_output_vec.extent(1) << "columns, Y has " << y_rhs_input_vec.extent(1) << " columns.";
+        Kokkos::Impl::throw_runtime_exception (os.str ());
+      }
       auto gsHandle = handle->get_point_gs_handle();
       if(gsHandle->get_algorithm_type() == GS_CLUSTER)
       {
         throw std::runtime_error("Block versions of Gauss-Seidel are incompatible with algorithm GS_CLUSTER");
       }
+
       gsHandle->set_block_size(block_size);
       symmetric_gauss_seidel_apply(handle,num_rows,num_cols,
                                    row_map,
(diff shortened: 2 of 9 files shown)


src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
```diff
@@ -187,14 +187,17 @@ namespace KokkosSparse{
         {
           size_type row_begin = _xadj(row);
           size_type row_end = _xadj(row + 1);
-          nnz_scalar_t sum = _Yvector(row);
-          for (size_type adjind = row_begin; adjind < row_end; ++adjind)
+          for(nnz_lno_t vec = 0; vec < (nnz_lno_t) _Xvector.extent(1); vec++)
```
srajama1: I thought this will be the inner loop, seems like it is the outer loop and we read the matrix multiple times.

src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
```diff
@@ -287,19 +290,22 @@ namespace KokkosSparse{
...
-                  {
-                    size_type adjind = i + row_begin;
-                    nnz_lno_t colIndex = _adj(adjind);
-                    nnz_scalar_t val = _adj_vals(adjind);
-                    lsum += val * _Xvector(colIndex);
-                  }, dotprod);
-                Kokkos::single(Kokkos::PerThread(teamMember),[=] () {
-                    _Xvector(row) += _omega * (_Yvector(row) - dotprod) * _inverse_diagonal(row);
-                  });
+                for(nnz_lno_t vec = 0; vec < (nnz_lno_t) _Xvector.extent(1); vec++)
```
srajama1: Same as above

src/sparse/impl/KokkosSparse_gauss_seidel_impl.hpp
```diff
@@ -245,21 +250,22 @@ namespace KokkosSparse{
...
-                                    nnz_scalar_t val = _adj_vals(adjind);
-                                    valueToUpdate += val * _Xvector(colIndex);
-                                  },
-                                  product);
-
-          Kokkos::single(Kokkos::PerThread(teamMember),[=] () {
-              nnz_scalar_t invDiagonalVal = _permuted_inverse_diagonal(ii);
-              _Xvector(ii) += omega * (_Yvector(ii) - product) * invDiagonalVal;
-            });
+          for(nnz_lno_t vec = 0; vec < (nnz_lno_t) _Xvector.extent(1); vec++){
```
srajama1: Same as above

src/sparse/impl/KokkosSparse_gauss_seidel_spec.hpp
```diff
@@ -110,19 +110,19 @@ namespace KokkosSparse {
                                             KokkosKernels::Experimental::KokkosKernelsHandle< \
                                                                                              const OFFSET_TYPE, const ORDINAL_TYPE, const SCALAR_TYPE, \
                                                                                              EXEC_SPACE_TYPE, MEM_SPACE_TYPE, SLOW_MEM_SPACE> , \
-                                            Kokkos::View<const OFFSET_TYPE *, LAYOUT_TYPE, \
+                                            Kokkos::View<const OFFSET_TYPE *, Kokkos::LayoutLeft, \
```
mhoemmen: There should be no need to specialize on LayoutLeft vs. LayoutRight for rank-1 Views.  Kokkos::View permits free assignment between LayoutLeft and LayoutRight in that case.  We should thus just be able to leave off the layout and let the default take over.

src/sparse/impl/KokkosSparse_gauss_seidel_spec.hpp
```diff
@@ -110,19 +110,19 @@ namespace KokkosSparse {
...
                                                          Kokkos::Device<EXEC_SPACE_TYPE, MEM_SPACE_TYPE>, \
                                                          Kokkos::MemoryTraits<Kokkos::Unmanaged> >, \
-                                            Kokkos::View<const ORDINAL_TYPE *, LAYOUT_TYPE, \
+                                            Kokkos::View<const ORDINAL_TYPE *, Kokkos::LayoutLeft, \
                                                          Kokkos::Device<EXEC_SPACE_TYPE, MEM_SPACE_TYPE>, \
```
mhoemmen: I know this isn't your problem :) but the only operations that need to know about the memory space are those that allocate Views.  If something doesn't allocate a View, we should just treat all Views as nonowning AnonymousSpace.

unit_test/sparse/Test_Sparse_block_gauss_seidel.hpp
```diff
@@ -143,37 +139,33 @@ int run_block_gauss_seidel_1(
...
+template<typename vec_t>
+vec_t create_x_vector(vec_t& kok_x, double max_value = 10.0) {
+  typedef typename vec_t::value_type scalar_t;
+  auto h_x = Kokkos::create_mirror_view (kok_x);
+  for (size_t j = 0; j < h_x.extent(1); ++j){
+    for (size_t i = 0; i < h_x.extent(0); ++i){
+      scalar_t r =
+          static_cast <scalar_t> (rand()) /
+          static_cast <scalar_t> (RAND_MAX / max_value);
+      h_x.access(i, j) = r;
```
mhoemmen: Again, it's weird that you call `access` -- how come?

src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
brian-kelley: @srajama1 I would have liked to do that, but then I would need arbitrary-sized storage somewhere for the accumulated dot products in each column of X/Y. I could just allocate a view for that.

unit_test/sparse/Test_Sparse_block_gauss_seidel.hpp
brian-kelley: @mhoemmen Originally I wanted the implementation to support rank-1 and rank-2 inputs, but that's not actually necessary since I now wrap the user's view in a rank-2 no matter what. I'll fix this.

src/sparse/impl/KokkosSparse_gauss_seidel_spec.hpp
brian-kelley: @mhoemmen Right, before this PR the ETI unnecessarily instantiated apply for both Left and Right layout rowptrs/colinds/values. I always just take a LayoutLeft unmanged view of those now. I changed the ETI so that LAYOUT_TYPE is the layout of X and Y, since those are now rank-2.

src/sparse/KokkosSparse_gauss_seidel.hpp
```diff
@@ -521,50 +546,37 @@ namespace KokkosSparse{
...
-
       Internal_alno_row_view_t_ const_a_r (row_map.data(), row_map.extent(0));
       Internal_alno_nnz_view_t_ const_a_l (entries.data(), entries.extent(0));
       Internal_ascalar_nnz_view_t_ const_a_v (values.data(), values.extent(0));
 
-      Internal_xscalar_nnz_view_t_ nonconst_x_v (x_lhs_output_vec.data(), x_lhs_output_vec.extent(0));
-      Internal_yscalar_nnz_view_t_ const_y_v (y_rhs_input_vec.data(), y_rhs_input_vec.extent(0));
-
-
+      Internal_xscalar_nnz_view_t_ nonconst_x_v (x_lhs_output_vec.data(), x_lhs_output_vec.extent(0), x_lhs_output_vec.extent(1));
```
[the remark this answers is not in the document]
brian-kelley: @mhoemmen I don't know, that's just how it was before and I didn't think to try direct assignment. I'll see if that works.

src/common/KokkosKernels_Handle.hpp
```diff
@@ -195,6 +195,7 @@ class KokkosKernelsHandle
   typedef typename size_type_persistent_work_view_t::HostMirror size_type_persistent_work_host_view_t; //Host view type
   typedef typename Kokkos::View<nnz_scalar_t *, HandleTempMemorySpace> scalar_temp_work_view_t;
   typedef typename Kokkos::View<nnz_scalar_t *, HandlePersistentMemorySpace> scalar_persistent_work_view_t;
+  typedef typename Kokkos::View<nnz_scalar_t **, Kokkos::LayoutLeft, HandlePersistentMemorySpace> scalar_persistent_work_view2d_t;
```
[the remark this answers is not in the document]
brian-kelley: @mhoemmen Which operator= are you talking about? KokkosKernels::Handle doesn't have an explicit one (well, it looks like it did but it has been commented out). I don't think there's ever an occasion where you'd want to copy a handle directly.

src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
srajama1: Is it really arbitrary are is it size number of rows ? I believe reading the matrix once is going to help in terms of performance. If it is an extra array we could allocate it during symbolic and hold it in the handle.

brian-kelley: @srajama1 It's the number of columns in X and Y, which is arbitrary in that there's no constant upper bound. I don't know the largest it can be in practice. The idea is to allocate a view to hold the 'sum' accumulator, and this would be same dimensions as X. I couldn't allocate it during symbolic since the number of columns in X and Y isn't known until apply().

What if I just processed columns in fixed-size batches, say 8 columns at a time? At worst that's 1/8th the amount of matrix accesses as there is currently, and I wouldn't need to mess around with global or shared allocation at all.

srajama1: I like keeping it at 8 or a multiple of that. Please keep an internal variable for now.

src/sparse/KokkosSparse_gauss_seidel.hpp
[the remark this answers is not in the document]
brian-kelley: @mhoemmen It actually didn't work, because in the rank-1 case the implicit conversion from LayoutRight to LayoutLeft isn't allowed.

src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
mhoemmen: @brian-kelley wrote:
> What if I just processed columns in fixed-size batches, say 8 columns at a time?

I think the sparse matrix-vector multiply kernel unrolls up to 16 (or 17 -- ask @crtrott about that ;-) ).  You can decide how to do this per kernel, though it might make sense to do some measurements and figure out where the knee of the speedup curve happens (or how much the optimizations increase build times and sizes).

src/common/KokkosKernels_Handle.hpp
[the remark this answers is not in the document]
mhoemmen: Oops, my bad, it's (sort of?) a copy constructor.  (I say "sort of" because it takes the right-hand side by nonconst reference, yet it looks like the intent is not to take ownership from the right-hand side.)

src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
brian-kelley: @mhoemmen @srajama1 I wasn't planning on using a ``#pragma unroll`` or manual unrolling, but do you think they would help? SPMV does some extremely aggressive unrolling on CPU with fully unrolled batches of 16/17 (depending on n mod 16), and then 15 different unrolled versions in a Duff's device to deal with the remainder.

Here's what I have now - this is the RangePolicy version of apply, processing row ``ii``. apply_batch_size = 8 is constexpr. Could easily make that 16 instead.

```
          scalar_t sum[apply_batch_size];
          nnz_lno_t num_vecs = _Xvector.extent(1);
          for(nnz_lno_t batch_start = 0; batch_start < num_vecs; batch_start += apply_batch_size)
          {
            nnz_lno_t this_batch_size = apply_batch_size;
            if(batch_start + this_batch_size >= num_vecs)
              this_batch_size = num_vecs - batch_start;
            //the current batch of columns given by: batch_start, this_batch_size
            for(nnz_lno_t i = 0; i < this_batch_size; i++)
              sum[i] = _Yvector(ii, batch_start + i);
            for (size_type adjind = row_begin; adjind < row_end; ++adjind){
              nnz_lno_t colIndex = _adj(adjind);
              nnz_scalar_t val = _adj_vals(adjind);
              for(nnz_lno_t i = 0; i < this_batch_size; i++)
                sum[i] -= val * _Xvector(colIndex, batch_start + i);
            }
            nnz_scalar_t invDiagonalVal = _permuted_inverse_diagonal(ii);
            for(nnz_lno_t i = 0; i < this_batch_size; i++)
              _Xvector(ii, batch_start + i) += omega * sum[i] * invDiagonalVal;
```

src/common/KokkosKernels_Handle.hpp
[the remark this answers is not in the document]
brian-kelley: @mhoemmen Oh I see. I think all the Handle getters are const (or, they should be) so I'll try to pass ``right_side_handle`` by const&.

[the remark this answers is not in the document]
brian-kelley: @mhoemmen I wasn't able to use const&, since the subhandle getters aren't const.

src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
mhoemmen: @brian-kelley I don't think you necessarily have to do that, and it could make sense for the constants to differ.  @iyamazaki reported the SGS issue [here](https://github.com/kokkos/kokkos-kernels/issues/488); I'll call him up and see if he had a particular number of columns in mind.

mhoemmen: @brian-kelley Hm, he doesn't have a phone number listed in the directory yet.  Might he be in CSRI today?

src/common/KokkosKernels_Handle.hpp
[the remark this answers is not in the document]
mhoemmen: @brian-kelley Thanks for trying!  It's OK to leave that broken for later :-)

src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
srajama1: @brian-kelley : The sizes we care are around 3, so this is pretty good. Whenever you are ready, we can push this in.

iyamazaki: Thank you! This is great!! I'll give it a try from ifpack2 and let you know.

brian-kelley:
@iyamazaki @mhoemmen @srajama1 The way it works now, the data accesses are optimal for <= 8 XY columns. The worst case (9 columns) is that it gets accessed once per 4.5 columns, which is still way better than what we had.

Updated spot checks:
Bowman:
PASSED TESTS
intel-16.4.258-Pthread-release build_time=1640 run_time=3112
intel-16.4.258-Pthread_Serial-release build_time=2786 run_time=5909
intel-16.4.258-Serial-release build_time=1583 run_time=2840
intel-17.2.174-OpenMP-release build_time=2104 run_time=1089
intel-17.2.174-OpenMP_Serial-release build_time=2954 run_time=4356
intel-17.2.174-Pthread-release build_time=1415 run_time=2845
intel-17.2.174-Pthread_Serial-release build_time=2599 run_time=5549
intel-17.2.174-Serial-release build_time=1425 run_time=2790
intel-18.2.199-OpenMP-release build_time=1696 run_time=1115
intel-18.2.199-OpenMP_Serial-release build_time=3119 run_time=3876
intel-18.2.199-Pthread-release build_time=1335 run_time=3177
intel-18.2.199-Pthread_Serial-release build_time=2717 run_time=5654
intel-18.2.199-Serial-release build_time=1288 run_time=2843
FAILED TESTS

White:
PASSED TESTS
cuda-10.0.130-Cuda_Serial-release build_time=1243 run_time=1487
cuda-9.2.88-Cuda_OpenMP-release build_time=1295 run_time=1111
gcc-6.4.0-OpenMP_Serial-release build_time=603 run_time=1274
gcc-7.2.0-OpenMP-release build_time=415 run_time=483
gcc-7.2.0-OpenMP_Serial-release build_time=613 run_time=1570
gcc-7.2.0-Serial-release build_time=280 run_time=659
gcc-7.4.0-OpenMP-release build_time=390 run_time=497
ibm-16.1.0-Serial-release build_time=1425 run_time=1003
FAILED TESTS

srajama1:
@brian-kelley : Feel free to merge when ready .. we need to add this to the release notes.

brian-kelley:
@srajama1 It's ready, you can merge it right now. I just emailed Christian about getting merge permission so I can do it myself in the future.

mhoemmen:
@brian-kelley There's always a trade-off between build times and sizes, and optimizations like this.  I'm pretty OK with optimizing for <= 8 columns, as long as 9 isn't too slow.

Issue #504, opened 2019-11-27
statement is unreachable warning messages in gs handle with cuda


ndellingwood:
These warning messages are occurring in cuda builds detected during testing of PR #491 and #503 :

```
/ascldap/users/ndellin/kokkos-kernels-cmake/src/sparse/KokkosSparse_gauss_seidel_handle.hpp(592): warning: statement is unreachable
          detected during:
            instantiation of "__nv_bool KokkosSparse::ClusterGaussSeidelHandle<size_type_, lno_t_, scalar_t_, ExecutionSpace, TemporaryMemorySpace, PersistentMemorySpace>::use_teams() const [with size_type_=const size_t, lno_t_=const int, scalar_t_=const double, ExecutionSpace=Kokkos::Cuda::execution_space, TemporaryMemorySpace=Kokkos::CudaSpace::memory_space, PersistentMemorySpace=Kokkos::CudaSpace::memory_space]" 
/ascldap/users/ndellin/kokkos-kernels-cmake/src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp(980): here
            instantiation of "void KokkosSparse::Impl::ClusterGaussSeidel<HandleType, lno_row_view_t_, lno_nnz_view_t_, scalar_nnz_view_t_>::initialize_numeric() [with HandleType=KokkosKernels::Experimental::KokkosKernelsHandle<const size_t, const int, const double, Kokkos::Cuda, Kokkos::CudaSpace, Kokkos::CudaSpace>, lno_row_view_t_=Kokkos::View<const size_t *, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::CudaSpace::execution_space, Kokkos::CudaSpace::memory_space>, Kokkos::MemoryTraits<1U>>, lno_nnz_view_t_=Kokkos::View<const int *, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::CudaSpace::execution_space, Kokkos::CudaSpace::memory_space>, Kokkos::MemoryTraits<1U>>, scalar_nnz_view_t_=Kokkos::View<const double *, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::CudaSpace::execution_space, Kokkos::CudaSpace::memory_space>, Kokkos::MemoryTraits<1U>>]" 
/ascldap/users/ndellin/kokkos-kernels-cmake/src/sparse/impl/KokkosSparse_gauss_seidel_spec.hpp(278): here
[...]

PR #505, opened 2019-11-27
KokkosSparse_gauss_seidel_handle.hpp: fix unreachable code warning

ndellingwood:
Issue #504

### DIFF
--- a/src/sparse/KokkosSparse_gauss_seidel_handle.hpp
+++ b/src/sparse/KokkosSparse_gauss_seidel_handle.hpp
@@ -564,32 +564,33 @@ namespace KokkosSparse{
     
     bool use_teams() const
     {
+      bool return_value = false;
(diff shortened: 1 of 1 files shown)


src/sparse/KokkosSparse_gauss_seidel_handle.hpp
```diff
@@ -564,32 +564,33 @@ namespace KokkosSparse{
     
     bool use_teams() const
     {
+      bool return_value = false;
 #if defined( KOKKOS_ENABLE_SERIAL )
       if (Kokkos::Impl::is_same< Kokkos::Serial , ExecutionSpace >::value) {
```
mhoemmen: We should at some point replace `Kokkos::Impl::is_same` with `std::is_same` (which is in the `<type_traits>` header).

Issue #506, opened 2019-12-03
SpMV unit-test tolerance is incorrectly set


lucbv:
In `Test_Sparse_spmv.hpp` the following pattern is found in most tests:
```
  typedef typename crsMat_t::values_type::non_const_type scalar_view_t;
  typedef typename scalar_view_t::value_type ScalarA;
  double eps = std::is_same<ScalarA,float>::value ? 2*1e-3 : 1e-7;
```
which is wrong when testing with `Kokkos::Komplex<float>`

lucbv:
This has now been fixed.

PR #503, opened 2019-11-26
spgemm: shmem alignment updates for numeric routines

ndellingwood:
Part of issue #500

PR #507, opened 2019-12-03
SpMV tests: fixing the logic to detect the level of precision needed

lucbv:
This work should fix the problem described in issue #506

The new logic introduce for the spmv_struct test should lead to more consistent
checks and hopefully make the test path on GPU platforms.

### DIFF
--- a/unit_test/sparse/Test_Sparse_spmv.hpp
+++ b/unit_test/sparse/Test_Sparse_spmv.hpp
@@ -15,16 +15,18 @@
 
 namespace Test {
 
-template < class VectorType0, class VectorType1, class AT_Type >
+template < class VectorType0, class VectorType1 >
 struct fSPMV {
-  typedef int value_type;
-  typedef Kokkos::Details::ArithTraits<typename AT_Type::non_const_value_type> AT;
+  using value_type = int;
+  using AT         = Kokkos::ArithTraits<typename VectorType1::non_const_value_type>;
+  using ATM        = Kokkos::ArithTraits<typename AT::mag_type>;
+  using mag_type   = typename AT::mag_type;
 
   VectorType0 expected_y;
   VectorType1 y;
-  double eps;
+  mag_type    eps;
 
-  fSPMV(const VectorType0 & _ex_y, const VectorType1 & _y, const double _eps)
+  fSPMV(const VectorType0 & _ex_y, const VectorType1 & _y, const mag_type _eps)
   : expected_y(_ex_y)
   , y(_y)
   , eps(_eps)
@@ -32,7 +34,11 @@ struct fSPMV {
 
   KOKKOS_INLINE_FUNCTION
   void operator()( const int i, value_type& err ) const {
-    if(AT::abs(expected_y(i)-y(i))>eps) {
+
+    const mag_type error = AT::abs(expected_y(i)-y(i))
+      / (AT::abs(expected_y(i)) > ATM::zero() ? AT::abs(expected_y(i)) : ATM::one());
+
+    if(error > eps) {
       err++;
       printf("expected_y(%d)=%f, y(%d)=%f\n", i, AT::abs(expected_y(i)), i, AT::abs(y(i)));
     }
@@ -42,16 +48,17 @@ struct fSPMV {
 
 template <typename crsMat_t, typename x_vector_type, typename y_vector_type>
 void sequential_spmv(crsMat_t input_mat, x_vector_type x, y_vector_type y,
-    typename y_vector_type::non_const_value_type alpha, typename y_vector_type::non_const_value_type beta){
+                     typename y_vector_type::non_const_value_type alpha,
+                     typename y_vector_type::non_const_value_type beta){
 
-  typedef typename crsMat_t::StaticCrsGraphType graph_t;
-  typedef typename graph_t::row_map_type size_type_view_t;
-  typedef typename graph_t::entries_type lno_view_t;
-  typedef typename crsMat_t::values_type::non_const_type scalar_view_t;
+  using graph_t          = typename crsMat_t::StaticCrsGraphType;
+  using size_type_view_t = typename graph_t::row_map_type;
+  using lno_view_t       = typename graph_t::entries_type;
+  using scalar_view_t    = typename crsMat_t::values_type::non_const_type;
 
-  typedef typename size_type_view_t::non_const_value_type size_type;
-  typedef typename lno_view_t::non_const_value_type lno_t;
-  typedef typename scalar_view_t::non_const_value_type scalar_t;
+  using size_type = typename size_type_view_t::non_const_value_type;
+  using lno_t     = typename lno_view_t::non_const_value_type;
+  using scalar_t  = typename scalar_view_t::non_const_value_type;
 
 
   typename scalar_view_t::HostMirror h_values = Kokkos::create_mirror_view(input_mat.values);
@@ -73,7 +80,6 @@ void sequential_spmv(crsMat_t input_mat, x_vector_type x, y_vector_type y,
   KokkosKernels::Impl::safe_device_to_host_deep_copy (y.extent(0), y, h_y);
   Kokkos::fence();
 
-
   lno_t nr = input_mat.numRows();
 
   for (lno_t i = 0; i < nr; ++i){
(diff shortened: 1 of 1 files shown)


unit_test/sparse/Test_Sparse_spmv.hpp
```diff
@@ -93,15 +93,24 @@ void sequential_spmv(crsMat_t input_mat, x_vector_type x, y_vector_type y,
...
+  // so let us use y_value_type to determine
+  // the appropriate tolerance precision.
+  using y_value_type  = typename y_vector_type::non_const_value_type;
+  using y_value_trait = Kokkos::ArithTraits<y_value_type>;
+  // Note on 2019-12-03 by LBV:
+  // one really want the following definition,
+  // const y_value_trait::mag_type eps = f(y_value_trait::eps());
+  // where f() introduces a fudge factor that accounts for
+  // rounding error and such...
+  const double eps = std::is_same<y_value_trait::mag_type, float>::value ? 2*1e-3 : 1e-7;
```
mhoemmen: Wow, both of those tolerances are pretty loose.  I know this was there before, but should it really be this loose?

ndellingwood: `typename` is needed before `std::is_same` to remove compilation error

unit_test/sparse/Test_Sparse_spmv.hpp
```diff
@@ -208,11 +233,19 @@ void check_spmv_mv(crsMat_t input_mat, x_vector_type x, y_vector_type y, y_vecto
...
+    // so let us use y_value_type to determine
+    // the appropriate tolerance precision.
+    using y_value_type  = typename y_vector_type::non_const_value_type;
+    using y_value_trait = Kokkos::ArithTraits<y_value_type>;
+    // Note on 2019-12-03 by LBV:
+    // one really want the following definition,
+    // const y_value_trait::mag_type eps = f(y_value_trait::eps());
+    // where f() introduces a fudge factor that accounts for
+    // rounding error and such...
+    const double eps = std::is_same<y_value_trait::mag_type, float>::value ? 2*1e-3 : 1e-7;
```
ndellingwood: `typename` is needed in `std::is_same` to remove compilation error - `typename y_value_trait::mag_type`

unit_test/sparse/Test_Sparse_spmv.hpp
```diff
@@ -110,27 +119,35 @@ void check_spmv(crsMat_t input_mat, x_vector_type x, y_vector_type y,
...
+  // so let us use y_value_type to determine
+  // the appropriate tolerance precision.
+  using y_value_type  = typename y_vector_type::non_const_value_type;
+  using y_value_trait = Kokkos::ArithTraits<y_value_type>;
+  // Note on 2019-12-03 by LBV:
+  // one really want the following definition,
+  // const y_value_trait::mag_type eps = f(y_value_trait::eps());
+  // where f() introduces a fudge factor that accounts for
+  // rounding error and such...
+  const double eps = std::is_same<y_value_trait::mag_type, float>::value ? 2*1e-3 : 1e-7;
```
ndellingwood: `typename` is needed in `std::is_same` to remove compilation error - `typename y_value_trait::mag_type`

ndellingwood:
@lucbv with the changes I marked in the code (missing `typename`s) this builds and the failing tests I encountered with complex_float are now passing:

```
[ndellin@kokkos-dev-2 unit_test]$ ./KokkosKernels_sparse_cuda --gtest_filter=*spmv_struct*
Note: Google Test filter = *spmv_struct*
[==========] Running 8 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 8 tests from cuda
[ RUN      ] cuda.sparse_spmv_struct_double_int_int_TestExecSpace
[       OK ] cuda.sparse_spmv_struct_double_int_int_TestExecSpace (197 ms)
[ RUN      ] cuda.sparse_spmv_struct_double_int_size_t_TestExecSpace
[       OK ] cuda.sparse_spmv_struct_double_int_size_t_TestExecSpace (192 ms)
[ RUN      ] cuda.sparse_spmv_struct_float_int_int_TestExecSpace
[       OK ] cuda.sparse_spmv_struct_float_int_int_TestExecSpace (129 ms)
[ RUN      ] cuda.sparse_spmv_struct_float_int_size_t_TestExecSpace
[       OK ] cuda.sparse_spmv_struct_float_int_size_t_TestExecSpace (134 ms)
[ RUN      ] cuda.sparse_spmv_struct_kokkos_complex_double_int_int_TestExecSpace
[       OK ] cuda.sparse_spmv_struct_kokkos_complex_double_int_int_TestExecSpace (319 ms)
[ RUN      ] cuda.sparse_spmv_struct_kokkos_complex_double_int_size_t_TestExecSpace
[       OK ] cuda.sparse_spmv_struct_kokkos_complex_double_int_size_t_TestExecSpace (306 ms)
[ RUN      ] cuda.sparse_spmv_struct_kokkos_complex_float_int_int_TestExecSpace
[       OK ] cuda.sparse_spmv_struct_kokkos_complex_float_int_int_TestExecSpace (208 ms)
[ RUN      ] cuda.sparse_spmv_struct_kokkos_complex_float_int_size_t_TestExecSpace
[       OK ] cuda.sparse_spmv_struct_kokkos_complex_float_int_size_t_TestExecSpace (253 ms)
[----------] 8 tests from cuda (1738 ms total)

[----------] Global test environment tear-down
[==========] 8 tests from 1 test case ran. (1738 ms total)
[  PASSED  ] 8 tests.
```

This was tested on kokkos-dev-2 with a branch built from the cmake-rebased branch with conflicts resolved, fixes from PR #503 and this PR, tested against Kokkos develop branch.
Build env configured with:

```
module load sems-cmake/3.12.2 sems-cuda/9.2 sems-gcc/7.2.0

../../cm_generate_makefile.bash --kokkoskernels-path=${PWD}/../../ --kokkos-path=<KOKKOSPATH> --with-serial --with-cuda --with-scalars=double,complex_double,float,complex_float --with-ordinals=int,long --with-offsets=int --with-layouts=layoutleft --arch=SNB,Volta70 --compiler=<KOKKOSPATH>/bin/nvcc_wrapper --with-tpls=cublas --kokkos-make-j=12
```

lucbv:
@ndellingwood you forgot to comment on line 202 ; )
I added the necessary `typename`.

unit_test/sparse/Test_Sparse_spmv.hpp
lucbv: agreed, as you can see in my comment I am not fond of what is currently implemented.
I feel something like `100*y_value_trait::eps()` should be working, @ndellingwood any chance you could try with that?

ndellingwood: @lucbv we tested with 
`1000*y_value_trait::eps()` and spmv_struct tests passed, seeing lots of failures with spmv after these changes:

```
[  FAILED  ] cuda.sparse_spmv_double_int_int_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_double_int_size_t_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_float_int_int_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_float_int_size_t_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_kokkos_complex_double_int_int_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_kokkos_complex_double_int_size_t_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_kokkos_complex_float_int_int_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_kokkos_complex_float_int_size_t_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_double_int_int_LayoutLeft_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_double_int_size_t_LayoutLeft_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_float_int_int_LayoutLeft_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_float_int_size_t_LayoutLeft_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_kokkos_complex_double_int_int_LayoutLeft_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_kokkos_complex_double_int_size_t_LayoutLeft_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_kokkos_complex_float_int_int_LayoutLeft_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_kokkos_complex_float_int_size_t_LayoutLeft_TestExecSpace
```

mhoemmen: Are these large problems, with a lot of rows or entries per row?

lucbv: I am actually considering doing something like this:
```
const double eps = max_nnz_per_row*y_vector_trait::eps()
```
This would mean that we tolerate an accumulation of round-off error which is equivalent to the number of additions happening to compute a single entry in `y`, does that make sense?

mhoemmen: @lucbv makes sense for infinity-norm error, which I think is the right metric here.

lucbv:
@mhoemmen @srajama1 @ndellingwood 
Here is what I have been able to diagnose using `tolerance=10*max_nnz_per_row*Kokkos::ArithTraits<value_type>::eps()`
1. the issue is arising when `vector_length > 1` which made me suspect that the order of operations in the row wise reduction is at the root of this problem.
2. when using only positive values in the CrsMatrix the issue also vanishes which makes me think that the different order of operation triggers a catastrophic arithmetic cancellation.

We can't really go around the issue, the order of operations will never be the same in the row-wise reduction unless we run the tests with `vector_length==1` which means that we do not test algorithm the way it will be used in production.
Making all the entries in the matrices positive is possible but I am not sure that it is the right path either. If you have any suggestions on how to handle this problem please let me know and we will make the appropriate adjustment otherwise I suggest we keep the currently used tolerances.

mhoemmen:
@lucbv I'd rather that we tested the algorithm as it's run in production.  Perhaps we need to think harder about the error bound.

ndellingwood:
@lucbv are changes still needed for this PR based on your diagnosis?

lucbv:
@ndellingwood yes I need to make a couple changes, will be ready in an hour or so...

ndellingwood:
@lucbv another tweak to the PR looks to be necessary, the following failure occurred with intel-18.0.5-OpenMP during spot-check on kokkos-dev-2:

```
KokkosKernels::UnitTests::spmv_struct: 174 errors of 64000 with params: 2 1.000000 0.000000
/ascldap/users/ndellin/kokkos-kernels-cmake/unit_test/sparse/Test_Sparse_spmv.hpp:214: Failure
Value of: num_errors==0
  Actual: false 
Expected: true
```

lucbv:
@ndellingwood I am doing a local intel/18.0.5 build, hopefully I can reproduce your errors...

ndellingwood:
Here's spot-check on kokkos-dev-2, the errors were due to a -Werror unused variable that I fixed in a different recent PR.

```
Going to test compilers:  gcc/7.3.0 gcc/9.1 intel/18.0.5 clang/8.0 cuda/10.1
...
PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=568 run_time=523
clang-8.0-Pthread_Serial-release build_time=355 run_time=345
cuda-10.1-Cuda_OpenMP-release build_time=632 run_time=484
gcc-7.3.0-Pthread-release build_time=207 run_time=195
gcc-9.1-Serial-release build_time=214 run_time=186
intel-18.0.5-OpenMP-release build_time=373 run_time=172
FAILED TESTS
gcc-7.3.0-OpenMP-release (build failed)
gcc-9.1-OpenMP-release (build failed)
```

Will merge this is but leave the issue open.

Issue #509, opened 2019-12-04
Dot-based GEMM Cuda specialization - update for beta == 0 case


ndellingwood:
The Dot-based GEMM Cuda specialization needs updating for compatibility with Trilinos to handle the case of `beta==0` for matrices `C` with `NaN` entries.

ndellingwood:
Related PRs #510 #524

PR #510, opened 2019-12-05
Dot based gemm: Add different beta cases and allow fewer than 32 teams 

seheracer:
This PR fixes the failures in Trilinos tests:
https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-6226-test&field2=buildstarttime&compare2=84&value2=NOW

1) NaN failures are fixed by having a different case for beta=0 that actually initializes the entries of C to 0.

2) Diff failures (that happen to only occur on very short vectors -- 50 or 72 entries-- with very tight epsilon) are fixed by allowing fewer than 32 team -- 1 team in those cases. This eliminated the nondeterministic access order of the teams to C array to update the results.

### DIFF
--- a/src/impl/tpls/KokkosBlas3_gemm_tpl_spec_decl.hpp
+++ b/src/impl/tpls/KokkosBlas3_gemm_tpl_spec_decl.hpp
@@ -343,8 +343,9 @@ namespace Impl {
 // performed on very long vectors, so, each dot product is distributed among
 // numDivPerDot teams.     
 
-struct TagInit{};   // The tag for the initialization parallel_for 
-struct TagMult{};   // The tag for the multiplication parallel_for 
+struct TagZero{};   // The init tag for beta=0 
+struct TagInit{};   // The init tag for beta!=0 and beta !=1 
+struct TagMult{};   // The multiplication tag 
 template<class ExecSpace, class AV, class BV, class CV>
 struct DotBasedGEMM{
 
@@ -356,12 +357,12 @@ struct DotBasedGEMM{
   using size_A = typename AV::size_type;
   using scalar_C = typename CV::non_const_value_type;
   using size_C = typename CV::size_type;
+  using CVT = Kokkos::Details::ArithTraits<scalar_C>;
 
   const scalar_A alpha;
   const scalar_C beta;
 
   // The following types (especially dotSize) could have simply been int,
-  // e.g., see lines 488-490.
   const size_C numCrows;           
   const size_C numCcols;
 
@@ -382,8 +383,8 @@ struct DotBasedGEMM{
     size_C appxNumTeams = (dotSize * ndots) / workPerTeam; // Estimation for appxNumTeams
 
     // Adjust appxNumTeams in case it is too small or too large
-    if(appxNumTeams < 32)
-      appxNumTeams = 32;
+    if(appxNumTeams < 1)
+      appxNumTeams = 1;
     if(appxNumTeams > 1024)
       appxNumTeams = 1024;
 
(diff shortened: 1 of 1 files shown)


ndellingwood:
Cross-referencing #509

src/impl/tpls/KokkosBlas3_gemm_tpl_spec_decl.hpp
```diff
@@ -409,20 +410,29 @@ struct DotBasedGEMM{
...
+    
     // Multiply alpha*A^TB and add it to beta*C
     Kokkos::TeamPolicy<TagMult, ExecSpace> policy2(numTeams, Kokkos::AUTO);
     Kokkos::parallel_for("Perform Dot Product Based GEMM", policy2, *this);
 
   }
 
+  KOKKOS_INLINE_FUNCTION
+  void operator() (const TagZero&, const size_C &i) const {
+    const size_C rowId = i / numCcols;
```
mhoemmen: Kokkos has MDRangePolicy for multidimensional index ranges.  It looks like that's what you want to do here.

src/impl/tpls/KokkosBlas3_gemm_tpl_spec_decl.hpp
```diff
@@ -409,20 +410,29 @@ struct DotBasedGEMM{
...
+    const size_C colId = i % numCcols;
+    C(rowId, colId) = CVT::zero();
```
mhoemmen: You shouldn't need to write a separate kernel for this; you can do `Kokkos::deep_copy(C, Scalar{})` and it will fill C with zeros.

src/impl/tpls/KokkosBlas3_gemm_tpl_spec_decl.hpp
ndellingwood: If you try MDRangePolicy here make sure to performance test, the matrix shapes (tall skinny) may result in poor performance without tuning tile sizes.

seheracer: @mhoemmen @ndellingwood thanks for useful comments! I will try it.

src/impl/tpls/KokkosBlas3_gemm_tpl_spec_decl.hpp
seheracer: Thanks! I will replace this kernel with deep_copy.

ndellingwood: @seheracer I'd suggest tracking performance changes when testing this change as well, internally the `deep_copy` will use an `MDRangePolicy` for this 2D View but there could be issues again with default tile dims not being well-suited for the shapes of matrices targeted here.

src/impl/tpls/KokkosBlas3_gemm_tpl_spec_decl.hpp
seheracer: MDRangePolicy results in reduction in **total** runtime of GEMM by **5.95%**, **4.82%**, **0.82%**, and **-1.61%** for 4x4, 8x8, 16x16, and 32x32 C matrices, respectively (averaged over seven different values for the number of rows in A and B in range [10 to 10M]).

src/impl/tpls/KokkosBlas3_gemm_tpl_spec_decl.hpp
seheracer: Compared to MDRangePolicy, deep_copy **increased** the total runtime of GEMM by **29.62%**, **22.43%**, **20.72%**, and **14.07%** for 4x4, 8x8, 16x16, and 32x32 C matrices, respectively (averaged over seven different values for the number of rows in A and B in range [10 to 10M]).

ndellingwood: Ouch! Thanks for testing.

ndellingwood:
@seheracer is this ready for merge as is, or are you adding any additional `MDRangePolicy` changes? Since you worked on the current PR to confirm it resolved the issues in Trilinos I'm inclined to merge this as is since any extra work will require extra Trilinos testing, does that sound okay or would you prefer I hold off?

seheracer:
@ndellingwood merging this as is sounds okay, thanks!

ndellingwood:
Merging, thanks @seheracer !

PR #518, opened 2019-12-11
block-sgs unit test: Fix case for rank-1 view accepting 2 args

ndellingwood:
This resolves unit test failures like the following:

*sparse_block_gauss_seidel_rank1_double_int_int_TestExecSpace
Constructor for Kokkos View 'Y VECTOR' has mismatched number of arguments.
Number of arguments = 2 but dynamic rank = 1


ndellingwood:
@brian-kelley this came up when disabling kokkos' deprecated code during testing.

ndellingwood:
Thanks @brian-kelley !

Issue #517, opened 2019-12-11
CMake build system does not work properly on CUDA platforms


lucbv:
@ndellingwood 
I am just using this issue to track a couple of changes that need to happen in the CMake build system of Kokkos-Kernels for my build to work on white:

1. the tests of Kokkos are enabled by default which make building kokkos-kernels much slower
   proposed fix: add a variable called `KOKKOS_DO_TESTS` and set it to off by default, also add logic to allow users to turn it on explicitly from the command line
```
 KOKKOS_DO_EXAMPLES=OFF
 KOKKOS_DO_TESTS=OFF
 KOKKOS_MAKEINSTALL_J=4

    --no-tests)
      KOKKOS_DO_TESTS=OFF
      ;;

echo cmake ... -DKokkos_ENABLE_TESTS=${KOKKOS_DO_TESTS} ...
cmake ... -DKokkos_ENABLE_TESTS=${KOKKOS_DO_TESTS}
```
Also the above code follows the logic used to turn on/off the compilation of tests in kokkos, but there seems to be no way to turn on the examples explicitly (at all?) since the default value if `OFF` and the command line option only allows you to turn it off...

2. there are typos in `CMakeList.txt` on line 135 and 138 (extra "}" after variable names), it currently reads:
```
  IF(KOKKOSKERNELS_INST_EXECSPACE_CUDA} AND KOKKOSKERNELS_INST_MEMSPACE_CUDASPACE})
    LIST(APPEND DEVICE_LIST "<Cuda,CudaSpace>")
  ENDIF()
  IF(KOKKOSKERNELS_INST_EXECSPACE_CUDA} AND KOKKOSKERNELS_INST_MEMSPACE_CUDAUVMSPACE})
    LIST(APPEND DEVICE_LIST "<Cuda,CudaUVMSpace>")
ENDIF()
```
but should be
```
  IF(KOKKOSKERNELS_INST_EXECSPACE_CUDA AND KOKKOSKERNELS_INST_MEMSPACE_CUDASPACE)
    LIST(APPEND DEVICE_LIST "<Cuda,CudaSpace>")
  ENDIF()
  IF(KOKKOSKERNELS_INST_EXECSPACE_CUDA AND KOKKOSKERNELS_INST_MEMSPACE_CUDAUVMSPACE)
    LIST(APPEND DEVICE_LIST "<Cuda,CudaUVMSpace>")
ENDIF()
```

ndellingwood:
[...]

PR #519, opened 2019-12-11
CMakeLists, cm_generate_makefile: Remove typo, add/fix options

ndellingwood:
Part of issue #517


ndellingwood:
@lucbv a couple comments - 
If you add `-O3` to the cxxflags you'll see it appear twice on compile lines in RELEASE builds (which is set as default without passing --debug; you'll also get warnings from nvcc_wrapper); Jeremy said this is expected, CMake doesn't do a good job of deduplicating flags.

Also, warnings we saw like `Could NOT find Kokkos_OPTIONS (missing: CUDA_UVM)` are harmless, later Jeremy may add a QUIET option to silence things like this.

jjwilke:
@lucbv @ndellingwood Sorry, should have clarified. If you want to exactly control optimization flags, then don't set a build type. If no `CMAKE_BUILD_TYPE` is set, then CMake won't add any optimization or debug flags.

ndellingwood:
I'll update the script so that `CMAKE_BUILD_TYPE` isn't set by default to allow optimization flags to be controlled.

PR #520, opened 2019-12-12
cm_generate_makefile: more improvements

ndellingwood:
Add flag to make BUILD_TYPE optional; add options to set memspaces

### DIFF
--- a/cm_generate_makefile.bash
+++ b/cm_generate_makefile.bash
@@ -121,6 +121,17 @@ get_kernels_layouts_list() {
   done
 }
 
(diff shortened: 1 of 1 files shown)


ndellingwood:
@lucbv a couple follow up items from PR #519 for the cm_generate_makefile script.

ndellingwood:
@lucbv also, if you add `-O3` in the cxxflags you shouldn't see annoying warnings, unless you also specify `--release` in which case `-O3` will be duplicated on the compile lines as that will explicitly set the build type to release (which is what was being done by default previously).

PR #521, opened 2019-12-12
Test_Sparse_gauss_seidel: UVM Fix

ndellingwood:
Use auto instead of explicity type when calling create_mirror_view
This fixes compilation errors of type
"Incompatible View copy construction" when UVM is enabled


ndellingwood:
Detected during Cuda builds of Trilinos integration testing which enable UVM.
This PR fixes a compilation error detected during those builds.

PR #522, opened 2019-12-12
KokkosKernelsConfig.cmake fixes for TPLs

jjwilke:
* Guard `add_target` with ifs to make sure not added twice 
* Misc fixes for the append config line function


ndellingwood:
@jjwilke what issues did you run into to require this change and in what types of builds?

jjwilke:
Related to https://github.com/kokkos/kokkos/issues/2587.  After fixing this for Kokkos, realized KokkosKernels would have the same problem. 

Testing on CUDA builds, I caught the other issue.

jjwilke:
If you have multiple calls to `find_package(KokkosKernels)` or `find_dependency(KokkosKernels)` CMake would load `KokkosKernelsConfig.cmake` twice and calling `add_library` twice for the same library.

This was never actually an issue for any KokkosKernels projects, but would have been.

ndellingwood:
@jjwilke thanks for the explanation, shall I merge into develop or will that break your current work-flow?

jjwilke:
This should merge into develop ASAP.

ndellingwood:
@jjwilke merged!

Issue #539, opened 2019-12-16
gemv transpose mode issues with cublas and M == 0


ndellingwood:
Recent updates with added testing for `gemv` (see #514) added a testing condition that appears to be unsupported by cublas in Transpose mode when `M == 0` [(see comments in the documentation)](https://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-gemv).

As a first pass I'm going to guard the test to prevent the failure with TPLs not supporting this case.

The kokkos-kernels fall-back implementation supports this (returning `beta*y` if `beta != 0; returns `y` if `beta == 0`). If this is generally expected behavior of users we'll need to "short-circuit" the TPL macros to handle this case.

Adding @iyamazaki  @srajama1 @mhoemmen for feedback on what expected behavior for this case should be, so TPL macros can be updated if necessary.

ndellingwood:
Detected testing with the following:

Machine - kokkos-dev-2

kokkos develop branch SHA kokkos/kokkos@0378c0fa18fd366a6fe5e94b04b11e8e46836df3

kokkos-kernels develop branch SHA 5d5d07ccc0b16f097542f714870d1dd50451c571

Configuration:
```
module load sems-cuda/10.1 sems-gcc/7.3.0 sems-cmake/3.12.2

../../cm_generate_makefile.bash --kokkos-path=${HOME}/kokkos --compiler=${HOME}/kokkos/bin/nvcc_wrapper --with-cuda --arch=Volta70 --with-tpls=cublas --kokkos-make-j=12
```

ndellingwood:
PR disabling the M==0 test with TPLs in #540

mhoemmen:
@ndellingwood wrote:
> added a testing condition that appears to be unsupported by cublas in Transpose mode when `M == 0`....

Do you mean the condition on `x`, that it have at least `(1+(m-1)*abs(incx))` elements if `transa != CUBLAS_OP_N`?  That should only be a problem if `incx != 1`.  cuBLAS seems to have inherited this condition from [the Reference BLAS](http://www.netlib.org/lapack/explore-html/d7/d15/group__double__blas__level2_gadd421a107a488d524859b4a64c1901a9.html).

CUDA needs to be able to handle this case.  I would rather that we fix kokkos-kernels, not that we hide this bug by disabling the test.

ndellingwood:
[...]

PR #540, opened 2019-12-16
gemv test: disable M == 0, mode='T' case with TPLs

ndellingwood:
Part of issue #539


ndellingwood:
Also adding @iyamazaki

mhoemmen:
See [my comment here](https://github.com/kokkos/kokkos-kernels/issues/539#issuecomment-565919015).  This looks like a kokkos-kernels bug that we could easily fix with an `if` test.

mhoemmen:
@ndellingwood and I discussed this on the phone today.  Nathan agrees that it's better to fix the bug first.

ndellingwood:
Closing in favor of #551

PR #552, opened 2019-12-19
Sptrsv supernode

iyamazaki:
This PR adds the options for supernodal-based sparse-triangular solve (with SuperLU or Cholmod) to the existing level-based scheduling sparse-triangular solve.


ndellingwood:
This is ready for review, some extra CMake TPL support may be needed.

ndellingwood:
Running spot-check on kokkos-dev-2

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+  scalar_t ONE = scalar_t (1);
+
+  Kokkos::Timer timer;
+  const int nrows = Mtx.graph.numRows ();
+
+  // ==============================================
+  // > create a handle
+  cusparseStatus_t status;
+  cusparseHandle_t handle = 0; 
+  status = cusparseCreate (&handle);
```
mhoemmen: Doesn't kokkos-kernels already have a wrapper for cuSPARSE handles?  It would make sense to work with that existing framework.

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+
+  // > create the rhs ** on host **
+  // A*sol generates rhs: rhs is dense, use spmv
+  host_scalar_view_t rhs_host ("rhs_host", nrows);
+  KokkosSparse::spmv ("N", ONE, Mtx, sol_host, ZERO, rhs_host);
+
+  // ==============================================
+  // step 1: apply forward-pivot to rhs on the host
+  host_scalar_view_t tmp_host ("temp", nrows);
+  forwardP_supernode<scalar_t> (nrows, perm_r, 1, rhs_host.data(), nrows, tmp_host.data(), nrows);
```
mhoemmen: Why not make permutation a device kernel?  Then you could run entirely with device buffers.  In general, all of this setup looks like it wouldn't be much trouble to get working on device.

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+  cusparseSetMatDiagType (descrL, CUSPARSE_DIAG_TYPE_NON_UNIT);
+  cusparseSetMatType (descrL,CUSPARSE_MATRIX_TYPE_GENERAL);
+
+  // ==============================================
+  // step 2: query how much memory used in csrsv2, and allocate the buffer
+  // pBuffer returned by cudaMalloc is automatically aligned to 128 bytes.
+  int pBufferSize;
+  void *pBufferL = 0;
+  cusparseOperation_t transL = (col_majorL ? CUSPARSE_OPERATION_TRANSPOSE : CUSPARSE_OPERATION_NON_TRANSPOSE);
+  if (std::is_same<scalar_t, double>::value == true) {
```
mhoemmen: `== true` is superfluous here, since the type of `std::is_same<U, V>::value` is `bool`.

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+  forwardP_supernode<scalar_t> (nrows, perm_r, 1, rhs_host.data(), nrows, tmp_host.data(), nrows);
+
+  // copy rhs to the default host/device
+  scalar_view_t rhs ("rhs", nrows);
+  scalar_view_t sol ("sol", nrows);
+  Kokkos::deep_copy (rhs, tmp_host);
+
+  // ==============================================
+  // step 2: solve L*y = x
+  Kokkos::fence ();
```
mhoemmen: Note that this only fences Kokkos' default CUDA stream.  Could that possibly be different than cuSPARSE's stream (that comes in through the handle)?

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+    cusparseSetMatFillMode (descrU, CUSPARSE_FILL_MODE_LOWER);
+  } else {
+    cusparseSetMatFillMode (descrU, CUSPARSE_FILL_MODE_UPPER);
+  }
+  cusparseSetMatDiagType (descrU, CUSPARSE_DIAG_TYPE_NON_UNIT);
+  cusparseSetMatType (descrU,CUSPARSE_MATRIX_TYPE_GENERAL);
+
+  // ==============================================
+  // step 2: query how much memory used in csrsv2, and allocate the buffer
+  // pBuffer returned by cudaMalloc is automatically aligned to 128 bytes.
```
mhoemmen: Why is 128 bytes important?

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+  } else {
+    const cuDoubleComplex alpha = make_cuDoubleComplex(1.0, 0.0);
+    status = cusparseZcsrsv2_solve (handle, transU, nrows, nnzU, &alpha, descrU,
+                                    reinterpret_cast <cuDoubleComplex*> (valuesU.data()), row_mapU.data(), entriesU.data(), infoU,
+                                    reinterpret_cast <cuDoubleComplex*> (sol.data()),
+                                    reinterpret_cast <cuDoubleComplex*> (rhs.data()),
+                                    policy, pBufferU);
+  }
+  Kokkos::fence();
+  std::cout << "  Cusparse Solve Time   : " << timer.seconds() << std::endl;
```
mhoemmen: Best practice is to get the timer result before printing, not during printing.  `std::cout` might buffer output, and `timer.seconds()` might not get called until after buffering.

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+  if (CUSPARSE_STATUS_SUCCESS != status) {
+    std::cout << "solve status error name " << (status) << std::endl;
```
mhoemmen: `status` is just some enum value, an integer.  cuSPARSE probably has a function for converting these enum values to strings, just like [the CUDA run-time API does](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__ERROR.html).  Also, you don't need parentheses around `status` here.

perf_test/sparse/KokkosSparse_sptrsv_cholmod.cpp
```diff
@@ -0,0 +1,579 @@
...
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact Siva Rajamanickam (srajama@sandia.gov)
+//
+// ************************************************************************
+//@HEADER
+*/
+
+#include <cstdio>
```
mhoemmen: Why do you need all these include files?  Which ones do you actually use?

perf_test/sparse/KokkosSparse_sptrsv_cholmod.cpp
```diff
@@ -0,0 +1,579 @@
...
+      return 0;
+    }
+  }
+
+  std::cout << std::endl;
+  for (size_t i = 0; i < tests.size(); ++i) {
+    std::cout << "tests[" << i << "] = " << tests[i] << std::endl;
+  }
+
+  Kokkos::initialize(argc,argv);
```
mhoemmen: Please use `Kokkos::ScopeGuard` instead of `initialize` and `finalize`.  That's just one line of code instead of two, and `ScopeGuard` will ensure that `finalize` gets called even if an uncaught exception happens.

src/common/KokkosKernels_Handle.hpp
```diff
@@ -48,9 +48,20 @@
...
 #ifndef _KOKKOSKERNELHANDLE_HPP
 #define _KOKKOSKERNELHANDLE_HPP
 
+#if defined(KOKKOSKERNELS_ENABLE_TPL_CBLAS)   && \
+    defined(KOKKOSKERNELS_ENABLE_TPL_LAPACKE) && \
+   (defined(KOKKOSKERNELS_ENABLE_TPL_SUPERLU) || \
+    defined(KOKKOSKERNELS_ENABLE_TPL_CHOLMOD))
+
+ // Enable supernodal sptrsv
+ #define KOKKOSKERNELS_ENABLE_SUPERNODAL_SPTRSV
```
mhoemmen: It would be better to make this a CMake option that users can disable even if all those TPLs are enabled.

src/common/KokkosKernels_Handle.hpp
```diff
@@ -209,6 +220,10 @@ class KokkosKernelsHandle
   typedef
     typename KokkosSparse::Experimental::SPTRSVHandle<const_size_type, const_nnz_lno_t, const_nnz_scalar_t, HandleExecSpace, HandleTempMemorySpace, HandlePersistentMemorySpace>
       SPTRSVHandleType;
+#ifdef KOKKOSKERNELS_ENABLE_SUPERNODAL_SPTRSV
+  typedef
```
mhoemmen: Prefer C++11 `using` syntax to `typedef`.
```c++
using integer_view_host_t = typename SPTRSVHandleType::integer_view_host_t;
```

src/common/KokkosKernels_Handle.hpp
```diff
@@ -48,9 +48,20 @@
...
+    defined(KOKKOSKERNELS_ENABLE_TPL_CHOLMOD))
+
+ // Enable supernodal sptrsv
+ #define KOKKOSKERNELS_ENABLE_SUPERNODAL_SPTRSV
+
+#endif
+
 namespace KokkosKernels{
 
 namespace Experimental{
```
mhoemmen: Given that Clark is planning on using this stuff, it sounds like the intent is for this to be production capability.  In that case, why put it in the Experimental namespace?  If it's enabled by default, you'll need to support it anyway -- at least you'll need to make the build work and tests pass.

src/common/KokkosKernels_Handle.hpp
```diff
@@ -549,19 +564,84 @@ class KokkosKernelsHandle
...
+  bool is_sptrsv_lower_tri () {
+    return this->sptrsvHandle->is_lower_tri ();
+  }
+
+  bool is_sptrsv_column_major () {
+    return this->sptrsvHandle->is_column_major ();
   }
+#endif
   void destroy_sptrsv_handle(){
     if (is_owner_of_the_sptrsv_handle && this->sptrsvHandle != nullptr)
```
mhoemmen: This looks like a reimplementation of `std::shared_ptr`.  Why does kokkos-kernels reimplement `std::shared_ptr`?  Is the idea that the handle needs to be valid on device?  We have `std::weak_ptr` in case we need to express the "nonowning observer that can take ownership at will" idea.

src/sparse/KokkosSparse_sptrsv.hpp
```diff
@@ -58,6 +58,16 @@
 #include "KokkosSparse_sptrsv_symbolic_spec.hpp"
 #include "KokkosSparse_sptrsv_solve_spec.hpp"
 
+#ifdef KOKKOSKERNELS_ENABLE_TPL_SUPERLU
+ #include "slu_ddefs.h"
```
mhoemmen: It's a REALLY BAD IDEA to expose the SuperLU headers to downstream code.  SuperLU is known to be bad about defining really short macros like `MIN` that collide with users' code.  Please hide SuperLU includes in a `.cpp` file; don't put them in headers that applications can see!

In general, please avoid exposing TPL headers (especially for C libraries that tend to pollute the global namespace) in headers that applications might see.

src/sparse/KokkosSparse_sptrsv.hpp
```diff
@@ -58,6 +58,16 @@
...
+ #include "KokkosSparse_sptrsv_superlu.hpp"
+#endif
+
+#ifdef KOKKOSKERNELS_ENABLE_TPL_CHOLMOD
+ #include "cholmod.h"
```
mhoemmen: See my comments above about exposing TPL headers.

src/sparse/KokkosSparse_sptrsv.hpp
```diff
@@ -210,10 +219,584 @@ namespace Experimental {
...
+      std::cout << std::endl
+                << " ** etree needs to be set before calling sptrsv_symbolic with SuperLU **"
+                << std::endl << std::endl;
+      return;
+    }
+
+    // ===================================================================
+    // read CrsGraph from SuperLU factor
+    #ifdef KOKKOS_SPTRSV_SUPERNODE_PROFILE
+    std::cout << " > Read SuperLU factor into KokkosSparse::CrsMatrix (invert diagonal and copy to device)" << std::endl;
```
mhoemmen: 1. The name "PROFILE" implies performance profiling, not printing verbose debugging output.
2. Why is this a compile-time option?  Why not make it a run-time option?  You could store in the handle a `bool` called "`verbose`" that tells functions to print verbose output like this.

src/sparse/KokkosSparse_sptrsv.hpp
```diff
@@ -210,10 +219,584 @@ namespace Experimental {
...
+    } else {
+      graphU_host = read_superlu_graphU<host_graph_t> (&L, &U); 
+    }
+    #ifdef KOKKOS_SPTRSV_SUPERNODE_PROFILE
+    std::cout << "   Conversion Time (from SuperLU to CSR): " << tic.seconds() << std::endl;
+    #endif
+
+    // ===================================================================
+    // setup supnodal info
+    SCformat *Lstore = (SCformat*)(L.Store);
```
mhoemmen: This looks like a cast from `void*`.  If so, then prefer C++ `reinterpret_cast<SCformat*>(L.Store)` to the C-style cast you did here.

src/sparse/KokkosSparse_sptrsv_cholmod.hpp
```diff
@@ -0,0 +1,139 @@
...
+///
+/// This file provides KokkosSparse::sptrsv.  This function performs a
+/// local (no MPI) sparse triangular solve on matrices stored in
+/// compressed row sparse ("Crs") format.
+
+#ifndef KOKKOSSPARSE_SPTRSV_CHOLMOD_HPP_
+#define KOKKOSSPARSE_SPTRSV_CHOLMOD_HPP_
+
+#ifdef KOKKOSKERNELS_ENABLE_TPL_CHOLMOD
+#include "cholmod.h"
```
mhoemmen: See my comments about about it being a bad idea to expose C headers from third-party libraries to downstream code.

src/sparse/KokkosSparse_sptrsv_handle.hpp
```diff
@@ -43,17 +43,29 @@
...
+ // Enable supernodal sptrsv
+ #define KOKKOSKERNELS_ENABLE_SUPERNODAL_SPTRSV
+ #include <KokkosSparse_CrsMatrix.hpp>
+
+#endif
+
 namespace KokkosSparse {
 namespace Experimental {
 
 // TP2 algorithm has issues with some offset-ordinal combo to be addressed
```
mhoemmen: What does this comment mean?  If some combinations of types don't work, you can use `static_assert` to make them fail at compile time.

src/sparse/KokkosSparse_sptrsv_handle.hpp
```diff
@@ -109,26 +143,111 @@ class SPTRSVHandle {
...
+
+  // verbose
+  bool verbose;
+#endif
+
 public:
 
-  SPTRSVHandle ( SPTRSVAlgorithm choice, const size_type nrows_, bool lower_tri_, bool symbolic_complete_ = false ) :
+  SPTRSVHandle ( SPTRSVAlgorithm choice, const size_type nrows_, bool lower_tri_, bool symbolic_complete_ = false, bool numeric_complete_ = false ) :
     level_list(),
```
mhoemmen: `Kokkos::View` default initialization will do this for you.  You don't need to explicitly default-construct `Kokkos::View` like this.

src/sparse/KokkosSparse_sptrsv_handle.hpp
```diff
@@ -109,26 +143,111 @@ class SPTRSVHandle {
...
     nodes_per_level(),
     nodes_grouped_by_level(),
     nrows(nrows_),
     nlevel(0),
```
mhoemmen: Prefer member initialization to explicit initialization in the constructor.  When you declare `nlevel` as a private member, say this:
```c++
int nlevel = 0;
```
That way, you don't have to remember to set it to zero in the constructor.  If you later add another constructor, you won't have to repeat code.

src/sparse/KokkosSparse_sptrsv_handle.hpp
```diff
@@ -109,26 +143,111 @@ class SPTRSVHandle {
...
+      // lower-triangular is stored in CSC
+      col_major = true;
+    } else {
+      // upper-triangular is stored in CSR
+      col_major = false;
+    }
+#endif
+  }
 
 #if 0
```
mhoemmen: If you don't intend to support this code, then please delete it.

src/sparse/KokkosSparse_sptrsv_superlu.hpp
```diff
@@ -0,0 +1,719 @@
...
+///
+/// This file provides KokkosSparse::sptrsv.  This function performs a
+/// local (no MPI) sparse triangular solve on matrices stored in
+/// compressed row sparse ("Crs") format.
+
+#ifndef KOKKOSSPARSE_SPTRSV_SUPERLU_HPP_
+#define KOKKOSSPARSE_SPTRSV_SUPERLU_HPP_
+
+#ifdef KOKKOSKERNELS_ENABLE_TPL_SUPERLU
+#include "slu_ddefs.h"
```
mhoemmen: See my above note about how exposing SuperLU headers to downstream code is a Bad Idea.

src/sparse/KokkosSparse_sptrsv_superlu.hpp
```diff
@@ -0,0 +1,719 @@
...
+                        'U', 'N', nscol,
+                        reinterpret_cast <lapack_complex_double*> (&Lx[psx]), nsrow);
+      }
+    }
+    int nnzD = hr(j1);
+    for (int j = 0; j < nscol; j++) {
+      for (int i = 0; i <= j; i++) {
+        hv(hr(j1 + j) + i) = Lx[psx + i + j*nsrow];
+      }
+#if 0
```
mhoemmen: Why is this code commented out?

src/sparse/KokkosSparse_sptrsv_superlu.hpp
```diff
@@ -0,0 +1,719 @@
...
+    int nscol = nb[k+1] - j1;
+
+    int i1 = mb[j1];
+    int nsrow = mb[j1+1] - i1;
+
+    /* the diagonal "dense" block (!! first !!)*/
+    int psx = colptrL[j1];
+    if (invert_diag) {
+      if (std::is_same<scalar_t, double>::value == true) {
+        LAPACKE_dtrtri (LAPACK_COL_MAJOR,
```
mhoemmen: I am not a fan of this.  You could have just copied and pasted the `extern "C"` declaration format for BLAS functions that kokkos-kernels already uses, to get calls to LAPACK's `_TRTRI` functions.  It's bad encapsulation to have this "if scalar type is this, call this, else call that, else ..." code pasted directly into the SuperLU code here.  You should just make this into a function, and change it to call regular LAPACK inside instead of LAPACKE.  You don't have to make it a public function if you don't intend to support it yet.

ndellingwood:
Spot-check on kokkos-dev-2 passes:

```
Going to test compilers:  gcc/7.3.0 gcc/9.1 intel/18.0.5 clang/8.0 cuda/10.1
Testing compiler gcc/7.3.0
Testing compiler gcc/9.1
  Starting job gcc-7.3.0-Pthread-release
  Starting job gcc-7.3.0-OpenMP-release
  PASSED gcc-7.3.0-Pthread-release
  Starting job gcc-9.1-OpenMP-release
  PASSED gcc-7.3.0-OpenMP-release
Testing compiler intel/18.0.5
  Starting job gcc-9.1-Serial-release
  PASSED gcc-9.1-OpenMP-release
Testing compiler clang/8.0
  Starting job intel-18.0.5-OpenMP-release
  PASSED gcc-9.1-Serial-release
  Starting job clang-8.0-Cuda_OpenMP-release
  PASSED intel-18.0.5-OpenMP-release
Testing compiler cuda/10.1
  Starting job clang-8.0-Pthread_Serial-release
  PASSED clang-8.0-Cuda_OpenMP-release
  PASSED clang-8.0-Pthread_Serial-release
  Starting job cuda-10.1-Cuda_OpenMP-release
  PASSED cuda-10.1-Cuda_OpenMP-release
PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=605 run_time=518
clang-8.0-Pthread_Serial-release build_time=248 run_time=333
cuda-10.1-Cuda_OpenMP-release build_time=646 run_time=481
gcc-7.3.0-OpenMP-release build_time=235 run_time=177
gcc-7.3.0-Pthread-release build_time=215 run_time=176
gcc-9.1-OpenMP-release build_time=261 run_time=162
gcc-9.1-Serial-release build_time=238 run_time=181
intel-18.0.5-OpenMP-release build_time=441 run_time=160
```

src/sparse/KokkosSparse_sptrsv_superlu.hpp
```diff
@@ -0,0 +1,719 @@
...
+                        'U', 'N', nscol,
+                        reinterpret_cast <double*> (&Lx[psx]), nsrow);
+      } else {
+        LAPACKE_ztrtri (LAPACK_COL_MAJOR,
+                        'U', 'N', nscol,
+                        reinterpret_cast <lapack_complex_double*> (&Lx[psx]), nsrow);
+      }
```
mhoemmen: What about `float` and `complex<float>`?  kokkos-kernels is supposed to support at least the four types that the BLAS supports.  Please at least `static_assert` or throw here.

src/sparse/KokkosSparse_sptrsv_superlu.hpp
```diff
@@ -0,0 +1,719 @@
...
+        hv(hr(jcol) + ioff) = Uval[i];
+      }
+      // move up the pointers for all the supernodal blocks
+      hr(jcol) += offset;
+    }
+
+    if (invert_diag) {
+      if (offset > 0 && invert_offdiag) {
+        if (std::is_same<scalar_t, double>::value == true) {
+          cblas_dtrmm (CblasColMajor,
```
mhoemmen: My comments about LAPACKE apply here to CBLAS.

src/sparse/KokkosSparse_sptrsv_superlu.hpp
```diff
@@ -0,0 +1,719 @@
...
+                     reinterpret_cast <double*> (&hv(nnzD+nscol)), nscol+offset);
+        } else {
+          // NOTE: use double pointers
+          cblas_ztrmm (CblasColMajor,
+                CblasRight, CblasUpper, CblasNoTrans, CblasNonUnit,
+                offset, nscol,
+                reinterpret_cast <double*> (&one), 
+                reinterpret_cast <double*> (&hv(nnzD)), nscol+offset,
+                reinterpret_cast <double*> (&hv(nnzD+nscol)), nscol+offset);
+        }
```
mhoemmen: What about `float` and `complex<float>`?

cmake/Dependencies.cmake
```diff
@@ -1,5 +1,5 @@
 TRIBITS_PACKAGE_DEFINE_DEPENDENCIES(
   LIB_REQUIRED_PACKAGES KokkosCore KokkosContainers KokkosAlgorithms 
-  LIB_OPTIONAL_TPLS quadmath MKL BLAS LAPACK CUSPARSE MAGMA
+  LIB_OPTIONAL_TPLS quadmath MKL BLAS LAPACK CUSPARSE MAGMA SUPERLU CHOLMOD LAPACKE CBLAS
```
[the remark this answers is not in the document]
jjwilke: Not that I know of.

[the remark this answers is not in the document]
jjwilke: We might have a bootstrap issue here for LAPACKE and CBLAS. The tpls don't exist in Trilinos (AFAIK) so adding the TPLs to `Dependencies.cmake` would also require adding Tribits find modules.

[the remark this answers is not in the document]
jjwilke: Trilinos does have find modules for SuperLU and Cholmod, but with different capitalization.
`SUPERLU -> SuperLU` and `CHOLMOD -> Cholmod`.

This problem of capitalize vs dont' capitalize is going to follow us forever...

ndellingwood:
@mhoemmen thanks for thorough review! I skimmed some comments and can provided a couple points of clarification I'm aware of, and let @iyamazaki add other responses.

Magma - we do not have nightly testing running with magma, though the tpls and scripts are setup to handle this on apollo (where development and PR testing was done of the capability). I'll get this setup soon, dropped the ball on it.

Experimental - the code hasn't been used yet outside of development and the API is finalized yet, hope is to get some feedback and review. Longer term, I think the plan is to expose the solver through Amesos2.

cmake/Dependencies.cmake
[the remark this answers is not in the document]
mhoemmen: @jjwilke I'm not sure how easy it would be to change these in a backwards-compatible way so that Trilinos could accept either capitalization, even for specific TPLs. @bartlettroscoe might have some ideas.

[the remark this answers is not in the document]
bartlettroscoe: > I'm not sure how easy it would be to change these in a backwards-compatible way so that Trilinos could accept either capitalization, even for specific TPLs.

This is an ugly problem.  The names and capitalization of these TPLs should have been considered very carefully from the beginning.  But with people breaking backward compatiblity left and right right now (because of the anticipated Trilinos 13.0 release) this is a great time to fix a bunch of these issues.

@jjwilke, can we get a list of TPLs defined in {{Trilinos/TPLsList.cmake}} that should be renamed or recapitalized to match standard CMake names for these TPLs? 

I think this is why `find_package()` can also find `<lower_case_name>_config.cmake` filkes in addition to `<UpperCaseName>Config.cmake` files as this solves the capitalization problem, at least for installed packages that provide config.cmake files.

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+// Questions? Contact Siva Rajamanickam (srajama@sandia.gov)
+//
+// ************************************************************************
+//@HEADER
+*/
+
+#ifndef KOKKOSSPARSE_SPTRSV_AUX
+#define KOKKOSSPARSE_SPTRSV_AUX
+
+/* ========================================================================================= */
```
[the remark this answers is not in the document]
iyamazaki: We put them inside KokkosSparse::PerfTest::Experimental namespace.

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+  using execution_space = typename scalar_view_t::execution_space;
+
+  mag_t ZERO = mag_t(0);
+  scalar_t ONE = scalar_t(1);
+
+  // normB
+  mag_t normB = ZERO;
+  Kokkos::parallel_reduce( Kokkos::RangePolicy<execution_space>(0, rhs.extent(0)), 
+    KOKKOS_LAMBDA ( const lno_t i, mag_t &tsum ) {
+      tsum += STS::abs (rhs(i)) * STS::abs (rhs(i));
```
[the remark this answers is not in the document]
iyamazaki: We switched to calling KokkosBlas::nrm2.

cmake/Dependencies.cmake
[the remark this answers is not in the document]
mhoemmen: @bartlettroscoe wrote:
> But with people breaking backward compatiblity left and right right now (because of the anticipated Trilinos 13.0 release) this is a great time to fix a bunch of these issues.

Is there a way to add "aliases" for TPL names?  Could I set up a TPL so that users could say `TPL_ENABLE_Bar=ON` and `TPL_ENABLE_Foo=ON` gets set?

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+  row_map_view_host_t hr = Kokkos::create_mirror_view (row_map);
+  cols_view_host_t    hc = Kokkos::create_mirror_view (entries);
+  values_view_host_t  hv  = Kokkos::create_mirror_view (values);
+
+  Kokkos::deep_copy (hr, row_map);
+  Kokkos::deep_copy (hc, entries);
+  Kokkos::deep_copy (hv, values);
+
+  // compress
+  int nnzA0 = hr (n);
```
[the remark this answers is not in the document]
iyamazaki: We switched to using size_type.

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
```diff
@@ -0,0 +1,626 @@
...
+  using memory_space    = typename execution_space::memory_space;
+
+  using host_execution_space = typename host_values_view_t::execution_space;
+  using host_memory_space    = typename host_execution_space::memory_space;
+
+  using host_scalar_view_t = Kokkos::View< scalar_t*, host_memory_space >;
+  using      scalar_view_t = Kokkos::View< scalar_t*,      memory_space >;
+
+  scalar_t ZERO = scalar_t (0);
+  scalar_t ONE = scalar_t (1);
```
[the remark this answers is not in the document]
iyamazaki: We started switching to this, but then got this error: error: calling the default constructor for "Kokkos::complex<double>" does not produce a constant value.

mhoemmen:
@ndellingwood wrote:
> Magma - we do not have nightly testing running with magma, though the tpls and scripts are setup to handle this on apollo (where development and PR testing was done of the capability). I'll get this setup soon, dropped the ball on it.

No worries!  I can explain in person why MAGMA might be more important for us in the future, but there's no rush.

cmake/Dependencies.cmake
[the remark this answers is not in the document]
bartlettroscoe: > Is there a way to add "aliases" for TPL names? Could I set up a TPL Foo so that users could say TPL_ENABLE_Bar=ON and TPL_ENABLE_Foo=ON gets set?

@mhoemmen, not sure.  Would have to look at all of the use cases that would need to work (and it is easy to leave out use cases that don't occur to one right away).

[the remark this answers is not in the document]
mhoemmen: @bartlettroscoe The main reason would be to support changes in capitalization.  On the other hand, CMake already tells you about unused options, so perhaps that's good enough to support a change.

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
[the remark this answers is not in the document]
iyamazaki: For now, I switched to using:
 const scalar_t ZERO (0.0);
 const scalar_t ONE (1.0);
I hope this is okay..

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
iyamazaki: We removed "== true".

perf_test/sparse/KokkosSparse_sptrsv_cholmod.cpp
iyamazaki: We removed the ones not being used..

src/common/KokkosKernels_Handle.hpp
iyamazaki: We switched to "using"!!

perf_test/sparse/KokkosSparse_sptrsv_cholmod.cpp
iyamazaki: We switched to using Kokkos::ScopeGuard!!

perf_test/sparse/KokkosSparse_sptrsv_aux.hpp
iyamazaki: We moved the timer out of std::cout.

ndellingwood:
@iyamazaki I put in a PR against this branch on your fork iyamazaki/kokkos-kernels#1 that merges develop back in and resolves conflicts. If it passes testing go ahead and merge it into your branch, then push your updated sptrsv-supernode branch back up to this PR.

ndellingwood:
Spoke with @srajama1, this PR will go in for the promotion and allow for development; we'll resolve the TPL issues for the next promotion, the goal being to provide proper kokkos-kernels implementations for the missing BLAS and LAPACK routines needed (called from cblas and lapacke currently).
The code is well-guarded and integration testing with the changes in this PR passed.
spot-check on kokkos-dev-2 passed, results posted [here](https://github.com/iyamazaki/kokkos-kernels/pull/1#issuecomment-572252542) during work to resolve merge conflicts.

ndellingwood:
Merging in for release 3.0
@iyamazaki please create an issue with follow up items, like support for needed blas/lapack routines used via cblas and lapacke, review suggestions from @mhoemmen not yet addressed, etc.

Note the spot-check on kokkos-dev-2 has results posted in a PR to @iyamazaki branch when the merge conflicts were resolved.

PR #558, opened 2020-01-07
cm_test_all_sandia: update clang8 flags for clang+cuda builds

ndellingwood:
Add -Wno-pass-failed to remove warnings with Cuda_OpenMP builds
with clang/8 + cuda/10 in kokkos-dev-2


ndellingwood:
This change needed to disable warnings triggered by `-Werror` clang/8 + cuda/10 builds like the following:

```
/ascldap/users/ndellin/kokkos-kernels/src/blas/impl/KokkosBlas3_gemm_impl.hpp:402:63: error: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Werror,-Wpass-failed=transform-warning]
  Kokkos::parallel_for(Kokkos::TeamThreadRange(team,blockA0), [&] (const int i) {
```

Issue #559, opened 2020-01-08
trsm_complex_double fails under OpenMP and Serial, on Intel 18 only


brian-kelley:
I ran spot checks on Bowman with double and complex_double enabled as scalars. On intel 16 and 17, everything passes. But on intel 18, complex trsm fails. Probably a compiler bug...

```
[ RUN      ] openmp.trsm_complex_double
/ascldap/users/bmkelle/FixSortingTest/kokkos-kernels/unit_test/blas/Test_Blas3_trsm.hpp:187: Failure
Value of: true
Expected: test_flag
Which is: false
/ascldap/users/bmkelle/FixSortingTest/kokkos-kernels/unit_test/blas/Test_Blas3_trsm.hpp:187: Failure
Value of: true
Expected: test_flag
Which is: false
/ascldap/users/bmkelle/FixSortingTest/kokkos-kernels/unit_test/blas/Test_Blas3_trsm.hpp:187: Failure
Value of: true
Expected: test_flag
Which is: false
/ascldap/users/bmkelle/FixSortingTest/kokkos-kernels/unit_test/blas/Test_Blas3_trsm.hpp:187: Failure
Value of: true
Expected: test_flag
Which is: false
/ascldap/users/bmkelle/FixSortingTest/kokkos-kernels/unit_test/blas/Test_Blas3_trsm.hpp:187: Failure
Value of: true
Expected: test_flag
Which is: false
/ascldap/users/bmkelle/FixSortingTest/kokkos-kernels/unit_test/blas/Test_Blas3_trsm.hpp:187: Failure
Value of: true
Expected: test_flag
Which is: false
```

crtrott:
Before you blame a compiler bug test this with Intel 19 (for example on Blake compiling for Skylake (SKX)). It might be that it is doing an optimization now which exposes a code bug.

vqd8a:
I just want to say I have run double complex TRSM tests on Blake with different Intel compilers. The tests passed on Intel 17 and Intel 19 but failed on Intel 18 as @brian-kelley saw on Bowman. I think it is possibly an Intel 18 compiler bug.

| Compilers | using MKL    | Fallback implementation    |
|-------------|------------|------------|
|intel 17.4.196       | passed      | passed     |
|intel 18.0.128       | passed      | **failed**       |
|intel 18.1.163      | passed      | **failed**        |
|intel 19.1.144      | passed      | passed      |
|intel 19.3.199      | passed      | passed      |

[...]

Issue #344, opened 2018-11-15
Various complex<double> unit test fails with XL/16.1 OpenMP build


ndellingwood:
Output from Jenkins:

Failed Test 1
```
06:46:10 [ RUN      ] openmp.gemm_complex_double
06:46:10 /home/jenkins/white/workspace/KokkosKernels_White_XL_16_1_OpenMP_Serial/kokkos-kernels/unit_test/blas/Test_Blas3_gemm.hpp:147: Failure
06:46:10 Value of: (diff_C_average < 1.05*diff_C_expected )
06:46:10   Actual: false
06:46:10 Expected: true
etc.
```

Failed Test 2
```
06:46:10 [ RUN      ] openmp.batched_scalar_team_trsm_l_u_nt_n_dcomplex_dcomplex
06:46:10 /home/jenkins/white/workspace/KokkosKernels_White_XL_16_1_OpenMP_Serial/kokkos-kernels/unit_test/../test_common/KokkosKernels_TestUtils.hpp:87: Failure
06:46:10 The difference between double(AT1::abs(val1)) and double(AT2::abs(val2)) is 0.99556106126897947, which exceeds double(AT3::abs(tol)), where
06:46:10 double(AT1::abs(val1)) evaluates to 0.99556106126897947,
06:46:10 double(AT2::abs(val2)) evaluates to 0, and
06:46:10 double(AT3::abs(tol)) evaluates to 2.2204460492503131e-13.
```

Failed Test 3
```
06:46:10 [ RUN      ] openmp.batched_scalar_team_trsm_l_u_nt_n_dcomplex_double
06:46:10 /home/jenkins/white/workspace/KokkosKernels_White_XL_16_1_OpenMP_Serial/kokkos-kernels/unit_test/../test_common/KokkosKernels_TestUtils.hpp:87: Failure
06:46:10 The difference between double(AT1::abs(val1)) and double(AT2::abs(val2)) is 0.99580002746667939, which exceeds double(AT3::abs(tol)), where
06:46:10 double(AT1::abs(val1)) evaluates to 0.99580002746667939,
06:46:10 double(AT2::abs(val2)) evaluates to 0, and
06:46:10 double(AT3::abs(tol)) evaluates to 2.2204460492503131e-13.
```

ndellingwood:
@kyungjoo-kim would you have time to look at the gemm test?
@vqd8a would you have time to look at the trsm tests?

kyungjoo-kim:
[...]

PR #562, opened 2020-01-13
Don't override Trilinos/user F77_BLAS_MANGLE

brian-kelley:
If user set F77_BLAS_MANGLE to a custom value, don't overwrite it
using the automatic value. If it was set by Trilinos, don't change it.

If there is no fortran mangle suffix, don't add the macro append "##"
operator, since that generates invalid syntax.

For a little more detail, if ``FortranCInterface_GLOBAL_SUFFIX`` is empty then the macro definition was left with something like "name ##" which is not allowed (there must be a token on both sides of ``##``).

I actually ran into both of these issues at the same time when building on Vortex with GCC. In that case both bugs triggered because
- it's an IBM system so CMake's FortranCInterface thought the mangle scheme was "name" (no underscore)
- I was using the version of LAPACK for GCC, so to avoid the above issue in Trilinos I was manually overriding F77_BLAS_MANGLE in Trilinos to be "name ## _"

I haven't actually tested this at all, but @jjwilke  I would like to see what you think about these changes before I do the spot checks.


mhoemmen:
Note that on some IBM systems, the Fortran compiler's mangling scheme differs from the BLAS' mangling scheme.

ndellingwood:
@brian-kelley go ahead and start spot-checks, please test on an ibm and non-ibm system

brian-kelley:
Spot checks (RIDE is an IBM system). Failures are unrelated - due to #559 and #344.

RIDE:
```
PASSED TESTS
cuda-9.2.88-Cuda_OpenMP-release build_time=592 run_time=416
cuda-9.2.88-Cuda_Serial-release build_time=583 run_time=523
gcc-6.4.0-OpenMP_Serial-release build_time=193 run_time=394
gcc-7.2.0-OpenMP-release build_time=135 run_time=128
gcc-7.2.0-OpenMP_Serial-release build_time=203 run_time=360
gcc-7.2.0-Serial-release build_time=124 run_time=227
ibm-16.1.0-Serial-release build_time=510 run_time=397
FAILED TESTS
ibm-16.1.0-OpenMP-release (test failed)
ibm-16.1.0-OpenMP_Serial-release (test failed)
```

Bowman:
```
PASSED TESTS
intel-16.4.258-Pthread-release build_time=719 run_time=1025
intel-16.4.258-Pthread_Serial-release build_time=1036 run_time=2007
intel-16.4.258-Serial-release build_time=705 run_time=948
intel-17.2.174-OpenMP-release build_time=879 run_time=563
intel-17.2.174-OpenMP_Serial-release build_time=1226 run_time=1471
intel-17.2.174-Pthread-release build_time=800 run_time=880
intel-17.2.174-Pthread_Serial-release build_time=1129 run_time=1814
intel-17.2.174-Serial-release build_time=770 run_time=888
FAILED TESTS
intel-18.2.199-OpenMP-release (test failed)
intel-18.2.199-OpenMP_Serial-release (test failed)
intel-18.2.199-Pthread-release (test failed)
intel-18.2.199-Pthread_Serial-release (test failed)
intel-18.2.199-Serial-release (test failed)
```

PR #563, opened 2020-01-13
test_all_sandia: remove intel/18 testing

ndellingwood:
remove from testing for now due to issue reported in #559


srajama1:
I will push after spot-check, Thanks @ndellingwood !

ndellingwood:
@srajama1 intel/19 is not available on bowman

PR #569, opened 2020-01-16
gemv: call scal for degenerate matrices

ndellingwood:
Related to #539

PR #570, opened 2020-01-17
gemv: call scal for degenerate matrices

ndellingwood:
Related to #539

### DIFF
--- a/src/blas/KokkosBlas2_gemv.hpp
+++ b/src/blas/KokkosBlas2_gemv.hpp
@@ -52,6 +52,8 @@
 #include <sstream>
 #include <type_traits> // requires C++11, but so does Kokkos
 
+#include <KokkosBlas1_scal.hpp>
+
 namespace KokkosBlas {
 
 /// \brief Dense matrix-vector multiply: y = beta*y + alpha*A*x.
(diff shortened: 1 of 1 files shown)


src/blas/KokkosBlas2_gemv.hpp
```diff
@@ -52,6 +52,8 @@
 #include <sstream>
 #include <type_traits> // requires C++11, but so does Kokkos
 
+#include <KokkosBlas1_scal.hpp>
```
mhoemmen: @ndellingwood In theory, if you want to call a KokkosBlas function from another KokkosBlas function, you might want to call the low-level ones that live in separate headers.  Otherwise, circular header dependencies could show up.  It's probably unlikely you would get circular dependencies, though.

ndellingwood:
Cherry pick of PR #569

src/blas/KokkosBlas2_gemv.hpp
ndellingwood: @mhoemmen just to clarify, I should call to the unification layer in the spec file?

https://github.com/kokkos/kokkos-kernels/blob/d675a3b83f2c6fb8d5f0263cdd9bf3ee1301697b/src/blas/impl/KokkosBlas1_scal_spec.hpp#L120-L125

ndellingwood: and then letting `scal_tpl_spec_avail` and `scal_eti_spec_avail` handle the compile time deduction of the appropriate lower-level call based on the given types in the public gemv routine

ndellingwood: I had previously called the fallback impl of gemv for the degenerate case (transpose mode) but had overly restrictive checks, as issues also came up when testing host blas, I replaced with `scal` here but can return to using the fallback `gemv` for all degenerate cases (i.e. equivalent result as calling `scal`)

ndellingwood:
Will use solution from #573 instead to prevent potential circular dependency issues.

PR #575, opened 2020-01-22
Fix #574 (improve KokkosBlas::dot accuracy for float & complex<float>)

### LINKED ISSUE #574: KokkosBlas::dot: Poor accuracy with float on Mac Clang causes Trilinos test failure
See https://github.com/trilinos/Trilinos/issues/6633.

mhoemmen:
For vectors of `float`, the two-argument overload of `KokkosBlas::dot` (that takes two vectors (rank-1 Views) as input, and returns the dot product result as a scalar value on host) gets inaccurate results on Mac Clang.  This commit fixes that overload of `KokkosBlas::dot` so that if the dot product result type is `float` resp. `complex<float>`, the function uses `double` resp. `complex<double>` as the type for intermediate sums.

This change, when applied to Trilinos in Pull Request https://github.com/trilinos/Trilinos/pull/6634, fixes Trilinos issue https://github.com/trilinos/Trilinos/issues/6633.

Closes #574

### DIFF
--- a/src/Kokkos_InnerProductSpaceTraits.hpp
+++ b/src/Kokkos_InnerProductSpaceTraits.hpp
@@ -160,8 +160,8 @@ public:
   /// complex.  In that case, see the partial specialization for
   /// Kokkos::complex below to see our convention for which input gets
   /// conjugated.
-  static KOKKOS_FORCEINLINE_FUNCTION dot_type
-  dot (const val_type& x, const val_type& y) {
+  static KOKKOS_FORCEINLINE_FUNCTION
+  dot_type dot (const val_type& x, const val_type& y) {
     return x * y;
   }
 };
@@ -196,8 +196,8 @@ public:
   mag_type norm (const val_type& x) {
     return ArithTraits<val_type>::abs (x);
   }
-  static KOKKOS_FORCEINLINE_FUNCTION dot_type
-  dot (const val_type& x, const val_type& y) {
+  static KOKKOS_FORCEINLINE_FUNCTION
+  dot_type dot (const val_type& x, const val_type& y) {
     return Kokkos::conj (x) * y;
   }
 };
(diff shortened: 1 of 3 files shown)


ndellingwood:
Thanks @mhoemmen ! Testing on kokkos-dev-2 now.

lucbv:
@mhoemmen is there any worry that doing the reduction in higher precision increases the cost of the reduction?
This might not be what everyone wants, would it make sense to have a mechanism in place for people who want real low precision arithmetic to get it?

mhoemmen:
Btw, I had forgotten to test after making a change to include `complex<float>` in the list of types that get extended precision.  A new change is coming.

@lucbv wrote:
> Is there any worry that doing the reduction in higher precision increases the cost of the reduction?

1. The change only uses higher precision for `float` and `complex<float>`.  I have one customer who cares about `Scalar=float`, and their main motivations are memory usage and sparse mat-vec performance.
2. Recent GPUs compute with `double` fast enough that we shouldn't need to worry about this.

lucbv:
I get your point but I still feel that a user might easily be frustrated that the library does not provide the most logical implementation as default but instead provides code that is only needed for a corner case, i.e. for a given compiler on a given platform... I would be happier if we can find a way to apply your fix to the clang/mac combination only.

mhoemmen:
@lucbv Why would we ever want less accurate dot product results?

ndellingwood:
Tests relaunched, thanks @mhoemmen !

ndellingwood:
This PR necessary to avoid clobbering changes in Trilinos, will be cherry-picked to the release-candidate-3.0 branch after merge.

ndellingwood:
Cross-referencing matched changes in merged PR trilinos/Trilinos#6634

mhoemmen:
@lucbv Sorry if I sounded a bit flippant there.  I've just noticed that tests with Scalar=`float` or `complex<float>` occasionally fail on other platforms too.  People historically work around that by relaxing the tolerances to near-uselessness.  I'd rather fix the problem by doing the Right Thing that generally will improve accuracy.

mhoemmen:
@ndellingwood wrote:
> This PR necessary to avoid clobbering changes in Trilinos, will be cherry-picked to the release-candidate-3.0 branch after merge.

I'm totally OK if you want to solve this problem in a different way -- Trilinos does not depend on _how_ the problem was solved.

ndellingwood:
Passed spot-check on kokkos-dev-2:

```

Repository Status:  66c2a6c7c66e2a2e74f4c3a6ebf1694a483eb41e Fix #574 (improve KokkosBlas::dot accuracy for float)


Going to test compilers:  gcc/7.3.0 gcc/8.3.0 gcc/9.1 intel/18.0.5 clang/8.0 cuda/10.1
...

PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=116 run_time=151
clang-8.0-Pthread_Serial-release build_time=49 run_time=152
cuda-10.1-Cuda_OpenMP-release build_time=156 run_time=140
gcc-7.3.0-OpenMP-release build_time=44 run_time=47
gcc-7.3.0-Pthread-release build_time=40 run_time=75
gcc-8.3.0-Serial-release build_time=41 run_time=71
gcc-9.1-OpenMP-release build_time=53 run_time=46
gcc-9.1-Serial-release build_time=45 run_time=71
intel-18.0.5-OpenMP-release build_time=132 run_time=46
```

ndellingwood:
@mhoemmen thanks for the PR and improvements! I'll get this merged shortly (release candidate is frozen pending another PR, once complete I'll line up the ready-PRs for merge + cherry-pick)
@lucbv brought up a valid point that can be revisited later towards a solution supporting lower precision.

srajama1:
@lucbv I think users will notice it only if they were looking for lower precision :). Yes we could give the standard implementation as an option once we have a handle for all our calls.

PR #578, opened 2020-01-23
Dist1 coloring PPS improvements

brian-kelley:
Distance-1 coloring with parallel prefix sum worklist used to require 2 passes: a scan to get new worklist index view, and a for to populate the worklist. The original paper about it even says that the 2-pass approach is a downside of PPS compared to atomics.

This makes PPS use a single scan pass that scatters directly to the new worklist without an index view. This saves some time, one kernel launch per iteration and a significant amount of memory: ``(num_edges / 2) * sizeof(nnz_lno_t)`` for EB, and ``num_rows * sizeof(nnz_lno_t)`` for VB.

Also, renamed ``_conflictlist`` member to ``_conflictlist_scheme`` for clarity - it's not a list, it's an enum representing how the worklist is built. I also use those actual enum constants (``COLORING_NOCONFLICT``, ``COLORING_ATOMIC``, ``COLORING_PPS``) instead of the raw integer values (0,1,2) throughout the code.

Bowman checks:
PASSED TESTS
intel-16.4.258-Pthread-release build_time=724 run_time=1027
intel-16.4.258-Pthread_Serial-release build_time=1034 run_time=1998
intel-16.4.258-Serial-release build_time=706 run_time=963
intel-17.2.174-OpenMP-release build_time=869 run_time=566
intel-17.2.174-OpenMP_Serial-release build_time=1225 run_time=1444
intel-17.2.174-Pthread-release build_time=814 run_time=899
intel-17.2.174-Pthread_Serial-release build_time=1131 run_time=1803
intel-17.2.174-Serial-release build_time=778 run_time=891

RIDE checks:
PASSED TESTS
cuda-9.2.88-Cuda_OpenMP-release build_time=517 run_time=418
cuda-9.2.88-Cuda_Serial-release build_time=507 run_time=527
gcc-6.4.0-OpenMP_Serial-release build_time=194 run_time=396
gcc-7.2.0-OpenMP-release build_time=120 run_time=128
gcc-7.2.0-OpenMP_Serial-release build_time=214 run_time=357
gcc-7.2.0-Serial-release build_time=112 run_time=226

### DIFF
--- a/src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
+++ b/src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
@@ -2838,9 +2838,6 @@ public:
     size_type_temp_work_view_t new_edge_conflict_indices
     (Kokkos::ViewAllocateWithoutInitializing("new_edge_conflict_indices"), num_work_edges);
 
-    size_type_temp_work_view_t
-    pps(Kokkos::ViewAllocateWithoutInitializing("prefix_sum"), num_work_edges);
-
     char_temp_work_view_type edge_conflict_marker
     (Kokkos::ViewAllocateWithoutInitializing("edge_conflict_marker"), num_work_edges);
 
--- a/src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
+++ b/src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
@@ -2257,10 +2257,10 @@ public:
     view_type _recolorList;
 
     ppsWorklistFunctorVB(
-        nnz_lno_t nv,
+        nnz_lno_t nv_,
         const view_type& vertexList,
         const view_type& recolorList)
-      : _nv(nv), _vertexList(vertexList), _recolorList(recolorList)
+      : _nv(nv_), _vertexList(vertexList), _recolorList(recolorList)
     {}
 
     KOKKOS_INLINE_FUNCTION
--- a/src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
+++ b/src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
@@ -789,7 +789,7 @@ protected:
 
   bool _serialConflictResolution; //if true use serial conflict resolution
   bool _ticToc; //if true print info in each step
-  char _conflictlist; //0 for no conflictlist, 1 for atomic, 2 for pps
+  ConflictList _conflict_scheme; //Enum: COLORING_NOCONFLICT, COLORING_ATOMIC, COLORING_PPS
 
   double _pps_ratio; //the minimum number of reduction on the size of the conflictlist to create a new conflictlist
   nnz_lno_t _min_vertex_cut_off; //minimum number of vertices to reduce the conflictlist further.
@@ -819,7 +819,7 @@ public:
     GraphColor<HandleType,lno_row_view_t_,lno_nnz_view_t_>(nv_, ne_, row_map, entries, coloring_handle),
     _serialConflictResolution(coloring_handle->get_serial_conflict_resolution()),
     _ticToc(coloring_handle->get_tictoc()),
-    _conflictlist(),
+    _conflict_scheme(coloring_handle->get_conflict_list_type()),
     _pps_ratio(coloring_handle->get_min_reduction_for_conflictlist()),
     _min_vertex_cut_off(coloring_handle->get_min_elements_for_conflictlist()),
     _edge_filtering(coloring_handle->get_vb_edge_filtering()),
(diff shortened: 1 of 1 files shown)


src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
```diff
@@ -1248,23 +1223,15 @@ class GraphColor_VB:public GraphColor <HandleType,lno_row_view_t_,lno_nnz_view_t
...
-
         Kokkos::parallel_scan ("KokkosGraph::GraphColoring::PrefixSum",
             my_exec_space(0, current_vertexListLength_),
-            parallel_prefix_sum<nnz_lno_temp_work_view_t>(current_vertexList_, next_iteration_recolorList_, pps_work_view));
-
-        MyExecSpace().fence();
-        Kokkos::parallel_for ("KokkosGraph::GraphColoring::CreateNewWorkArray",
-            my_exec_space(0, current_vertexListLength_),
-            create_new_work_array<nnz_lno_temp_work_view_t>(current_vertexList_, next_iteration_recolorList_, pps_work_view));
+            ppsWorklistFunctorVB<nnz_lno_temp_work_view_t>(this->nv, current_vertexList_, next_iteration_recolorList_));
```
srajama1: Does the values in recolocList need to be reinitialized ?

src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
```diff
@@ -2293,63 +2250,32 @@ class GraphColor_VB:public GraphColor <HandleType,lno_row_view_t_,lno_nnz_view_t
   };
 
 
-  /**
-   * Functor for parallel prefix sum
-   */
-  template <typename view_type>
-  struct parallel_prefix_sum{
```
srajama1: I thought this was used in EB too ? Does it ?

src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
```diff
@@ -2293,63 +2250,32 @@ class GraphColor_VB:public GraphColor <HandleType,lno_row_view_t_,lno_nnz_view_t
...
-    }
-  };
-
-
-  /**
-   * Functor for creating new worklist using pps
-   */
   template <typename view_type>
-  struct create_new_work_array{
+  struct ppsWorklistFunctorVB {
```
srajama1: Is the idea we are replacing generic PPS with something that works mainly for VB ?

src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
```diff
@@ -3287,62 +3201,36 @@ class GraphColor_EB:public GraphColor <HandleType,in_row_index_view_type_,in_non
...
-    size_type_temp_work_view_t _pps_view;
-
-    parallel_prefix_sum(
-        size_type_temp_work_view_t edge_conflict_indices,
-        char_temp_work_view_type edge_conflict_marker,
-        size_type_temp_work_view_t pps_view):
-          _edge_conflict_indices(edge_conflict_indices),
-          _edge_conflict_marker(edge_conflict_marker),
-          _pps_view(pps_view){}
+  struct ppsWorklistFunctorEB {
```
srajama1: I think what I assumed above was correct, based on this functor. Looks like we have two PPS functors now.

src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
brian-kelley: @srajama1 It doesn't need to be initialized because the exact size of the next worklist is produced by a parallel_reduce of functorFindConflicts.

src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
brian-kelley: @srajama1 VB and EB are slightly different now, but the single-pass scan idea is the same. VB is a scan over the current vertex worklist, and the scan populates the next worklist with vertices that were uncolored. These get marked in the current worklist by adding ``nv`` to them.

EB is a scan over the current edge worklist, and the next worklist is filled with edges where ``edge_conflict_marker(i)`` is nonzero.

srajama1: I am ok with two different scans.

src/graph/impl/KokkosGraph_Distance1Color_impl.hpp
srajama1: I meant reinitialized, from second iteration to third iteration

PR #582, opened 2020-01-28
sptrsv: cusparse fix

ndellingwood:
throw error when requesting to use cusparse on host-allocated data


ndellingwood:
Issue caught during Trilinos integration testing

PR #585, opened 2020-01-29
CUDA graph wrapper, using in Gauss-Seidel apply

brian-kelley:
(sorry for the wall of text, but some of this can become permanent documentation)

CUDA graphs (CUDA 10.1+) can eliminate most of the launch overhead in a repeated sequence of kernel launches, where during each repeat the kernel inputs are the same. For Kokkos, a graph is constructed by recording a sequence of kernel launches during a dry run (the kernel inputs are known, but the kernels aren't actually executed during recording).

This adds a user-friendly CUDA graph wrapper in ExecSpaceUtils. Currently, only a linear sequence of kernels are supported. Eventually it might be worth it to support multiple streams and an arbitrary dependency graph.

Depending on the complexity of the kernel inputs, using the wrapper for a set of kernels can involve as few as 5 new lines of code.

The wrapper has an arbitrary ``LaunchParams`` type that represents the input to kernels in the graph. If these inputs don't match those that were used to record the graph, the graph must be re-recorded. ``LaunchParams`` requires default constructor, ``operator==``, and a constructor that takes whatever values might change from launch to launch. E.g. if a sparse kernel is called with different X and Y vectors, the graph must be re-recorded if those vectors change.

- The call ``cugraph.begin_recording(args...)`` returns true if the graph needs to be re-recorded (or recorded for the first time), and false if the params match the existing graph. args... are used to construct the current LaunchParams instance.
  - If ``begin_recording(...)`` returned true, run kernels like normal, except policies are constructed by the wrapper, e.g. ``cugraph.range_policy<Tag>(begin, end)`` or ``cugraph.team_policy(nteams, nthreads)``.
  - After all kernels have been "executed" in a dry run, call ``cugraph.end_recording()`` to finalize the graph.
- Finally, call ``cugraph.launch()`` to actually execute the stored graph. This has about the same overhead as a single normal kernel launch, but runs all the kernels as they were recorded. It is also done on a separate stream and is not affected by CUDA_LAUNCH_BLOCKING, although a stream synchronize is done after the launch.

If CUDA is not enabled or is < 10.1, a dummy version of the graph wrapper is enabled with an identical interface, but begin_recording() always returns false (always re-record) and the policies returned are the real policies (so recording is not a dry run). This means that user code doesn't have to be duplicated depending on whether this is available.

CUDA graphs rely on Kokkos changes that are only in develop, so after the 3.0 release a small hotfix will be needed in the Trilinos snapshot to kokkos/core/src/impl/Kokkos_AnalyzePolicy.hpp.

Right now, this wrapper is just used in the apply of Gauss-Seidel and Cluster Gauss-Seidel. Multiple sweeps and iterations over the color sets (which used to involve many kernels) are collapsed into one graph. Whenever Ifpack2 calls apply() with the same vectors (as in multiple solver iterations) the graph is completely reused.

The spot checks on white/RIDE now have Cuda 10.1, so this feature is tested.

RIDE:
PASSED TESTS
cuda-10.1.105-Cuda_OpenMP-release build_time=597 run_time=401
cuda-10.1.105-Cuda_Serial-release build_time=599 run_time=503
cuda-9.2.88-Cuda_OpenMP-release build_time=525 run_time=396
cuda-9.2.88-Cuda_Serial-release build_time=525 run_time=498
gcc-6.4.0-OpenMP_Serial-release build_time=198 run_time=393
gcc-7.2.0-OpenMP-release build_time=129 run_time=129
gcc-7.2.0-OpenMP_Serial-release build_time=255 run_time=360
gcc-7.2.0-Serial-release build_time=148 run_time=227

Bowman:
PASSED TESTS
intel-16.4.258-Pthread-release build_time=743 run_time=1033
intel-16.4.258-Pthread_Serial-release build_time=1069 run_time=2009
intel-16.4.258-Serial-release build_time=708 run_time=934
intel-17.2.174-OpenMP-release build_time=875 run_time=581
intel-17.2.174-OpenMP_Serial-release build_time=1249 run_time=1465
intel-17.2.174-Pthread-release build_time=817 run_time=915
intel-17.2.174-Pthread_Serial-release build_time=1152 run_time=1789
intel-17.2.174-Serial-release build_time=788 run_time=891

### DIFF
--- a/src/common/KokkosKernels_ExecSpaceUtils.hpp
+++ b/src/common/KokkosKernels_ExecSpaceUtils.hpp
@@ -222,7 +222,7 @@ struct CudaGraphWrapper
   bool recording = false;
 };
 
-#if defined(KOKKOS_ENABLE_CUDA) && 10000 <= CUDA_VERSION
+#if defined(KOKKOS_ENABLE_CUDA) && 10010 <= CUDA_VERSION
 #define HAVE_CUDAGRAPHS
 
 //Dummy parallel call that makes compiler generate some declarations/initializations
--- a/src/common/KokkosKernels_ExecSpaceUtils.hpp
+++ b/src/common/KokkosKernels_ExecSpaceUtils.hpp
@@ -43,6 +43,7 @@
 
 #include "Kokkos_Core.hpp"
 #include "Kokkos_Atomic.hpp"
+#include "Kokkos_Concepts.hpp"
 
 #ifndef _KOKKOSKERNELSUTILSEXECSPACEUTILS_HPP
 #define _KOKKOSKERNELSUTILSEXECSPACEUTILS_HPP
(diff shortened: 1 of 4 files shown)


src/common/KokkosKernels_ExecSpaceUtils.hpp
```diff
@@ -136,7 +137,271 @@ inline int kk_get_suggested_team_size(const int vector_size, const ExecSpaceType
...
+  static constexpr P_t P = Kokkos::Experimental::WorkItemProperty::HintLightWeight;
+  
+  using WrappedTeamPolicy =
+    typename Kokkos::Experimental::Impl::PolicyPropertyAdaptor<
+    P_t, Kokkos::TeamPolicy<Kokkos::Cuda>>::policy_out_t;
+  using WrappedRangePolicy =
+    typename Kokkos::Experimental::Impl::PolicyPropertyAdaptor<
+    P_t, Kokkos::RangePolicy<Kokkos::Cuda>>::policy_out_t;
+
+  #define CUDA_CALL(x) \
```
mhoemmen: This is a popular macro to define, so this definition will quite likely collide with other code.  In general, it's unfortunate to expose CUDA API functions in a header file.

brian-kelley: @mhoemmen Good points. None of the native CUDA API calls actually depend on template parameters so I will put those in a separate .cpp.

brian-kelley:
I think I'm going to rework the "if(begin_recording)" thing. What I'm realizing is that
- it's easier for the caller to know if the graph needs to be re-recorded, than the graph wrapper. So the caller should just be able to re-record on demand.
- the graph can often be reused even if some inputs change. For example, in MTGS if the X and Y vectors change, the views (pointer+dimensions) could be stored in a struct and copied to a view on device. I think if all threads read a single global location it's fast because of cache + warp broadcast, but I would have to check if that's a significant overhead.

mhoemmen:
I thought David Hollman was working on an asynchronous CPU back-end for testing, that captures all kernels and only executes them in sequence on a fence.  That should be a good test case for a generic graph execution interface.

src/common/KokkosKernels_ExecSpaceUtils.hpp
lucbv: Actually this already exists in Kokkos, if you grep for `CUDA_SAFE_CALL` you will find an implementation of this.
Why not use the macro from Kokkos, this would avoid code duplication...

mhoemmen: @lucbv Kokkos itself really shouldn't be defining `CUDA_SAFE_CALL`, since it's not namespaced.  It's also a common macro that CUDA documentation seems to ask users to define.

Issue #598, opened 2020-02-06
complex_double misalignment still breaking SPGEMM


brian-kelley:
Memory alignment problems are still causing test failures for SGPEMM on CUDA: 
```
[ RUN      ] cuda.sparse_spgemm_kokkos_complex_double_int_int_TestExecSpace
terminate called after throwing an instance of 'std::runtime_error'
  what():  cudaFuncGetAttributes( &attr, cuda_parallel_launch_local_memory<DriverType>) error( cudaErrorMisalignedAddress): misaligned address /home/bmkelle/FixMultZeroRow-testing-1580941058/build/kokkos-install/include/Cuda/Kokkos_Cuda_KernelLaunch.hpp:439
Traceback functionality not available

Aborted
```

The above was from RIDE with GCC 7.2.0, CUDA 10.1, configured with:
```
../kokkos-kernels/cm_generate_makefile.bash --with-cuda --with-scalars=double,complex_double --compiler=/ascldap/users/bmkelle/FixMultZeroRow-testing-1580941058/kokkos/bin/nvcc_wrapper --debug --arch=Power8,Kepler37
```
But this affects release builds and CUDA 9.2 as well.

brian-kelley:
Note: replacing ``alignof(scalar_type)`` with ``sizeof(scalar_type)`` doesn't fix this.

mhoemmen:
Does the test code create an array of `std::complex<double>` and `reinterpret_cast` it to `Kokkos::complex<double>`?

brian-kelley:
@mhoemmen @ndellingwood and I think it's still in the shared memory logic (so we think no). This takes a raw pointer and breaks it into pieces by casting this pointer to different types and then incrementing it, so if that's wrong anywhere it would explain the problem.

brian-kelley:
Found it:
```
#define CHECKALIGN(ptr) \
  if((size_t) (ptr) % alignof(*(ptr))) \
  printf("Alignment of %s wrong (%s:%d)!\n", #ptr, __FILE__, __LINE__);

....
Alignment of vals wrong (/ascldap/users/bmkelle/KK_FixComplexAlignment/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_impl_kkmem.hpp:610)!
```

ndellingwood:
> Found it:

[...]

PR #599, opened 2020-02-06
Fixed SPGEMM for zero rows in several places

brian-kelley:
The behavior fixes were to define the the average and maximum nonzeros per row for a zero-row matrix to be zero, not INT_MIN.

Added SPGEMM unit test for zero row case that would have caught this.

This is a high priority issue affecting EMPIRE.

Bowman spot checks passed. RIDE spot checks failed due to #598 - this affects develop branch and isn't related to these changes.

However, #598 won't affect Trilinos because shared memory alignment logic goes through a different path in order to work with ensemble scalar types. @ndellingwood and I are working on a permanent fix for that.

### DIFF
--- a/unit_test/sparse/Test_Sparse_spgemm.hpp
+++ b/unit_test/sparse/Test_Sparse_spgemm.hpp
@@ -111,8 +111,6 @@ int run_spgemm(crsMat_t input_mat, crsMat_t input_mat2, KokkosSparse::SPGEMMAlgo
   bool equal = num_rows_2 == num_cols_1;
   if (!equal) return 1;
 
-
-
   lno_view_t row_mapC ("non_const_lnow_row", num_rows_1 + 1);
   lno_nnz_view_t  entriesC;
   scalar_view_t valuesC;
@@ -133,10 +131,8 @@ int run_spgemm(crsMat_t input_mat, crsMat_t input_mat2, KokkosSparse::SPGEMMAlgo
   );
 
   size_t c_nnz_size = kh.get_spgemm_handle()->get_c_nnz();
-  if (c_nnz_size){
-    entriesC = lno_nnz_view_t (Kokkos::ViewAllocateWithoutInitializing("entriesC"), c_nnz_size);
-    valuesC = scalar_view_t (Kokkos::ViewAllocateWithoutInitializing("valuesC"), c_nnz_size);
-  }
+  entriesC = lno_nnz_view_t (Kokkos::ViewAllocateWithoutInitializing("entriesC"), c_nnz_size);
+  valuesC = scalar_view_t (Kokkos::ViewAllocateWithoutInitializing("valuesC"), c_nnz_size);
   spgemm_numeric(
       &kh,
       num_rows_1,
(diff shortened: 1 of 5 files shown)


ndellingwood:
I think part of the need for this is that trilinos/Trilinos#5680 never had a matching PR back to kokkos-kernels repo, so those changes were clobbered by the promotion.

ndellingwood:
Thanks @brian-kelley for the due diligence adding a reproducer test

unit_test/sparse/Test_Sparse_spgemm.hpp
```diff
@@ -285,16 +280,17 @@ void test_spgemm(lno_t numRows, size_type nnz, lno_t bandwidth, lno_t row_size_v
...
   crsMat_t output_mat2;
   run_spgemm<crsMat_t, device>(input_mat, input_mat, SPGEMM_DEBUG, output_mat2);
 
-  SPGEMMAlgorithm algorithms [] = {SPGEMM_KK_MEMORY, SPGEMM_KK_SPEED, SPGEMM_KK_MEMSPEED, /*SPGEMM_CUSPARSE, */SPGEMM_MKL};
-
-  for (int ii = 0; ii < 4; ++ii){
+  std::vector<SPGEMMAlgorithm> algorithms = {SPGEMM_KK_MEMORY, SPGEMM_KK_SPEED, SPGEMM_KK_MEMSPEED};
 
-    SPGEMMAlgorithm spgemm_algorithm = algorithms[ii];
+#ifdef HAVE_KOKKOSKERNELS_MKL
```
srajama1: KOKKOSKERNELS_ENABLE_TPL_MKL is not needed ? I thought that is needed for some Trilinos path.

brian-kelley: @srajama1 Actually I need to check that. ``HAVE_KOKKOSKERNELS_MKL`` is the standard Tribits macro to use. Both that and ``KOKKOSKERNELS_ENABLE_TPL_MKL`` appear in the config.h.in but I'm not sure what the difference is.

brian-kelley: @jjwilke Is ``KOKKOSKERNELS_ENABLE_TPL_MKL`` the macro to use? Grepping for ``HAVE_KOKKOSKERNELS_MKL`` only shows up in this SPGEMM test and the config.h.in, so is it just a leftover thing from when we used the Tribits name conventions?

srajama1:
Also can we commit the test ?

unit_test/sparse/Test_Sparse_spgemm.hpp
```diff
@@ -456,6 +449,7 @@ void test_issue402()
 #define EXECUTE_TEST(SCALAR, ORDINAL, OFFSET, DEVICE) \
 TEST_F( TestCategory, sparse ## _ ## spgemm ## _ ## SCALAR ## _ ## ORDINAL ## _ ## OFFSET ## _ ## DEVICE ) { \
   test_spgemm<SCALAR,ORDINAL,OFFSET,DEVICE>(10000, 10000 * 30, 500, 10); \
+  test_spgemm<SCALAR,ORDINAL,OFFSET,DEVICE>(0, 0, 10, 10); \
```
brian-kelley: @srajama1 This is the test here, I just reused the same function.

PR #602, opened 2020-02-08
Permanent fix for scalar memory alignment for hashmaps

brian-kelley:
Explicitly align scalar_t pointers in hashmaps.
Account for the few extra bytes of padding (in practice, up to 4)
in every calculation of shared memory and pool chunk size.

@ndellingwood If you're OK with the changes I'll build and run Stokhos unit tests to make sure ensemble types work. Complex had 16-byte alignment for these tests (``KOKKOS_ENABLE_COMPLEX_ALIGN`` is on) so alignPtr is doing the correct thing here.

RIDE spot checks (double,complex_double)
PASSED TESTS
cuda-10.1.105-Cuda_OpenMP-release build_time=507 run_time=414
cuda-10.1.105-Cuda_Serial-release build_time=543 run_time=521
cuda-9.2.88-Cuda_OpenMP-release build_time=556 run_time=422
cuda-9.2.88-Cuda_Serial-release build_time=472 run_time=525
gcc-6.4.0-OpenMP_Serial-release build_time=227 run_time=386
gcc-7.2.0-OpenMP-release build_time=120 run_time=129
gcc-7.2.0-OpenMP_Serial-release build_time=191 run_time=362
gcc-7.2.0-Serial-release build_time=116 run_time=228
ibm-16.1.0-Serial-release build_time=490 run_time=395

Bowman spot checks (double, complex_double)
PASSED TESTS
intel-16.4.258-Pthread-release build_time=737 run_time=1030
intel-16.4.258-Pthread_Serial-release build_time=1092 run_time=2022
intel-16.4.258-Serial-release build_time=717 run_time=941
intel-17.2.174-OpenMP-release build_time=883 run_time=566
intel-17.2.174-OpenMP_Serial-release build_time=1261 run_time=1466
intel-17.2.174-Pthread-release build_time=805 run_time=928
intel-17.2.174-Pthread_Serial-release build_time=1142 run_time=1806
intel-17.2.174-Serial-release build_time=785 run_time=911

### DIFF
--- a/src/common/KokkosKernels_Utils.hpp
+++ b/src/common/KokkosKernels_Utils.hpp
@@ -1935,6 +1935,16 @@ struct array_sum_reduce
   }
 };
 
+template<typename InPtr, typename T>
+KOKKOS_INLINE_FUNCTION T* alignPtr(InPtr p)
+{
+  //ugly but computationally free and the "right" way to do this in C++
+  std::uintptr_t ptrVal = reinterpret_cast<std::uintptr_t>(p);
+  //ptrVal + (align - 1) lands inside the next valid aligned scalar_t,
+  //and the mask produces the start of that scalar_t.
+  return reinterpret_cast<T*>((ptrVal + alignof(T) - 1) & (~(alignof(T) - 1)));
+}
+
 }
 }
 
--- a/src/sparse/impl/KokkosSparse_spgemm_impl_kkmem.hpp
+++ b/src/sparse/impl/KokkosSparse_spgemm_impl_kkmem.hpp
@@ -168,8 +168,10 @@ struct KokkosSPGEMM
 
     while (team_cuckoo_key_size * 2 < tmp_team_cuckoo_key_size) team_cuckoo_key_size = team_cuckoo_key_size * 2;
     team_cuckoo_hash_func = team_cuckoo_key_size - 1;
-    team_shmem_key_size = ((shared_memory_size - sizeof(nnz_lno_t) * 4) / unit_memory);
-    thread_shmem_key_size = ((thread_memory - sizeof(nnz_lno_t) * 4) / unit_memory);
+    //How many extra bytes are needed to align a scalar_t after an array of nnz_lno_t, in the worst case?
+    constexpr size_t scalarAlignPad = (alignof(scalar_t) > alignof(nnz_lno_t)) ? (alignof(scalar_t) - alignof(nnz_lno_t)) : 0;
+    team_shmem_key_size = ((shared_memory_size - sizeof(nnz_lno_t) * 4 - scalarAlignPad) / unit_memory);
+    thread_shmem_key_size = ((thread_memory - sizeof(nnz_lno_t) * 4 - scalarAlignPad) / unit_memory);
     if (KOKKOSKERNELS_VERBOSE_){
       std::cout << "\t\tPortableNumericCHASH -- sizeof(scalar_t): " << sizeof(scalar_t) << "  sizeof(nnz_lno_t): " << sizeof(nnz_lno_t) << "  suggested_team_size: " << suggested_team_size << std::endl;
       std::cout << "\t\tPortableNumericCHASH -- thread_memory:" << thread_memory  << " unit_memory:" << unit_memory <<" initial key size:" << thread_shmem_key_size << std::endl;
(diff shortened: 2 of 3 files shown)


brian-kelley:
@ndellingwood Was SPGEMM the only reason why KOKKOS_ENABLE_COMPLEX_ALIGN was off in Trilinos builds? If it was, that cmake logic in Kokkos can be replaced with just having it on by default.

ndellingwood:
> Was SPGEMM the only reason why KOKKOS_ENABLE_COMPLEX_ALIGN was off in Trilinos builds? If it was, that cmake logic in Kokkos can be replaced with just having it on by default.

@brian-kelley No, there are issues throughout Trilinos with having Kokkos Complex alignas(16) depending on the arch and compiler (for example this didn't show up in any testing on Blake with intel/17, but did show up on whatever system the PR tester uses during complex testing with gcc/7.2.0).
As far as the segment of SPGEMM having the `KOKKOS_ENABLE_COMPLEX_ALIGN` guards, I added that to keep the shmem changes paired with the Kokkos setting and knowing that this would be disabled in Trilinos (essentially leaving it as it was before, which was compatible with with Trilinos).

ndellingwood:
I'll review a bit later, thanks for the work on this!

brian-kelley:
@ndellingwood At least, I can say that with ``KOKKOS_ENABLE_COMPLEX_ALIGN`` on and these changes, all Stokhos tests pass.

ndellingwood:
> @ndellingwood At least, I can say that with `KOKKOS_ENABLE_COMPLEX_ALIGN` on and these changes, all Stokhos tests pass.

@brian-kelley this is a great improvement, thanks! Would you be willing to apply the same treatment to the spgemm_speed algorithms as part of this PR, or would you prefer to keep this PR as is?


https://github.com/kokkos/kokkos-kernels/blob/1119007f03e01077f1d0dfe1b5c20360b5316e6f/src/sparse/impl/KokkosSparse_spgemm_impl_speed.hpp#L319-L356

ndellingwood:
@srajama1 

> This alignPtr business is needed only for hash values and not hash keys or ids ?

Yes, a heuristic determines the number of keys and ids etc, the "remainder" of memory is assigned to values without any check on memory alignment. My first pass was to make a minor adjustment of the number of keys and ids to allow for alignment of vals during the heuristic, a quick fix (or workaround) before realizing we had to support things like ensemble types (integration testing with Stokhos exposed this shortcoming); Brian's fix here is the right thing to do in any case.

> Someone has to teach me why is this a problem only in SpGEMM and not in other places. Don't we use shared memory anywhere ?

The reason this has to be done here is because in the numeric functors the used hashes, keys, begins, vals are assigned by pointer offset into the (team or thread) shared memory allocation. I can't speak for other routines in kokkos-kernels that use shmem and how this is handled (I'm not familiar with other routines' usage), but maybe they are allowing Kokkos to support this rather than manually assigning pointers, or they aren't assigning pointers to different offsets within an allocation?

brian-kelley:
@srajama1 @ndellingwood The reason why I'm not aligning ``nnz_lno_t`` also is because either

- For level 1 hashmaps, I'm assuming that the pointer returned by ``team_mem.get_shmem()`` is already aligned to ``nnz_lno_t``. Actually, I don't know if this is a valid assumption. @crtrott Does kokkos make any promises about alignment of a team's shared memory? 
- For level 2 hashmaps, the memory pool's storage is a view of ``nnz_lno_t`` so alignment isn't a problem.

brian-kelley:
@ndellingwood I'll actually update the PR. Since it's needed in multiple files I'll move "alignPtr" to Utils.

ndellingwood:
> Does kokkos make any promises about alignment of a team's shared memory?

It should, Christian spent some time fixing this in https://github.com/kokkos/kokkos/pull/1817

srajama1:
@ndellingwood @brian-kelley Thanks for the clarification ! It makes sense. This two pointer shifting business is old school, and I am partially to blame for teaching this trick to others !  Never knew it would cause this many issues on shared memory.

brian-kelley:
@srajama1 I updated this to also fix SPGEMM_SPEED and here's the new RIDE checks:

PASSED TESTS
cuda-10.1.105-Cuda_OpenMP-release build_time=524 run_time=417
cuda-10.1.105-Cuda_Serial-release build_time=538 run_time=522
cuda-9.2.88-Cuda_OpenMP-release build_time=527 run_time=420
cuda-9.2.88-Cuda_Serial-release build_time=485 run_time=525
gcc-6.4.0-OpenMP_Serial-release build_time=226 run_time=387
gcc-7.2.0-OpenMP-release build_time=132 run_time=128
gcc-7.2.0-OpenMP_Serial-release build_time=206 run_time=361
gcc-7.2.0-Serial-release build_time=114 run_time=227
ibm-16.1.0-Serial-release build_time=527 run_time=396

If you're still happy with this you can go ahead and merge it.

Issue #589, opened 2020-01-30
Add two BLAS/LAPACK calls needed by: Sptrsv supernode #552


e10harvey:
This issue is to address a subset of feedback from #552.

The following items should be addressed:
1. `KokkosSparse_sptrsv_aux.hpp` definitions of `{for,back}wardP_supernode` should be replaced with the correct KokkosBLAS / LAPACK / KokkosKernels routines. Will `KokkosKernels::Impl::permute_vector` or `KokkosKernels::Impl::permute_block_vector` work?
2. KokkosSparse_sptrsv_aux.hpp definition of `print_crsmat` should be replaced with a new routine (`print_2Dview`) in `KokkosKernels::Impl::print_2Dview` that uses `KokkosKernels::Impl::print_1Dview`
3. EDIT: We should have two specializations for `trmm`: ETI and TPL_CBLAS. Similarly we should have two specializations for `trtri`: ETI and TPL_LAPACKE (this is HostBlas). If I restructure this code correctly, the following behavior (in general) should be realized by adding KokkosBlas support for `trmm` and `trtri`:
- For `KokkosBlas3::trmm`:
  - If cblas is NOT available at compile time, the eti specialization (that uses Kokkos routines) will be selected.
  - if cblas is available at compile time, both eti and tpl specializations can be made available at compile time.
- Similarly for `KokkosBlas3::trtri`:
  - If lapacke is NOT available at compile time, the eti specialization will be selected. Otherwise, both the eti and tpl specialization can be made available.

EDIT: @srajama1: is no 3. correct?

@srajama1, please let me know if this is what you had in mind.

srajama1:
@e10harvey  : Yes for no. 1. I don't understand the question for permute_vector.

2. I don't see how 2D view can print CRS matrix. This is sparse. Nevertheless, we have CRS printing in many places, it would be good to have it an Util.

ndellingwood:
Is part of this issue to also provide the blas and lapack capabilities necessary to replace the cblas and lapacke calls in the supernodal sptrsv, or will that be discussed and addressed by a different issue?

srajama1:
@ndellingwood  This issue basically does that.

e10harvey:
> I don't understand the question for permute_vector.

This is mainly just a note to myself (which I should have made clear) -- I'm looking into which internal KokkosKernels routines we can use instead of these `{for,back}wardP_supernode` routines.

[...]

PR #605, opened 2020-02-17
sptrsv perf_test enhancements

e10harvey:
- perf_test/sparse: Improve test
  - Improve command line argument parsing for superlu perf test.
  - Conditionally set `scalar_t` based on configure value of KOKKOSKERNELS_SCALARS.

Related to #589.

Spot-check run in progress.


perf_test/sparse/KokkosSparse_sptrsv_superlu.cpp
```diff
@@ -796,8 +796,8 @@ int main(int argc, char **argv) {
   }
 
   {
-    using scalar_t = double;
-    //using scalar_t = Kokkos::complex<double>;
+    //using scalar_t = double;
```
srajama1: Can we pick this based on what scalars are defined ?

src/sparse/KokkosSparse_sptrsv_supernode.hpp
```diff
@@ -51,7 +51,9 @@
 #ifndef KOKKOSSPARSE_SPTRSV_SUPERNODE_HPP_
 #define KOKKOSSPARSE_SPTRSV_SUPERNODE_HPP_
 
-#if defined(KOKKOSKERNELS_ENABLE_TPL_LAPACKE) && defined(KOKKOSKERNELS_ENABLE_TPL_CBLAS)
+#if (defined(KOKKOSKERNELS_ENABLE_TPL_LAPACKE) && \
+    defined(KOKKOSKERNELS_ENABLE_TPL_CBLAS)) \
+    || defined(KOKKOSKERNELS_ENABLE_TPL_SUPERLU)
```
srajama1: Is this right to include cblas and lapacke when SuperLU is defined ?

e10harvey: I do not think it is the correct place, I think these includes should be hidden within `KokkosBlas` and I intend to do this while resolving #589. These small changes are just to fix compilation errors. `|| defined(KOKKOSKERNELS_ENABLE_TPL_SUPERLU)` makes the definition of `read_supernodal_graphL` available for `sptrsv_superlu`.

perf_test/sparse/KokkosSparse_sptrsv_superlu.cpp
e10harvey: I will modify this perf_test to exercise both the double and complex types.

src/sparse/KokkosSparse_sptrsv_supernode.hpp
srajama1: Thanks!

perf_test/sparse/KokkosSparse_sptrsv_superlu.cpp
srajama1: I think it should also be what is enabled in ETI at compile time.

e10harvey: Got it! Thanks!

src/sparse/KokkosSparse_sptrsv_supernode.hpp
iyamazaki: Thank you for looking at this. By the way, this file uses CBLAS/LAPACKE, but not SuperLU (in fact, it is also used with sptrsv_cholmod). Does CBLAS/LAPACKE gets enabled, if SuperLU get enabled?

e10harvey: No problem :) 

The problem here is that sptrsv_superlu needs the definition of `read_supernodal_graphL` -- otherwise, kokkos-kernels will not compile.

iyamazaki: So, you are trying to build it without CBLAS/LAPACKE enabled?

e10harvey: Thanks for the PR feedback. This is my mistake. It looks like I had a bug with my build makefile generation scripts where CBLAS and LAPACKE were not enabled but SUPERLU was.

I now see that the perf tests pull in `KokkosKernels_config.h` via `KokkosSparse_sptrsv.hpp->KokkosSparse_sptrsv_symbolic_spec.hpp`.

e10harvey: > So, you are trying to build it without CBLAS/LAPACKE enabled?

@iyamazaki: Yes, I was. This compilation issue arises when `--with-tpls=cholmod,superlu`.

srajama1: I am confused now. What is the goal here ? If you enable superlu, cholmod then cblas/lapacke are required. Your changes will get rid of cblas/lapacke dependency, but we are not there yet. Correct ?

perf_test/sparse/KokkosSparse_sptrsv_superlu.cpp
```diff
@@ -795,19 +795,33 @@ int main(int argc, char **argv) {
...
+  // If eti-type complex<double> is enabled at compile time, this is what
+  // the perf_test will use
+  #if defined(KOKKOSKERNELS_INST_COMPLEX_DOUBLE)
+    using scalar_t = Kokkos::complex<double>;
+    scalarTypeString = "(scalar_t = Kokkos::complex<double>)";
+  #else
+    // If eti-type double is enabled at compile time, this is what
+    // the perf_test will use
+    #if defined(KOKKOSKERNELS_INST_DOUBLE)
+      using scalar_t = double;
```
srajama1: Double should work fine, right ?

src/sparse/KokkosSparse_sptrsv_supernode.hpp
iyamazaki: Yea, I don't think we can compile the code without enabling CBLAS/LAPACKE, right now..

e10harvey: @srajama1: I was confused as well -- if we use `--with-tpls=cholmod,superlu`, one way to get things compiling is by adding `|| defined(KOKKOSKERNELS_ENABLE_TPL_SUPERLU)` here to fix a seemingly unrelated compilation error. If we use `--with-tpls=cholmod,superlu,cblas,blas,lapacke` everything works as expected without adding `|| defined(KOKKOSKERNELS_ENABLE_TPL_SUPERLU)` here.

> What is the goal here ?

The goal of this PR is to learn the kokkos-kernels PR guidelines and fix the build for the superlu and cholmod perf_tests and in the process.

> If you enable superlu, cholmod then cblas/lapacke are required.

OK, is the correct place to handle this dependency resolution in the new cmake build system (which announces whether a TPL is available) and the ETI specialization layer (which determines if a full specialization is available)?

> Your changes will get rid of cblas/lapacke dependency, but we are not there yet. Correct ?

I think that's right assuming I am starting to understand the ETI system. I plan on addressing this dependency in a separate PR -- I will update the description in #589 to see if we are on the same page.

These questions were really helpful, thank you.

perf_test/sparse/KokkosSparse_sptrsv_superlu.cpp
e10harvey: Yes, as long as `KOKKOSKERNELS_SCALARS='double'`.

e10harvey:
The spot-check completed on ride:
```bash
nohup ../scripts/test_all_sandia --kokkoskernels-path=$KOKKOSKERNELS_PATH --kokkos-path=$KOKKOS_PATH --arch="Power8,Pascal60" --spot-check &

**Comment by srajama1** [comment]
@e10harvey Can you start a spot check on a non-GPU machine as well ?

**Comment by e10harvey** [comment]
> @e10harvey Can you start a spot check on a non-GPU machine as well ?

@srajama1, no problem. I started another with kokkos@2.9.00 and kokkos-kernels@topic/sptrsv_build_patches on a sems machine. I won't have access to other non-GPU machines listed in `test_all_sandia` until the accounts are approved.

**Comment by srajama1** [comment]
Got it ! Thanks !

File: perf_test/sparse/KokkosSparse_sptrsv_superlu.cpp [hunk 4]
```diff
@@ -100,7 +102,7 @@ void print_factor_superlu(int n, SuperMatrix *L, SuperMatrix *U, int *perm_r, in
   }
   printf( "];\n" );
 
-#if 0
+#if defined(KOKKOSKERNELS_ENABLE_TPL_SUPERLU)
```
iyamazaki: I'm planning to remove these debugging functions in the next PR, but we may get compilation errors/warnings with %d in fprintf using complex scalar type?

e10harvey: I didn't encounter any when testing.

iyamazaki: These "spot-check" are not testing these codes, are they?

mhoemmen: @e10harvey Bold of you to assume that tests actually enable complex scalar types ;-)

e10harvey: > These "spot-check" are not testing these codes, are they?

Correct. The spot-check is not testing these diffs. 

> @e10harvey Bold of you to assume that tests actually enable complex scalar types ;-)

I ran the cholmod and superlu perf tests manually for both double and complex<double> :-)

We can either close or merge this PR -- it's really just an onboarding task for myself and the only real improvement is the super LU perf test argument processing.

srajama1:
@lucbv Thank you ! Merging it.

srajama1:
I merged this when I was trying to merge a PR by @lucbv by switching to the wrong window. I think it is still reasonable to merge this. 

@e10harvey @iyamazaki : Do you agree or should I revert it ? I would have guarded the fprintf with complex being defined, but I think you can submit another quick PR for it, right ?

e10harvey:
> @e10harvey @iyamazaki : Do you agree or should I revert it ? I would have guarded the fprintf with complex being defined, but I think you can submit another quick PR for it, right ?

I agree but I defer to @iyamazaki. @iyamazaki, shall I submit a PR for the fprintf guarding or are you still planning to remove these in your next PR?

iyamazaki:
Yes. I've already removed it. I am not quite done cleaning it up, but I am thinking to do a PR today after resolving conflicts. Thank you again for looking at it!!

PR #612, opened 2020-02-21
cm_test_all_sandia: bowman fix

ndellingwood:
swap gcc module for newer gcc


ndellingwood:
@lucbv this should fix the errors you were seeing earlier. There may be more changes to remove the use of update_lib script following kokkos/kokkos#2783 but that will be a separate PR pending testing.

lucbv:
Thanks @ndellingwood I'll try it on my next spot-check!

Issue #493, opened 2019-11-07
perf_test/sparse/spmv has its own MatrixMarket reader


brian-kelley:
Replace with IOUtils ``read_kokkos_crst_matrix()`` to get complete support for all the MatrixMarket formats, plus to remove duplicated code.

brian-kelley:
``SparseMatrix_MatrixMarket_read()`` at perf_test/sparse/spmv/matrix_market.hpp:87

Issue #492, opened 2019-11-05
test_common/KokkosKernels_MyCRSMatrix.hpp redefines existing CRS structures


brian-kelley:
In ``test_common/KokkosKernels_MyCRSMatrix.hpp``, there is a MyKokkosSparse namespace with redefinitions of StaticCrsGraph (which exists in KokkosContainers) and CrsMatrix (which exists in KokkosKernels/sparse). These classes only get used in performance tests. ``MyKokkosSparse::StaticCrsGraph`` keeps a num_cols member (unlike ``Kokkos::StaticCrsGraph``) but other than that, they seem to have identical features. Since it's confusing to have two definitions of the same thing (especially with the same name), I think we should remove KokkosKernels_MyCRSMatrix.hpp.

@srajama1 
@william76

PR #613, opened 2020-02-21
perf tests cleanup

### LINKED ISSUE #583: Can I run KokkosKernels spgemm with float or int32 type?
I am running under ./example/buildlib/perf_test.

brian-kelley:
Lots of cleanup/improvement to perf tests
- Remove MyCrsMatrix (instead, using KokkosSparse::CrsMatrix)
- Remove spmv/matrix_market.hpp (instead, using the IO utils)
- Make spgemm perf test not crash if ``--xyz`` is the last command line arg, but a value is expected after it (e.g. --openmp $numthreads). Instead, print error message and exit
- Make all perf tests use whatever scalar/ordinal/offset/layout types are enabled by ETI, rather than hardcoding double/int/int/left.
- Cleaned up lots of unused parameters in the various raw OpenMP spmv versions
- Added Ordinal/Offset (aka lno_t vs. size_type) distinction to the raw spmv versions

Also fixed an issue with Blas1 dot ETI: when scalar=float, the rank-1 version that returns a scalar should use double to accumulate the sum. This case was not being ETI'd, which broke the perf test build. This is now ETI'd correctly.

This PR resolves #583, #493, and #492.

I was able to build and run the unit and perf tests cleanly with serial, openmp and cuda enabled, and with double, float, complex_double and complex_float all enabled. I was also able to build without scalar=double, meaning float was used in all the tests.

RIDE spot check (with double and complex_double):
PASSED TESTS
cuda-10.1.105-Cuda_OpenMP-release build_time=578 run_time=418
cuda-10.1.105-Cuda_Serial-release build_time=515 run_time=521
cuda-9.2.88-Cuda_OpenMP-release build_time=504 run_time=425
cuda-9.2.88-Cuda_Serial-release build_time=553 run_time=526
gcc-6.4.0-OpenMP_Serial-release build_time=244 run_time=394
gcc-7.2.0-OpenMP-release build_time=127 run_time=128
gcc-7.2.0-OpenMP_Serial-release build_time=210 run_time=362
gcc-7.2.0-Serial-release build_time=119 run_time=227
ibm-16.1.0-Serial-release build_time=480 run_time=398

Bowman spot check (with double and complex_double):
PASSED TESTS
intel-16.4.258-Pthread-release build_time=733 run_time=1029
intel-16.4.258-Pthread_Serial-release build_time=1065 run_time=1993
intel-16.4.258-Serial-release build_time=715 run_time=949
intel-17.2.174-OpenMP-release build_time=881 run_time=569
intel-17.2.174-OpenMP_Serial-release build_time=1248 run_time=1468
intel-17.2.174-Pthread-release build_time=813 run_time=893
intel-17.2.174-Pthread_Serial-release build_time=1151 run_time=1784
intel-17.2.174-Serial-release build_time=791 run_time=876


brian-kelley:
I also made the SPGEMM perf test run with serial device, if it's enabled and neither CUDA nor OpenMP ran. Before, if you didn't give --openmp or --cuda, it would just print out a bunch of machine info and exit without doing a multiply, but it also didn't give a warning that nothing happened.

brian-kelley:
2 TODOs for this PR:
- decide how to handle the fact that the default for each ETI type is unconditionally enabled. This means that it's currently impossible for the user to get a perf test to use scalar=float, without changing source code (see #583)
- re run spot checks to make sure PRs since last week (and ideally, other pending PRs) didn't break this

ndellingwood:
@brian-kelley are any modifications related to your first TODO needed before merging this PR, or can that be handled in a separate PR (in which case I'll review for merge so you don't have to wait longer for this to go in)?

brian-kelley:
@ndellingwood No, the first TODO is not blocking this PR. It just means that users can't use float in the perf tests (yet) without changing source code themselves. This is fine as @jjwilke is already working on a solution in #620.

brian-kelley:
@ndellingwood Thanks for reviewing, I'll re-run the checks to make sure nothing here broke.

brian-kelley:
OK, rebasing this on develop built cleanly with Serial,Cuda,OpenMP on my machine so it's probably fine, but I'm still going to let the spot check frun.

brian-kelley:
Spot checks still passed, so I'm merging.

Issue #614, opened 2020-02-21
TPL support for SpMV


lucbv:
I am adding TPL support in SpMV, the main goal is to enable the use of cuSPARSE spmv functionalities but other TPLs such as MKL could be explored too?

So my plan is to proceed as follows:

- [x] Provide a first cut implementation that allows for SpMV using CUDA 10 and a subset of all possible type combinations. This will not be as general as we want but it will allow us to work out the kinks and will provide a first capability that can be tested.
- [x] Provide more generic available for multiple versions of cuda and hopefully all types
- [x] Add a macro similar to Kokkos' CUDA_SAFE_CALL for cusparse to help clean the code a bit
- [x] Implement logic to apply transpose operator to input matrix if needed
- [ ] Add utility functions to create cusparse `CrsMatrix` and `DenseVector` from kokkos-kernels objects
- [ ] Do some performance benchmarking and monitor the cost of the cusparse handle creation
- [ ] Update documentation as appropriate to mention then new possible algorithmic path

mhoemmen:
@lucbv Note that cuSPARSE uses `int` as the row offset type (type of the entries of `ptr`), while Tpetra uses whatever Kokkos says is the size type.  (The difference wasn't my idea -- Mike Heroux insisted on 64-bit row offsets.)  This means you'll either need to make a copy of `ptr` (use `Tpetra::Details::copyOffsets`, which can do bounds checking), or change Tpetra's storage format.

mhoemmen:
@lucbv If you want to make a copy of `ptr`, you could either store that in one of those kokkos-kernels "handle" things, or store it in Tpetra.  Either way, it could make sense to use (and maybe extend) the existing `Tpetra::{LocalOperator, LocalCrsMatrixOperator}` class hierarchy.  The point of `LocalCrsMatrixOperator` was to reduce build times and sizes by hiding the actual sparse mat-vec call from `Tpetra::CrsMatrix` compilation units, but it's also a point at which you know the structure of the local matrix.  It could be a place to store things like handles or copies of data structures.

mhoemmen:
@lucbv For an example of how to hide TPL header files from downstream code, see `CuSolverHandle` and `CuSolver` in `packages/tpetra/tsqr/src`.

lucbv:
[...]

PR #618, opened 2020-02-25
Spmv cusparse tpl

lucbv:
Adding cusparse backend for SpMV to allow users to benefit from latest improvement in nvidia code.
This will potentially provide better performance across different GPU architectures as it is complicated to tune the SpMV kernel for all graphic cards and compute versions.
The support will be provided for cuda 9.2 and 10 to begin with.

### DIFF
--- a/perf_test/sparse/KokkosSparse_spmv_struct_tuning.cpp
+++ b/perf_test/sparse/KokkosSparse_spmv_struct_tuning.cpp
@@ -480,7 +480,7 @@ int main(int argc, char **argv)
     }
 
     if(compare_cusparse) {
-#if defined(KOKKOSKERNELS_ENABLE_TPL_CUSPARSE) && defined(KOKKOS_ENABLE_CUDA_LAMBDA)
+#if defined(KOKKOSKERNELS_ENABLE_TPL_CUSPARSE)
       // The data needs to be reformatted for cusparse before launching the kernel.
       // Step one, extract raw data
       using graph_type = typename matrix_type::StaticCrsGraphType;
--- a/perf_test/sparse/KokkosSparse_spmv_struct_tuning.cpp
+++ b/perf_test/sparse/KokkosSparse_spmv_struct_tuning.cpp
@@ -92,6 +92,48 @@ void print_help() {
   printf("  --print-lp         : Print launch parameters to screen.\n");
 }
 
+template<typename graph_type>
+struct copy_crs_data {
+  using execution_space = typename graph_type::device_type::execution_space;
+  using cusparse_int_type = typename Kokkos::View<int*,
+						  typename graph_type::entries_type::array_layout,
+						  typename graph_type::device_type>;
+
+  // Dispatch tags
+  struct rowPtrTag{};
+  struct colIndTag{};
+
+  typename graph_type::row_map_type::const_type Arowptr;
+  typename graph_type::entries_type::const_type Acolind;
+  cusparse_int_type cusparse_Arowptr, cusparse_Acolind;
+
+  copy_crs_data(typename graph_type::row_map_type::const_type Arowptr_,
+		typename graph_type::entries_type::const_type Acolind_,
+		cusparse_int_type cusparse_Arowptr_,
+		cusparse_int_type cusparse_Acolind_) :
+    Arowptr(Arowptr_), Acolind(Acolind_),
+    cusparse_Arowptr(cusparse_Arowptr_),
+    cusparse_Acolind(cusparse_Acolind_) {};
+
+  void doCopy() {
+    Kokkos::RangePolicy<execution_space, rowPtrTag> rowPtrPolicy(0, Arowptr.extent(0));
+    Kokkos::parallel_for("copy rowPtr to cusparse", rowPtrPolicy, *this);
+
+    Kokkos::RangePolicy<execution_space, colIndTag> colIndPolicy(0, Acolind.extent(0));
+    Kokkos::parallel_for("copy colInd to cusparse", colIndPolicy, *this);
+  }
+
+  KOKKOS_INLINE_FUNCTION
+  void operator() (const rowPtrTag&, const size_t idx) const {
+    cusparse_Arowptr(idx) = int(Arowptr(idx));
+  }
+
+  KOKKOS_INLINE_FUNCTION
+  void operator() (const colIndTag&, const size_t idx) const {
+    cusparse_Acolind(idx) = int(Acolind(idx));
+  }
+};
+
 template<class AMatrix,
          class XVector,
          class YVector>
@@ -101,7 +143,7 @@ void struct_matvec(const int stencil_type,
                    const AMatrix& A,
                    const XVector& x,
                    typename YVector::const_value_type& beta,
-                   const YVector& y,
+                   YVector& y,
                    int team_size_int,
                    int vector_length,
                    int64_t rows_per_thread_int,
@@ -183,7 +225,7 @@ void matvec(typename YVector::const_value_type& alpha,
             const AMatrix& A,
             const XVector& x,
             typename YVector::const_value_type& beta,
-            const YVector& y,
+            YVector& y,
             int team_size,
             int vector_length,
             int64_t rows_per_thread,
@@ -444,20 +486,15 @@ int main(int argc, char **argv)
       using graph_type = typename matrix_type::StaticCrsGraphType;
       using cusparse_int_type = typename Kokkos::View<int*,
 						      typename graph_type::entries_type::array_layout,
-						      typename matrix_type::device_type>;
+						      typename graph_type::device_type>;
 
       typename graph_type::row_map_type::const_type     Arowptr = A.graph.row_map;
       typename graph_type::entries_type::const_type     Acolind = A.graph.entries;
       typename matrix_type::values_type::non_const_type Avals   = A.values;
-      cusparse_int_type Arowptr_cusparse, Acolind_cusparse;
-      Arowptr_cusparse = cusparse_int_type("Arowptr", Arowptr.extent(0));
-      Acolind_cusparse = cusparse_int_type("Acolind", Acolind.extent(0));
-      Kokkos::parallel_for(Arowptr.extent(0), KOKKOS_LAMBDA(const size_t idx) {
-      	  Arowptr_cusparse[idx] = Arowptr[idx];
-      	});
-      Kokkos::parallel_for(Acolind.extent(0), KOKKOS_LAMBDA(const size_t idx) {
-      	  Acolind_cusparse[idx] = Acolind[idx];
-      	});
+      cusparse_int_type Arowptr_cusparse("Arowptr", Arowptr.extent(0));
+      cusparse_int_type Acolind_cusparse("Acolind", Acolind.extent(0));
+      copy_crs_data<graph_type> myCopyFunctor(Arowptr, Acolind, Arowptr_cusparse, Acolind_cusparse);
+      myCopyFunctor.doCopy();
 
       int*    rows = reinterpret_cast<int*>(Arowptr_cusparse.data());
       int*    cols = reinterpret_cast<int*>(Acolind_cusparse.data());
--- a/src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
+++ b/src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
@@ -30,7 +30,7 @@
 // CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
 // EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
 // PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
-// PROFIS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
 // LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
@@ -74,11 +74,12 @@ namespace Impl {
     cudaDataType myCudaDataType;
     if(std::is_same<value_type, float>::value)  {myCudaDataType = CUDA_R_32F;}
     if(std::is_same<value_type, double>::value) {myCudaDataType = CUDA_R_64F;}
+    std::string label = "KokkosSparse::spmv[TPL_CUSPARSE," + Kokkos::ArithTraits<value_type>::name() + "]: ";
 
     /* initialize cusparse library */
     status = cusparseCreate(&handle);
     if (status != CUSPARSE_STATUS_SUCCESS) {
-      throw std::runtime_error("KokkosSparse::spmv[TPL_CUSPARSE,double]: cusparse was not initialized correctly");
+      throw std::runtime_error(label + "cusparse was not initialized correctly");
     }
 
     /* create matrix */
@@ -92,18 +93,18 @@ namespace Impl {
 			       CUSPARSE_INDEX_BASE_ZERO,
 			       myCudaDataType);
     if (status != CUSPARSE_STATUS_SUCCESS) {
-      throw std::runtime_error("KokkosSparse::spmv[TPL_CUSPARSE,double]: cusparse matrix was not created correctly");
+      throw std::runtime_error(label + "cusparse matrix was not created correctly");
     }
 
     /* create lhs and rhs */
     cusparseDnVecDescr_t vecX, vecY;
     status = cusparseCreateDnVec(&vecX, x.extent_int(0), const_cast<value_type*>(x.data()), myCudaDataType);
     if (status != CUSPARSE_STATUS_SUCCESS) {
-      throw std::runtime_error("KokkosSparse::spmv[TPL_CUSPARSE,double]: cusparse vecX was not created correctly");
+      throw std::runtime_error(label + "cusparse vecX was not created correctly");
     }
     status = cusparseCreateDnVec(&vecY, y.extent_int(0), const_cast<value_type*>(y.data()), myCudaDataType);
     if (status != CUSPARSE_STATUS_SUCCESS) {
-      throw std::runtime_error("KokkosSparse::spmv[TPL_CUSPARSE,double]: cusparse vecY was not created correctly");
+      throw std::runtime_error(label + "cusparse vecY was not created correctly");
     }
 
     size_t bufferSize = 0;
@@ -112,11 +113,11 @@ namespace Impl {
 				     &alpha, A_cusparse, vecX, &beta, vecY, myCudaDataType,
 				     CUSPARSE_CSRMV_ALG1, &bufferSize);
     if (status != CUSPARSE_STATUS_SUCCESS) {
-      throw std::runtime_error("KokkosSparse::spmv[TPL_CUSPARSE,double]: cusparse bufferSize computation failed");
+      throw std::runtime_error(label + "cusparse bufferSize computation failed");
     }
     cuError = cudaMalloc(&dBuffer, bufferSize);
     if (cuError != cudaSuccess) {
-      throw std::runtime_error("KokkosSparse::spmv[TPL_CUSPARSE,double]: cuda buffer allocation failed");
+      throw std::runtime_error(label + "cuda buffer allocation failed");
     }
 
     /* perform SpMV */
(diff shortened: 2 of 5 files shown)


crtrott:
Nathan: I think we are at >= 9.0 not 9.2

ndellingwood:
Edited my comment, thanks for clarification.

lucbv:
@ndellingwood 
currently this PR supports building without failing x)
It's really an early effort and I am currently trying to write a unit-test that exercises the tpl code path when enabled.
Once I have that I should be able to actually put in the cusparse interface, at that point I will be able to look at how the implementation details in the cusparse API guide the support possible from within kokkos-kernels.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
```diff
@@ -70,6 +71,84 @@ struct SPMV<class AT, class AO, class AD, class AM, class AS,
...
+
+  cudaError_t      cuError;
+  cusparseStatus_t status;
+  cusparseHandle_t handle=0;
+  // cusparseMatDescr_t descr=0;
+
+  /* initialize cusparse library */
+  status = cusparseCreate(&handle);
+  if (status != CUSPARSE_STATUS_SUCCESS) {
+    throw("KokkosSparse::spmv[TPL_CUSPARSE,double]: cusparse was not initialized correctly");
```
mhoemmen: 1. Please don't throw `const char[]`.  Use a subclass of `std::exception`.
2. Kokkos should have a function for reporting errors; please use that instead of throwing.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
```diff
@@ -70,6 +71,84 @@ struct SPMV<class AT, class AO, class AD, class AM, class AS,
...
+  }
+
+  /* create matrix */
+  cusparseSpMatDescr_t A_cusparse;
+  status = cusparseCreateCsr(&A_cusparse, A.numRows(), A.numCols(), A.nnz(),
```
mhoemmen: You're exposing cuSPARSE functions in the header file :(

lucbv: I am actually thinking of creating a small utility function that takes in a `Kokkos::CrsMatrix` and returns a `cusparseSpMatDescr_t` so part of this code will disappear down the road...
I will also look into other TPL specialization we have in kokkos-kernels and see how things are implemented. Depending on that I might move things around accordingly.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
lucbv: Once things are actually working (right now the spmv does not give the right answer), I'll look into the proper way of throwing exception or reporting errors. We must have some mechanism for that in other tpl interfaces, maybe in cublas_gemm interface for instance?

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
mhoemmen: @lucbv Check out how I hid the TPL header file includes for cuSOLVER in `tpetra/tsqr/src`.

lucbv:
Alright, this is now passing `cm_test_all_sandia --spot-check-tpls`, I do not think that the build on bowman was useful to check that the tpl work correctly but at least we know it does not break the regular code path.

### White
cuda-10.1.105-Cuda_Serial-release build_time=288 run_time=270
cuda-9.2.88-Cuda_OpenMP-release build_time=325 run_time=233
gcc-7.2.0-OpenMP-release build_time=87 run_time=52
gcc-7.2.0-Serial-release build_time=76 run_time=98
gcc-7.4.0-OpenMP-release build_time=88 run_time=40

### Bowman
intel-16.4.258-Pthread-release build_time=523 run_time=296
intel-16.4.258-Pthread_Serial-release build_time=682 run_time=530
intel-16.4.258-Serial-release build_time=520 run_time=250
intel-17.2.174-OpenMP-release build_time=683 run_time=91
intel-17.2.174-OpenMP_Serial-release build_time=848 run_time=296
intel-17.2.174-Pthread-release build_time=584 run_time=207
intel-17.2.174-Pthread_Serial-release build_time=775 run_time=416
intel-17.2.174-Serial-release build_time=550 run_time=197

### Kokkos-dev
PASSED TESTS
clang-4.0.1-Pthread_Serial-hwloc-release build_time=129 run_time=147
clang-4.0.1-Pthread_Serial-release build_time=123 run_time=181
clang-7.0.1-Cuda_OpenMP-hwloc-release build_time=349 run_time=281
clang-7.0.1-Cuda_OpenMP-release build_time=339 run_time=278
cuda-9.2-Cuda_OpenMP-release build_time=496 run_time=391
gcc-5.3.0-OpenMP-hwloc-release build_time=95 run_time=48
gcc-5.3.0-OpenMP-release build_time=96 run_time=47
gcc-7.3.0-Serial-hwloc-release build_time=80 run_time=84
gcc-7.3.0-Serial-release build_time=89 run_time=83
intel-17.0.1-OpenMP-hwloc-release build_time=158 run_time=33
intel-17.0.1-OpenMP-release build_time=158 run_time=33

lucbv:
@srajama1 @ndellingwood any thoughts on this PR?
I am keen on keeping it short and merging it since it actually works as intended now.
However I would like to work in a second PR on creating some utilities, such as a Kokkos to Cusparse matrix converter and also discuss with @mhoemmen some more on hiding the headers.

ndellingwood:
@lucbv I haven't looked carefully, but wanted to check that using the deprecated cusparse matrix-vector interface (`*csrmv`) is supported by all the versions of Cuda we need to support?

lucbv:
@ndellingwood the tpl is guarded to make sure it only gets enabled for cuda 9 and up.
The [*csrmv](https://docs.nvidia.com/cuda/archive/9.0/cusparse/index.html#cusparse-lt-t-gt-csrmv) routines are defined correctly in cuda 9 so there is no issue with it.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
lucbv: @mhoemmen 
I looked into `Kokkos_Cuda_Error.hpp` and the main blocker to use `CUDA_SAFE_CALL` is that it uses the type `cudaError` when the cusparse calls are retunring `cusparseStatus_t` instead. At this point I added better throws that use `std::runtime_exception`.
I will add to the todo list the creation of a `CUSPARSE_SAFE_CALL` macro that replicates mechanism demonstrated in Kokkos but for cusparse return type.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
lucbv: @mhoemmen as discussed I will outline a strategy to reduce the size of the macros in the `KokkosSparse_spmv_tpl_spec_decl.hpp` and hide the header in an implementation file.

lucbv:
@mhoemmen if you don't see any inconvenience with this I will update issue #618 to explain how I plan to address further your comments in a subsequent PR that will refactor this code. But I would like for the current version to be merged.

I will re-run the spot-check since I did one more push for the exception handling.

mhoemmen:
@lucbv wrote:
> if you don't see any inconvenience with this I will update issue #618 to explain how I plan to address further your comments in a subsequent PR that will refactor this code. But I would like for the current version to be merged.

I'm cool with that -- thanks! :-D

lucbv:
@mhoemmen @ndellingwood @srajama1 
code is ready and passes spot-check, I have documented some next steps in #618 to clean things up a bit but it does not need to happen now for this to be merged. Let me know if you have strong objection to something in the PR or if we can merge it.

### White --spot-check
PASSED TESTS
cuda-10.1.105-Cuda_OpenMP-release build_time=329 run_time=250
cuda-10.1.105-Cuda_Serial-release build_time=297 run_time=298
cuda-9.2.88-Cuda_OpenMP-release build_time=293 run_time=328
cuda-9.2.88-Cuda_Serial-release build_time=297 run_time=376
gcc-6.4.0-OpenMP_Serial-release build_time=126 run_time=165
gcc-7.2.0-OpenMP-release build_time=84 run_time=57
gcc-7.2.0-OpenMP_Serial-release build_time=117 run_time=163
gcc-7.2.0-Serial-release build_time=77 run_time=102
ibm-16.1.0-Serial-release build_time=403 run_time=114

### White --spot-check-tpls
PASSED TESTS
cuda-10.1.105-Cuda_Serial-release build_time=302 run_time=269
cuda-9.2.88-Cuda_OpenMP-release build_time=313 run_time=305
gcc-7.2.0-OpenMP-release build_time=89 run_time=52
gcc-7.2.0-Serial-release build_time=74 run_time=97
gcc-7.4.0-OpenMP-release build_time=90 run_time=40

### Kokkos-dev --spot-check
PASSED TESTS
clang-4.0.1-Pthread_Serial-hwloc-release build_time=126 run_time=159
clang-4.0.1-Pthread_Serial-release build_time=129 run_time=217
clang-7.0.1-Cuda_OpenMP-hwloc-release build_time=345 run_time=288
clang-7.0.1-Cuda_OpenMP-release build_time=348 run_time=321
cuda-9.2-Cuda_OpenMP-release build_time=401 run_time=414
gcc-5.3.0-OpenMP-hwloc-release build_time=98 run_time=57
gcc-5.3.0-OpenMP-release build_time=95 run_time=55
gcc-7.3.0-Serial-hwloc-release build_time=81 run_time=92
gcc-7.3.0-Serial-release build_time=81 run_time=91
intel-17.0.1-OpenMP-hwloc-release build_time=168 run_time=56
intel-17.0.1-OpenMP-release build_time=167 run_time=54

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
```diff
@@ -30,7 +30,7 @@
 // CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
 // EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
 // PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
-// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// PROFIS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
```
ndellingwood: typo here, maybe accidental bump on the keyboard?

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
```diff
@@ -44,9 +44,214 @@
...
+		      const XVector& x,					\
+		      const coefficient_type& beta,			\
+		      const YVector& y) {				\
+      Kokkos::Profiling::pushRegion("KokkosSparse::spmv[TPL_CUSPARSE,double]");	\
+      spmv_cusparse(mode, alpha, A, x, beta, y);			\
+      Kokkos::Profiling::popRegion();					\
+    }									\
+  };
+
+  KOKKOSSPARSE_SPMV_CUSPARSE(double, int, Kokkos::LayoutLeft,  true)
```
ndellingwood: Is complex support to be added later?

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
```diff
@@ -44,9 +44,214 @@
...
+									\
+    using coefficient_type = typename YVector::non_const_value_type;	\
+									\
+    static void spmv (const char mode[],				\
+		      const coefficient_type& alpha,			\
+		      const AMatrix& A,					\
+		      const XVector& x,					\
+		      const coefficient_type& beta,			\
+		      const YVector& y) {				\
+      Kokkos::Profiling::pushRegion("KokkosSparse::spmv[TPL_CUSPARSE,double]");	\
```
ndellingwood: Should the double in the label be SCALAR?

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
lucbv: Potentially but I am a bit worried about the alignment of `cuComplex` and that of `Kokkos::complex` for that reason I would like to defer this to allow careful testing when that feature goes in.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
lucbv: Yeah, likely an inadvertent keystroke... easy to fix though

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
ndellingwood: I'll need to look a bit more carefully at the ETI TPL layer you added, but if complex is not enabled then does this mean kokkos-kernels spmv will use the existing fallback implementation? If not, that'll need to be addressed or else this will cause issues in Trilinos if anyone enables cusparse.

lucbv: If you look at `KokkosSparse_spmv_tpl_spec_avail.hpp` you will see that the tpl is not made available for complex type so it will fall back to kokkos-kernel implementation of spmv in that case.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
ndellingwood: Thanks for clarifying, was pretty sure that seemed to be the case.

ndellingwood:
@lucbv approved, shall I go ahead and merge?

lucbv:
@ndellingwood I just pushed minor changes to address your comments, let me know what you think of them

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
mhoemmen: @ndellingwood Let me second Luc's concern about alignment of complex.  I remember y'all added an option to control alignment of `Kokkos::complex`; we could use that to decide whether to enable complex support.

perf_test/sparse/KokkosSparse_spmv_struct_tuning.cpp
```diff
@@ -92,6 +92,48 @@ void print_help() {
...
+  copy_crs_data(typename graph_type::row_map_type::const_type Arowptr_,
+		typename graph_type::entries_type::const_type Acolind_,
+		cusparse_int_type cusparse_Arowptr_,
+		cusparse_int_type cusparse_Acolind_) :
+    Arowptr(Arowptr_), Acolind(Acolind_),
+    cusparse_Arowptr(cusparse_Arowptr_),
+    cusparse_Acolind(cusparse_Acolind_) {};
+
+  void doCopy() {
+    Kokkos::RangePolicy<execution_space, rowPtrTag> rowPtrPolicy(0, Arowptr.extent(0));
```
mhoemmen: See `Tpetra::Details::copyOffsets`.  That would be a useful kokkos-kernels utility.

perf_test/sparse/KokkosSparse_spmv_struct_tuning.cpp
```diff
@@ -101,7 +143,7 @@ void struct_matvec(const int stencil_type,
                    const AMatrix& A,
                    const XVector& x,
                    typename YVector::const_value_type& beta,
-                   const YVector& y,
+                   YVector& y,
```
mhoemmen: Why not const?  It's a Kokkos::View.

perf_test/sparse/KokkosSparse_spmv_struct_tuning.cpp
```diff
@@ -438,26 +480,21 @@ int main(int argc, char **argv)
...
-      Arowptr_cusparse = cusparse_int_type("Arowptr", Arowptr.extent(0));
-      Acolind_cusparse = cusparse_int_type("Acolind", Acolind.extent(0));
-      Kokkos::parallel_for(Arowptr.extent(0), KOKKOS_LAMBDA(const size_t idx) {
-      	  Arowptr_cusparse[idx] = Arowptr[idx];
-      	});
-      Kokkos::parallel_for(Acolind.extent(0), KOKKOS_LAMBDA(const size_t idx) {
-      	  Acolind_cusparse[idx] = Acolind[idx];
-      	});
+      cusparse_int_type Arowptr_cusparse("Arowptr", Arowptr.extent(0));
+      cusparse_int_type Acolind_cusparse("Acolind", Acolind.extent(0));
```
mhoemmen: You shouldn't need to allocate column indices unless they don't have type `int`.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
```diff
@@ -44,9 +44,214 @@
...
+		     typename YVector::non_const_value_type const & alpha,
+		     const AMatrix& A,
+		     const XVector& x,
+		     typename YVector::non_const_value_type const & beta,
+		     const YVector& y) {
+    using offset_type  = typename AMatrix::non_const_size_type;
+    // using ordinal_type = typename AMatrix::non_const_ordinal_type;
+    using value_type   = typename AMatrix::non_const_value_type;
+
+#if defined(CUSPARSE_VERSION) && (10300 <= CUSPARSE_VERSION)
```
mhoemmen: I thought it was CUDA 10.1 that added the new interface.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
```diff
@@ -44,9 +44,214 @@
...
+    if (status != CUSPARSE_STATUS_SUCCESS) {
+      throw std::runtime_error("KokkosSparse::spmv[TPL_CUSPARSE,double]: cusparse handle was not desctroyed correctly");
+    }
+
+#else
+
+    /* Initialize cusparse */
+    cusparseStatus_t cusparseStatus;
+    cusparseHandle_t cusparseHandle=0;
+    cusparseStatus = cusparseCreate(&cusparseHandle);
```
mhoemmen: We should time this -- it could be expensive to create this handle.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
```diff
@@ -44,9 +44,214 @@
...
+    if(cusparseStatus != CUSPARSE_STATUS_SUCCESS) {
+      throw std::runtime_error("KokkosSparse::spmv[TPL_CUSPARSE,double]: error setting the matrix index base");
+    }
+
+    /* perform the actual SpMV operation */
+    if(std::is_same<int, offset_type>::value) {
+      if (std::is_same<value_type,float>::value) {
+	cusparseStatus = cusparseScsrmv(cusparseHandle, CUSPARSE_OPERATION_NON_TRANSPOSE,
+					A.numRows(), A.numCols(), A.nnz(),
+					(const float *) &alpha, descrA,
```
mhoemmen: Prefer `reinterpret_cast` to C-style casts.

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
ndellingwood: I think it's a newer "patch" update to 10.1 that added the new interface, 10.1.patch, but is not included in all versions of 10.1 (in particular not the version we have a sems module for nightly testing)

ndellingwood:
@mhoemmen do you request your comments be addressed prior to merge or as a follow up PR?

mhoemmen:
@ndellingwood I'm good with this -- Luc and I talked earlier this week.  Sorry to delay you!

ndellingwood:
> I assume we have an issue that tracks

@srajama1 yes, we missed adding this earlier, cross-reference issue #614

ndellingwood:
> I'm good with this

@mhoemmen just wanted to check, always valuable feedback and didn't want to bypass anything requiring immediate address that was overlooked


> It would be nice to see how expensive handle creation is

@srajama1 good call this would be useful, I'm going to go ahead with the merge (I think this is safe based on your follow up comment), this will get it through a round of nightlies (we have tpls in the nightly tests now :)

src/impl/tpls/KokkosSparse_spmv_tpl_spec_decl.hpp
lucbv: It's a bit complicated, but the bottom line is that cuda and cusparse do not necessarily have the same version numbering.

| Release number | CUDA_VERSION | CUSPARSE_VERSION
|---|---|---|
| 10.1.168 | 10100 | undef |
| 10.1.243 | 10100 | 10300 |

Note that in release 10.1.168, cusparse does define the following macros:
```
#define CUSPARSE_VER_MAJOR 10
#define CUSPARSE_VER_MINOR 2
#define CUSPARSE_VER_PATCH 0
#define CUSPARSE_VER_BUILD 0
```
where as in 10.1.243 it defines
```
#define CUSPARSE_VER_MAJOR 10
#define CUSPARSE_VER_MINOR 3
#define CUSPARSE_VER_PATCH 0
#define CUSPARSE_VER_BUILD 243
#define CUSPARSE_VERSION (CUSPARSE_VER_MAJOR * 1000 + \
                          CUSPARSE_VER_MINOR *  100 + \
                          CUSPARSE_VER_PATCH)
```

perf_test/sparse/KokkosSparse_spmv_struct_tuning.cpp
lucbv: I could bring back the const, the intent is to show that this view could be modified, although its data is really what is being modified.

perf_test/sparse/KokkosSparse_spmv_struct_tuning.cpp
lucbv: That's correct but since this is a test I did not want to write too much logic around the type of the ordinal.
It would make the test a bit faster but this part is not timed and we already know that the cusparse interface is changing in future versions. Ideally this test should be rewritten down the road for the new interface that will make this a bit cleaner...

PR #626, opened 2020-03-02
Fixed CMake for using add_subdirectory(kokkos-kernels) 

kuberry:
Changed cmake/kokkoskernels_tribits.cmake to produce KokkosKernels*.cmake files.

Changed CMakeLists.txt so that for an installation with parent, i.e.
add_subdirectory(kokkos-kernels), so that headers were installed in
include/kokkos-kernels consistent with where all of eti files are
installed.

Once installed, the install interface for src/CMakeLists.txt didn't
differentiate between building with a parent or not, so I changed it
to include/kokkos-kernels rather than include for installs with a
parent.

Now, KokkosKernels can be built, installed, built/linked against
through using:

>> add_subdirectory(kokkos-kernels)
>> target_link_libraries(compadre PUBLIC kokkoskernels)


cmake/kokkoskernels_tribits.cmake
```diff
@@ -10,7 +10,7 @@ MESSAGE(STATUS "The project name is: ${PROJECT_NAME}")
 MACRO(KOKKOSKERNELS_PACKAGE_POSTPROCESS)
   IF (KOKKOSKERNELS_HAS_TRILINOS)
     TRIBITS_PACKAGE_POSTPROCESS()
-  ELSEIF(NOT KOKKOSKERNELS_HAS_PARENT)
+  ELSE()
```
jjwilke: These changes are not the correct behavior. Even if you are building Kokkos Kernels in-tree, you would still want to install its package file.

src/CMakeLists.txt
```diff
@@ -128,7 +128,11 @@ ELSE()
   FOREACH(DIR ${KK_INCLUDE_DIRS})
     TARGET_INCLUDE_DIRECTORIES(kokkoskernels PUBLIC $<BUILD_INTERFACE:${DIR}>)
   ENDFOREACH()
-  TARGET_INCLUDE_DIRECTORIES(kokkoskernels PUBLIC $<INSTALL_INTERFACE:include>)
+  IF(KOKKOSKERNELS_HAS_PARENT)
+    TARGET_INCLUDE_DIRECTORIES(kokkoskernels PUBLIC $<INSTALL_INTERFACE:include/kokkos-kernels>)
```
jjwilke: You would again still want to the add this include directory for installation. The INSTALL_INTERFACE generator expression only adds this to the install, not the in-tree build.

jjwilke:
For the time being, you can just link to target `kokkoskernels`. What's missing is I apparently never added the alias `Kokkos::kokkoskernels`

jjwilke:
Unless I'm misunderstanding the problems, we should probably cancel this in favor of PR https://github.com/kokkos/kokkos-kernels/pull/627.

All of our tests are "in-tree" builds so I would be surprised if there were something fundamentally wrong with the configure other than missing the ALIAS name in the directions.

jjwilke:
Basically keep this, but a couple extra fixes over in PR:
https://github.com/kokkos/kokkos-kernels/pull/628

src/CMakeLists.txt
jjwilke: I tried to limit the number of HAS_PARENT if stmts in the new PR.

CMakeLists.txt
```diff
@@ -42,7 +42,8 @@ IF (KOKKOSKERNELS_HAS_TRILINOS)
  SET(KOKKOSKERNELS_HEADER_DIR ${TRILINOS_INCDIR})
  SET(KOKKOS_ENABLE_CUDA_UVM ${Kokkos_ENABLE_CUDA_UVM})
 ELSEIF(KOKKOSKERNELS_HAS_PARENT)
- SET(KOKKOSKERNELS_HEADER_DIR "${CMAKE_INSTALL_INCLUDEDIR}/kokkos-kernels")
+ SET(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/kokkos-kernels")
```
jjwilke: Don't overwrite CMAKE_INSTALL_INCLUDEDIR. Keep this as the "canonical" value

PR #640, opened 2020-03-05
Make sure cuda lambda support is enabled for cusparse tests

jjwilke:


ndellingwood:
I think the right thing is to guard the tests themselves regarding lambda support, rather than erroring out at configure when cusparse is enabled and if cuda lambda support is not enabled. 
I reproduced the error and submitting a PR with a guard on the test that fixes the error.

jjwilke:
Yeah, we can definitely do that. The error is currently only if the test is also enabled. The concern would be silently turning off tests that people were potentially expecting to run. Do we want to warn when certain tests are inactive, but could/should be?

ndellingwood:
> Do we want to warn when certain tests are inactive, but could/should be?

@jjwilke yes, that would be great. And I agree with your point about silently turning off tests. For the case that brought this issue up, Luc is replacing the lambda with a functor, but in general when we add tests would we be able to "register" them with the build system as using lambdas so the warning or configure error gets thrown if configured improperly? This would be awesome.

lucbv:
@ndellingwood @jjwilke 
I think the right thing to do here is to add appropriate guards/warning in the file: `perf_test/sparse/CMakeLists.txt` at line 32.
This is much more focused compared to enabling lambda's for all sparse tests. Also I believe the PR might not be necessary after PR #618 is merged.

PR #656, opened 2020-03-12
cm_test_all_sandia: temporarily remove -Werror from armclang on mayer

ndellingwood:
addresses -Werror with newer armclang not seen in previous versions

### DIFF
--- a/scripts/cm_test_all_sandia
+++ b/scripts/cm_test_all_sandia
@@ -543,9 +543,11 @@ elif [ "$MACHINE" = "mayer" ]; then
   BASE_MODULE_LIST="cmake/3.14.5,<COMPILER_NAME>/<COMPILER_VERSION>"
 #  ARM_MODULE_LIST="cmake/3.12.2,<COMPILER_NAME>/<COMPILER_VERSION>"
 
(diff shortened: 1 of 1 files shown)


ndellingwood:
Mayer recently updated the armclang version from 19.2 to 20.0. This change resulted in a new `-Werror` breaking nightly builds:

```
<https://jenkins-son.sandia.gov/job/KokkosKernels_Mayer_arm20_0/ws/kokkos-kernels/src/blas/impl/KokkosBlas3_gemm_impl.hpp>:403:63: error: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Werror,-Wpass-failed=transform-warning]
  Kokkos::parallel_for(Kokkos::TeamThreadRange(team,blockA0), [&] (const int i) {
```

This gets triggered by a `pragma omp simd` a few lines after 403, which the compiler should be able to handle.

This did not occur with previous armclang compilers, for now I think we should temporarily remove `-Werror` from the arm builds so that testing can resume and prevent further errors from slipping through.

PR #660, opened 2020-03-13
Kokkos-Kernels: Add fence for dot

MicheldeMessieres:
@ndellingwood @brian-kelley We have a similar fence in this file for the other Impl::Dot so I think we want this one too. This will resolve the Tpetra test TpetraCore_idot to work with CUDA_LAUNCH_BLOCKING off.

### DIFF
--- a/src/blas/KokkosBlas1_dot.hpp
+++ b/src/blas/KokkosBlas1_dot.hpp
@@ -244,6 +244,7 @@ dot (const RV& R, const XMV& X, const YMV& Y,
   YMV_Internal Y_internal = Y;
 
   Impl::Dot<RV_Internal, XMV_Internal, YMV_Internal>::dot(R_internal, X_internal, Y_internal);
+  Kokkos::fence();
 }
 }
 


mhoemmen:
@MicheldeMessieres wrote:
> This will resolve the Tpetra test TpetraCore_idot to work with CUDA_LAUNCH_BLOCKING off.

Wait, woah, what's this?  Shouldn't `KokkosBlas::dot` be synchronous?  Why does a synchronous reduction need a fence?

@iyamazaki @jennloe FYI

src/blas/KokkosBlas1_dot.hpp
```diff
@@ -244,6 +244,7 @@ dot (const RV& R, const XMV& X, const YMV& Y,
   YMV_Internal Y_internal = Y;
 
   Impl::Dot<RV_Internal, XMV_Internal, YMV_Internal>::dot(R_internal, X_internal, Y_internal);
+  Kokkos::fence();
```
mhoemmen: Oh, I think I get it.  The problem is that if you use the version of `Kokkos::parallel_reduce` that takes the output as a View, rather than as a reference to host memory, then the reduction is asynchronous.

I would call this overkill, though.  Most dot products are for a single vector, in which case the version of `parallel_reduce` that takes a reference to host memory is the right overload to call.  If I call the overload of `dot` that takes an output View, then I, the caller, should be responsible for fencing.

FYI @trilinos/tpetra @kddevin

kyungjoo-kim:
@brian-kelley I just want to make it clear. By the nature of reduction, the dot product can be synchronous but it does not have to synchronize a device. Later, we can just use a single stream for this dot product and if we put a fence there it would synchronize all.

MicheldeMessieres:
@brian-kelley I'll resolve it in Tpetra. Thanks.

mhoemmen:
Can't I just fence the execution space instance?

src/blas/KokkosBlas1_dot.hpp
crtrott: I agree with mark here ..

srajama1:
The question is even more broader, are we going to fence for all BLAS / LAPACK calls, some of them ? I agree with @kyungjoo-kim  it is safer to say they are asynchronous and let the user fence (device or execution space instance). However, I am happy to listen to a counter argument.

Issue #645, opened 2020-03-05
team_dot_complex_double fails on clang+cuda due to misalignment


brian-kelley:
On kokkos-dev2, with Clang 8 and CUDA 10.0 (a spot check build), team_dot_complex_double fails because of ``cudaErrorMisalignedAddress``. Since this passes in all other CUDA builds (even with ``16 == alignof(Kokkos::complex<double>)``), it seems likely this is just a compiler bug.

I can tell this error is caused by the ``KokkosBlas::Experimental::dot`` device function, since cudaGetLastError() reports the misaligned access right after the functor launches in the test Test_Blas1_team_dot.hpp, in ``impl_test_team_dot()``:
```
    Kokkos::fence();
    checkCuda();  // returns cudaSuccess
    Kokkos::parallel_for( "KokkosBlas::Test::TeamDot", policy, KOKKOS_LAMBDA ( const team_member &teamMember ) { 
       const int teamId = teamMember.league_rank();
       CHECKALIGN(&d_r(teamId));
       auto lhs = Kokkos::subview(a,Kokkos::make_pair(teamId*team_data_siz,(teamId < M-1)?(teamId+1)*team_data_siz:N));
       auto rhs = Kokkos::subview(b,Kokkos::make_pair(teamId*team_data_siz,(teamId < M-1)?(teamId+1)*team_data_siz:N));
       CHECKALIGN(lhs.data());
       CHECKALIGN(rhs.data());
       d_r(teamId) = KokkosBlas::Experimental::dot(teamMember, lhs, rhs);
    } );
    Kokkos::fence();
    checkCuda(); // returns cudaErrorMisalignedAddress
```
If it's not a compiler bug, I have no idea where the actual bug could be. This dot implementation is definitely calling ``TeamDot::team_dot`` in KokkosBlas1_team_dot_spec.hpp, and this takes actual views as input, not raw pointers. I checked that the X and Y passed to this functor are correctly aligned too.

Disabling ``KOKKOS_ENABLE_COMPLEX_ALIGN`` does make the test pass.

PR #661, opened 2020-03-16
Fix624: lots of distance-2 changes, fix Utils transpose

### LINKED ISSUE #624: Distance-2 coloring does not respect distance-1 neighbors having same color
Distance-2 coloring only checks for conflicts between pairs of vertices, separated by a path of length exactly 2. If there is no path of length 2 but a path of length 1 (u and v are immediate neighbors, and neither u nor v have themselves as neighbors) then the conflict will not be detected, and so the coloring will be incorrect.

If every row has the diagonal entry (every vertex has itself as a neighbor) then the algorithms are correct, but we don't enforce that or document it as a precondition. Consider it a bug.

brian-kelley:
Fix #624.

Refactor the D2 interface: deprecate old function ``graph_compute_distance2_color()``, and add 3 new functions ``graph_color_distance2()``, ``bipartite_color_rows()``, and ``bipartite_color_columns()``. The first new function does distance-2 coloring on an undirected/symmetric graph (so no d-1 or d-2 conflicts). The other two do one-sided coloring of a bipartite graph, so d-1 conflicts are not considered, and the input graph doesn't have to be symmetric (but it can be). If it isn't symmetric, the transpose is computed internally so the user doesn't have to.

All the new functions and the old one (if deprecated enabled) are tested.

Delete the matrix-squared version of D2 (it was temporary anyway).

Fix a bug in ``kk_generate_sparse_matrix`` where a row could have a negative number of entries, e.g. ``rowmap(k+1) < rowmap(k)``.


brian-kelley:
Testing:
kokkos-dev2:
PASSED TESTS
clang-8.0-Pthread_Serial-release build_time=87 run_time=368
cuda-10.1-Cuda_OpenMP-release build_time=274 run_time=305
cuda-9.2-Cuda_Serial-release build_time=263 run_time=462
gcc-7.3.0-OpenMP-release build_time=58 run_time=120
gcc-7.3.0-Pthread-release build_time=58 run_time=189
gcc-8.3.0-Serial-release build_time=60 run_time=204
gcc-9.1-OpenMP-release build_time=80 run_time=285
gcc-9.1-Serial-release build_time=74 run_time=206
intel-18.0.5-OpenMP-release build_time=172 run_time=150
FAILED TESTS
clang-8.0-Cuda_OpenMP-release (test failed)
(the only failure is due to #645)

RIDE:
PASSED TESTS
cuda-10.1.105-Cuda_OpenMP-release build_time=530 run_time=404
cuda-10.1.105-Cuda_Serial-release build_time=558 run_time=513
cuda-9.2.88-Cuda_OpenMP-release build_time=542 run_time=552
cuda-9.2.88-Cuda_Serial-release build_time=546 run_time=665
gcc-6.4.0-OpenMP_Serial-release build_time=189 run_time=388
gcc-7.2.0-OpenMP-release build_time=132 run_time=129
gcc-7.2.0-OpenMP_Serial-release build_time=211 run_time=359
gcc-7.2.0-Serial-release build_time=115 run_time=226
ibm-16.1.0-Serial-release build_time=509 run_time=396

brian-kelley:
Also, I ran kokkos-dev2 spot check with ``-DKokkos_ENABLE_DEPRECATED_CODE=ON``. Most builds tested cleanly (including the deprecated D2 test), although a couple of builds failed, e.g. an error in Kokkos with GCC+pthreads. None of these errors were related to this PR, though.

brian-kelley:
Update: OK, our transpose is fine, I just misunderstood what the temporary view meant. It turns out the only bug was in ``kk_generate_sparse_matrix``, but before I discovered that I misidentified the bug as being in transpose.

perf_test/graph/KokkosGraph_color_d2.cpp
```diff
@@ -356,32 +354,28 @@ void run_experiment(crsGraph_t crsGraph, int num_cols, Parameters params)
...
         case 1:
-            kh.create_distance2_graph_coloring_handle(COLORING_D2_MATRIX_SQUARED);
-            label_algorithm = "COLORING_D2_MATRIX_SQUARED";
-            break;
-        case 2:
             kh.create_distance2_graph_coloring_handle(COLORING_D2_SERIAL);
             label_algorithm = "COLORING_D2_SERIAL";
             break;
-        case 3:
+        case 2:
```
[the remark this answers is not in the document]
brian-kelley: Good idea, I'll do that.

brian-kelley:
@lucbv I did the things you suggested and the transpose changes are completely reverted. Re-running tests.

srajama1:
Man .. GitHub ate some of my comments .. will do them again ..

perf_test/graph/KokkosGraph_color_d2.cpp
```diff
@@ -210,34 +207,29 @@ int parse_inputs(KokkosKernels::Experiment::Parameters &params, int argc, char *
         else if(0 == strcasecmp(argv[i], "--algorithm"))
         {
             ++i;
-            if(0 == strcasecmp(argv[i], "COLORING_D2_MATRIX_SQUARED"))
+            if(0 == strcasecmp(argv[i], "COLORING_D2_SERIAL"))
             {
                 params.algorithm             = 1;
```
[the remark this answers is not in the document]
brian-kelley: @srajama1 Luc had the same concern, so I just made a new params struct called D2Parameters that lives in this file and only has the fields that D2 perf test uses. It also uses the correct enum types, instead of just int for everything.

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
```diff
@@ -565,64 +559,73 @@ class GraphColorDistance2
...
       color_view_type colors;
-      const_lno_row_view_type Crowmap;  //C <-> V graph (row c is the vertices incident to c)
-      const_lno_nnz_view_t    Ccolinds;
-      const nnz_lno_type numVerts;
+      rowmap_t Crowmap;  //C <-> V graph (row c is the vertices incident to c)
+      entries_t Ccolinds;
+      const lno_t numVerts;
     };
 
     template<int batch>
```
srajama1: Does this have to be template parameter ?

perf_test/graph/KokkosGraph_color_d2.cpp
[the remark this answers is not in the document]
brian-kelley: We should consider doing that for all the perf tests, honestly. I think test_common/TestParameters just started out for SPMV and SPGEMM but eventually has a ton of fields for different algorithms

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
```diff
@@ -1481,44 +1272,44 @@ class GraphColorDistance2
                 // * The Distance-1 code used the knowledge of the degree of the vertex to cap the number of iterations
                 //   but in distance-2 we'd need the total vertices at distance-2 which we don't easily have aprioi.
                 //   This could be as big as all the vertices in the graph if diameter(G)=2...
-                for(color_type offset = 0; !foundColor && offset < nv; offset += VB_D2_COLORING_FORBIDDEN_SIZE)
+                for(color_type offset = 1; offset <= nv; offset += VB_D2_COLORING_FORBIDDEN_SIZE)
```
srajama1: We don't stop immediately after finding colors now ? May be I am missing a break somewhere.

unit_test/graph/Test_Graph_graph_color_distance2.hpp
```diff
@@ -61,231 +61,291 @@ using namespace KokkosGraph::Experimental;
...
+    { \
+      test_bipartite<SCALAR, ORDINAL, OFFSET, DEVICE>(2000, 4000, 3000 * 20, 800, 10, true); \
+      test_bipartite<SCALAR, ORDINAL, OFFSET, DEVICE>(4000, 2000, 3000 * 20, 800, 10, true); \
     } \
-    TEST_F(TestCategory, graph##_##graph_color_d2_filtering##_##SCALAR##_##ORDINAL##_##OFFSET##_##DEVICE) \
-    {                                                                                           \
-      test_d2_filtering<SCALAR, ORDINAL, OFFSET, DEVICE>(5000, 7000, 5000 * 10, 200, 10); \
+    TEST_F(TestCategory, graph##_##graph_color_bipartite_col##_##SCALAR##_##ORDINAL##_##OFFSET##_##DEVICE) \
+    { \
+      test_bipartite<SCALAR, ORDINAL, OFFSET, DEVICE>(2000, 4000, 3000 * 20, 800, 10, false); \
```
srajama1: What happens if we give a square symmetric matrix here ?

src/graph/KokkosGraph_Distance2Color.hpp
```diff
@@ -36,34 +36,229 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
-// Questions? Contact William McLendon (wcmclen@sandia.gov)
+// Questions? Contact Siva Rajamanickam (srajama@sandia.gov)
```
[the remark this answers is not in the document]
brian-kelley: Will do

src/graph/KokkosGraph_Distance2Color.hpp
```diff
@@ -36,34 +36,229 @@
...
+ * This problem is equivalent to grouping the matrix rows into a minimal number of sets, so that
+ * within each set, the intersection of any two rows' entries is empty.
+ *
+ * Distance-1 conflicts (where r1 and c are neighbors) are not avoided,
+ * since columns are not colored. In general, there is no particular relationship between a row and column
+ * that happen to have the same index.
  *
- * If the graph is symmetric, give the same value for col_map and row_map,
- * and for row_entries and col_entries.
+ * However, if the input graph is symmetric and has diagonal entries in every row, then rows and columns
```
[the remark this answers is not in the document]
brian-kelley: No, I will add that. I'm already generating symmetric graphs for distance2 so it will be easy.

src/graph/KokkosGraph_Distance2ColorHandle.hpp
```diff
@@ -53,22 +54,16 @@
...
-    COLORING_D2_DEFAULT,             // Distance-2 Graph Coloring default algorithm
-    COLORING_D2_SERIAL,              // Distance-2 Graph Coloring (SERIAL)
-    COLORING_D2_MATRIX_SQUARED,      // Distance-2 Graph Coloring using Matrix Squared + D1 Coloring
-    COLORING_D2_SPGEMM,              // - Same as COLORING_D2_MATRIX_SQUARED
-    COLORING_D2,                     // Distance-2 Graph Coloring
-    COLORING_D2_VB,                  // Distance-2 Graph Coloring Vertex Based
-    COLORING_D2_VB_BIT,              // Distance-2 Graph Coloring Vertex Based BIT
-    COLORING_D2_VB_BIT_EF,           // Distance-2 Graph Coloring Vertex Based BIT + Edge Filtering
-    COLORING_D2_NB_BIT               // Distance-2 Graph Coloring Net Based BIT
+    COLORING_D2_DEFAULT,     // Distance-2 Graph Coloring default algorithm
```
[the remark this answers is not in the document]
brian-kelley: The default is SERIAL for Serial device and NB_BIT for all parallel devices.

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
brian-kelley: @srajama1 Down below when we actually color the vertex, the functor just returns immediately. That way we don't need nested breaks.

unit_test/graph/Test_Graph_graph_color_distance2.hpp
brian-kelley: @srajama1 It will still work, but it will compute the explicit transpose internally which is just inefficient. I will add a symmetric bipartite (i.e. no difference for coloring rows vs. columns) test.

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
brian-kelley: @srajama1 I was worried about efficiency with doing it at runtime, although I really need to measure it. Batch is always a power of 2 so it could still be represented by bitshift. I do see a code size concern with compile time.

src/common/KokkosKernels_SparseUtils.hpp
```diff
@@ -315,40 +315,25 @@ template <typename in_row_view_t,
           typename out_nnz_view_t,
           typename out_scalar_view_t,
           typename tempwork_row_view_t,
-          typename MyExecSpace>
+          typename MyExecSpace,
+          bool transpose_values>
```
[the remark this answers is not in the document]
brian-kelley: @srajama1 I think it makes sense for that to be compile-time since it saves a branch in the innermost loop, and there's already different user-facing functions for graph vs. matrix.

I'm trying to figure out whether there's really a problem with the old transpose, or whether something in D2 is just making it look like there's a problem. With the old transpose impl, the test_bipartite with columns ends up with a tranpose rowmap that's all zero (obviously wrong). When I rewrote transpose, I got a plausible rowmap there and everything passed (including sparse unit tests, which calls transpose many times).

However, I wrote a unit test for transpose itself to try to prove there's a bug, but they all passed with the old implementation. So I'm still trying to figure out what's really going on with valgrind.

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
```diff
@@ -73,113 +75,103 @@ namespace Impl {
...
-                        HandleType*             handle)
+    GraphColorDistance2(lno_t       nv_,
+                        lno_t       nc_,
+                        rowmap_t    row_map,
+                        entries_t   entries,
+                        rowmap_t    t_row_map,
+                        entries_t   t_entries,
+                        HandleType* handle)
         : nv(nv_)
         , nr(nv_)
```
lucbv: Is this always true or is this an assumption?
If the latter it should be tested... for a prolongator matrix the number of columns will be small compared to the number of rows.

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
```diff
@@ -73,113 +75,103 @@ namespace Impl {
...
         , t_xadj(t_row_map)
         , t_adj(t_entries)
         , gc_handle(handle)
         , _chunkSize(handle->get_vb_chunk_size())
         , _max_num_iterations(handle->get_max_number_of_iterations())
-        , _use_color_set(0)
         , _ticToc(handle->get_verbose())
         , _verbose(handle->get_verbose())
     {
+      //Internal logic check: distance-2 coloring (non-bipartite) requires a square graph
```
lucbv: Can we add an extra check such as `nr < nv`?
It seems to be implicitly expected but I am not sure that we test for it, that is not true do we want to throw or do do we need to think about a different logic in the coloring?

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
brian-kelley: @lucbv Rows and vertices are always equivalent (I should just get rid of 'nv'). You couldn't do distance-2 coloring on a nonsquare prolongator matrix, but you could do bipartite row or column coloring.

brian-kelley:
Wow, just figured out the bug in kk_transpose_graph():

```
 449 inline void kk_transpose_graph(
 450     typename in_nnz_view_t::non_const_value_type num_rows,
 451     typename in_nnz_view_t::non_const_value_type num_cols,
 452     in_row_view_t xadj,
 453     in_nnz_view_t adj,
 454     out_row_view_t t_xadj, //pre-allocated -- initialized with 0
 455     out_nnz_view_t t_adj,  //pre-allocated -- no need for initialize
 456     int vector_size = -1,
 457     int suggested_team_size = -1,
 458     typename in_nnz_view_t::non_const_value_type team_work_chunk_size = -1,
 459     bool use_dynamic_scheduling = true
 460     ){
 461   std::cout << "Hello from kk_transpose_graph().\n";
 462
 463   //allocate some memory for work for row pointers
 464   tempwork_row_view_t tmp_row_view(Kokkos::ViewAllocateWithoutInitializing("tmp_row_view"), num_co     ls + 1);
 465
 466   in_nnz_view_t tmp1;
 467   out_nnz_view_t tmp2;
 468
 469   //create the functor for tranpose.
 470   typedef TransposeMatrix <
 471       in_row_view_t, in_nnz_view_t, in_nnz_view_t,
 472       out_row_view_t, out_nnz_view_t, out_nnz_view_t,
 473       tempwork_row_view_t, MyExecSpace>  TransposeFunctor_t;
 474
 475   TransposeFunctor_t tm ( num_rows, num_cols, xadj, adj, tmp1,
 476                           t_xadj, t_adj, tmp2,
 477                           tmp_row_view,
 478                           false,
 479                           team_work_chunk_size);
```

The functor gets constructed with ``team_work_chunk_size=-1`` by default, before team_work_chunk_size gets set to the team size if it's -1. And a grep says that I'm the first person to actually use the graph transpose (matrix didn't have this problem).

perf_test/graph/KokkosGraph_color_d2.cpp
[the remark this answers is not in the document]
srajama1: Can you put a GitHub issue on this ?

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
brian-kelley: @lucbv I just deleted nv completely, since it was always equal to nr.

srajama1:
@brian-kelley : Can you add performance numbers for reproducibility and documentation ?

We spoke about few other things, you can take care of it and resolve the conversations

Looks like this is ready to go ..

perf_test/graph/KokkosGraph_color_d2.cpp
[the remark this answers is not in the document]
brian-kelley: #662

src/graph/impl/KokkosGraph_Distance2Color_impl.hpp
brian-kelley: I'm leaving batch as a template parameter for now but I can revisit this after the exawind code freeze.

brian-kelley:
Re-running the tests. I added the symmetric bipartite test, but I will put the transpose unit test in a different PR.

brian-kelley:
Testing is good:
@srajama1 or @lucbv can you give final approval?
PASSED TESTS
clang-8.0-Pthread_Serial-release build_time=86 run_time=371
cuda-10.1-Cuda_OpenMP-release build_time=272 run_time=323
cuda-9.2-Cuda_Serial-release build_time=262 run_time=431
gcc-7.3.0-OpenMP-release build_time=58 run_time=119
gcc-7.3.0-Pthread-release build_time=57 run_time=184
gcc-8.3.0-Serial-release build_time=60 run_time=176
gcc-9.1-OpenMP-release build_time=71 run_time=116
gcc-9.1-Serial-release build_time=66 run_time=173
intel-18.0.5-OpenMP-release build_time=164 run_time=121
FAILED TESTS
clang-8.0-Cuda_OpenMP-release (test failed) (because of #645)

PASSED TESTS
cuda-10.1.105-Cuda_OpenMP-release build_time=519 run_time=408
cuda-10.1.105-Cuda_Serial-release build_time=563 run_time=515
cuda-9.2.88-Cuda_OpenMP-release build_time=553 run_time=535
cuda-9.2.88-Cuda_Serial-release build_time=518 run_time=660
gcc-6.4.0-OpenMP_Serial-release build_time=206 run_time=388
gcc-7.2.0-OpenMP-release build_time=128 run_time=130
gcc-7.2.0-OpenMP_Serial-release build_time=199 run_time=361
gcc-7.2.0-Serial-release build_time=116 run_time=225
ibm-16.1.0-Serial-release build_time=491 run_time=395

PR #663, opened 2020-03-19
Test transpose util, add CRS sorting util

brian-kelley:
- Delete kk_transpose_graph, now there's just transpose_graph (transpose_graph was just a wrapper for kk_transpose_graph, but we don't need there to be 2 functions with slightly different interfaces)
- Body of transpose_graph now basically identical to transpose_matrix, except with a dummy values view
- Unit tests for transpose_matrix and transpose_graph
- Delete "TransposeMatrix2" which was not used anywhere and had a comment "TODO delete me"
- New util functions: sort_crs_matrix and sort_crs_graph: sort CRS rows. Needed that to write the transpose test, but it's a very useful thing to have. The implementation came from SPADD, and uses the SerialRadix/TeamBitonic sorts.
- Unit tests for sort_crs_matrix and sort_crs_graph.

kokkos-dev2:
PASSED TESTS
clang-8.0-Pthread_Serial-release build_time=68 run_time=374
cuda-10.1-Cuda_OpenMP-release build_time=240 run_time=322
cuda-9.2-Cuda_Serial-release build_time=241 run_time=432
gcc-7.3.0-OpenMP-release build_time=55 run_time=119
gcc-7.3.0-Pthread-release build_time=52 run_time=178
gcc-8.3.0-Serial-release build_time=52 run_time=177
gcc-9.1-OpenMP-release build_time=63 run_time=114
gcc-9.1-Serial-release build_time=58 run_time=167
intel-18.0.5-OpenMP-release build_time=165 run_time=120
FAILED TESTS
clang-8.0-Cuda_OpenMP-release (test failed, just the TeamDot + complex)

RIDE
PASSED TESTS
cuda-10.1.105-Cuda_OpenMP-release build_time=482 run_time=413
cuda-10.1.105-Cuda_Serial-release build_time=492 run_time=514
cuda-9.2.88-Cuda_OpenMP-release build_time=503 run_time=560
cuda-9.2.88-Cuda_Serial-release build_time=448 run_time=661
gcc-6.4.0-OpenMP_Serial-release build_time=158 run_time=387
gcc-7.2.0-OpenMP-release build_time=100 run_time=129
gcc-7.2.0-OpenMP_Serial-release build_time=155 run_time=361
gcc-7.2.0-Serial-release build_time=100 run_time=226
ibm-16.1.0-Serial-release build_time=497 run_time=396


src/common/KokkosKernels_SparseUtils.hpp
```diff
@@ -830,6 +728,177 @@ inline size_t kk_is_d1_coloring_valid(
...
+    KokkosKernels::Impl::SerialRadixSort2<lno_t, unsigned_lno_t, scalar_t>(
+        (unsigned_lno_t*) entries.data() + rowStart,
+        (unsigned_lno_t*) entriesAux.data() + rowStart,
+        values.data() + rowStart,
+        valuesAux.data() + rowStart, rowNum);
+  }
+
+  KOKKOS_INLINE_FUNCTION void operator()(const team_mem t) const
+  {
+    size_type i = t.league_rank();
```
lucbv: Are we assuming here that the `league_size` is equal to the number of rows to sort? That seems a bit risky, if the user does not realize that they might end up with a half-sorted `CrsMatrix`... Would it be OK to create the team policy inside the functor to make sure if verify this hypothesis.

src/common/KokkosKernels_SparseUtils.hpp
```diff
@@ -830,6 +728,177 @@ inline size_t kk_is_d1_coloring_valid(
...
+
+  rowmap_t rowmap;
+  entries_t entries;
+  entries_t entriesAux;
+};
+
+// Sort a CRS matrix: within each row, sort entries ascending by column.
+// At the same time, permute the values.
+template<typename execution_space, typename rowmap_t, typename entries_t, typename values_t>
+void sort_crs_matrix(const rowmap_t& rowmap, const entries_t& entries, const values_t& values)
```
lucbv: OK, with this it is safe to ignore my comment above about the `league_size`.

brian-kelley:
@srajama1 Good point, bitonic is not a stable sort so with duplicate keys, the order of values is not defined. If I were to compress rows though, they would always end up the same.

srajama1:
So the result will be not deterministic on GPUs and deterministic on CPUs, right ? Yes, compression takes all the problems away. Do we do the compression ? I don't seem to find it.

lucbv:
We do not do compression at the moment.
If we do compression we need to be careful about the state of the CrsMatrix though, does it compute internal values such as numCols and nnz correctly with over allocated entries and values?

brian-kelley:
@srajama1 It's deterministic on CPU and GPU in the sense that the result will be the same every time you run it. Bitonic has no data races or atomics. But if you show me a row with a column appearing multiple times, I wouldn't be able to tell you what order they'd end up in without running it.

I didn't do compression here, but I agree that would be useful. @lucbv, you can't have overallocated entries/values in CrsMatrix because it does use the view length to determine nonzeros. This isn't really a loss since I think it would be hard to compress in-place anyway, but it's very easy to compress into separately allocated views.

lucbv:
@brian-kelley if you look at line 738 in CrsMatrix, I think that numCols is actually computed at construction and cached since it does not correspond to the extent of the views stored in the CrsGraph or CrsMatrix, we would need to reset it correctly.

brian-kelley:
@lucbv Sure, but numCols doesn't correspond to the length of rowmap, entries or values. If we compress a matrix though, the number of columns stays the same since it still represents the same matrix.

lucbv:
correct I was typing to quickly I agree with you that the number of columns should not change if we only do compression, since it actually does not remove columns

brian-kelley:
OK, I added a sort-and-merge utility, moved stuff into SparseUtils, and the transpose test has many rows with zero entries. Running full tests again to be safe.

brian-kelley:
BTW, the compression keeps graph structure, so if an entry has value 0, it won't get deleted. I think that's the right choice because some things like Gauss-Seidel deal with structurally symmetric, numerically nonsymmetric matrices, so you'd want to maintain structural symmetriy.

srajama1:
> BTW, the compression keeps graph structure, so if an entry has value 0, it won't get deleted. I think that's the right choice because some things like Gauss-Seidel deal with structurally symmetric, numerically nonsymmetric matrices, so you'd want to maintain structural symmetriy.

Yes, that is the right way.

srajama1:
I am good with the changes if Luc is good with it.

lucbv:
Yes I am good with this too, I think it is a good improvement to test the matrix transpose algorithm!

brian-kelley:
OK, passed tests now.

srajama1:
Sorry to be this annoying person who always says the same thing, but I have to. New documentation and performance benchmarks please..

srajama1:
Also please add an issue and mark it "Fixed in Develop"  so @ndellingwood keeps track of it for the release :)

brian-kelley:
@srajama1 Added issue #664.

Issue #673, opened 2020-04-02
2-stage GS update breaking cuda/10+rdc build


ndellingwood:
Merge of PR #671 led to the cuda/10+rdc nightly build (Pascal arch) failing on White.

From a glance at the output there are various undefined references, possibly a sign of ETI-related issues when using sptrsv within the gs routines. Sptrsv tests were compiling and running without error prior to the PR, so I'm not sure what is causing the issue with it's usage in GS when rdc is enabled.

Sample errors from Jenkins job:
https://jenkins-son.sandia.gov/job/KokkosKernels_White_CudaSerial_cuda_10_gcc_720_pascal_rdc/36/consoleFull

```
...
01:16:19 ../../src
/libkokkoskernels.a(Sparse_gauss_seidel_symbolic_eti_DOUBLE_ORDINAL_INT_OFFSET_INT_LAYOUTLEFT_EXECSPACE_CUDA_MEMSPACE_CUDASPACE_MEMSPACE_CUDAUVMSPACE.cpp.o): In function `void KokkosSparse::Experimental::sptrsv_symbolic<KokkosKernels::Experimental::KokkosKernelsHandle<int const, int const, double const, Kokkos::Cuda, Kokkos::CudaSpace, Kokkos::CudaUVMSpace>, Kokkos::View<int*, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, Kokkos::MemoryTraits<0u> >, Kokkos::View<int*, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, Kokkos::MemoryTraits<0u> > >(KokkosKernels::Experimental::KokkosKernelsHandle<int const, int const, double const, Kokkos::Cuda, Kokkos::CudaSpace, Kokkos::CudaUVMSpace>*, Kokkos::View<int*, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, Kokkos::MemoryTraits<0u> >, Kokkos::View<int*, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, Kokkos::MemoryTraits<0u> >)':
01:16:19 
[...]

PR #671, opened 2020-03-28
Two-stage & Classical Gauss-Seidel

iyamazaki:
- Initial implementation of Two-stage & Classical Gauss-Seidel.
- New solver options are added to unit-test & perf-test for Gauss-Seidel

PR #675, opened 2020-04-02
specifying memory space for two-stage GS

iyamazaki:
This PR is to fix some of undefined references errors due to PR #671 for two-stage GS (e.g., issue #673 ).


brian-kelley:
Thanks @iyamazaki !

Issue #677, opened 2020-04-06
SPGEMM color algorithms not implemented, _spgemm_impl_color.hpp not used


brian-kelley:
The file ``KokkosSparse_spgemm_impl_color.hpp`` is included in ``KokkosSparse_spgemm_impl.hpp``, but nothing defined in it is used anywhere. The SPGEMM handle actually defines algorithm enum values SPGEMM_KK_COLOR, SPGEMM_KK_MULTICOLOR and SPGEMM_KK_MULTICOLOR2, but those enum constants don't appear anywhere else (I don't know what would happen if you tried to use them, probably just a crash or an empty product).

@srajama1 Is this something we can get rid of? Do you know what Mehmet's idea was in using D2 coloring in SPGEMM? I could try to understand the code if it's of interest.

srajama1:
I think this was added to try out work by McCourt et al. which uses d-2 coloring on structure C for SpGEMM. I don't remember the exact details if we actually tried it and not chose to move to production or if we never tried it. It is quite possible it is the later as we didn't have a good d-2 coloring on GPUs at that point anyway. I think we can remove the dummy arguments now. We can add it back, if we decide to try it again.

brian-kelley:
@srajama1 OK, that sounds interesting. So I'll remove or comment out those enum values then, remove the include of the file in spgemm_impl.hpp (just for the sake of compile time), but leave the spgemm_impl_color.hpp where it is.

brian-kelley:
"Sparse Matrix-Matrix Products Executed Through Coloring": https://www.researchgate.net/publication/276307474_Sparse_Matrix-Matrix_Products_Executed_Through_Coloring

PR #679, opened 2020-04-08
Fix and clean up spgemm color

brian-kelley:
- Fix wrong include guards in ``KokkosSparse_spgemm_impl_color.hpp`` by removing them. It had the same macro name as another include guard, and the file is intended to be included in one place. The other spgemm_impl files don't have them, either.
- Call ``transpose_graph()`` in that file instead of ``kk_transpose_graph()``, which was removed. This wasn't noticed until trying to migrate two-stage GS stuff to the Trilinos, which didn't have the above include guard problem.
- Removed the SPGEMM coloring-based algorithms from the enum of algorithm choices (#677). These weren't actually implemented. Might be an interesting research topic later (McCourt paper) but for now don't give users the impression that it's supported.
- Clean up some commented out lines from the SPGEMM handle.

Tested clean on kokkos-dev2 and ride.


iyamazaki:
Thank you, @brian-kelley !!

PR #700, opened 2020-04-24
cm_test_all_sandia: increase intel coverage on kokkos-dev*

ndellingwood:


ndellingwood:
@brian-kelley this increases intel coverage during spot-check on the kokkos-dev* machines and adds clang/9, let me know if this closes enough of the coverage gaps.

ndellingwood:
>  there was the other thing with GCC 4.8.4 and OpenMP. Is that something that belongs in the spot check on kokkos-dev2

@brian-kelley good call, I just pushed an update to add that as well. Hopefully we can drop gcc/4.8.4, I think Trilinos is planning to require c++14 soon which will be nice to move past old compilers.

PR #708, opened 2020-04-29
Fix 703: correct CrsMatrix documentation

brian-kelley:
for raw host array constructor. It said it takes COO format, but it really takes CRS format. Also rename "rows" to "rowmap" to avoid confusion. Also remove unused parameter "pad". I'm 99% sure nobody ever passed anything in for pad, but I'm running spot checks just in case.

### DIFF
--- a/src/sparse/KokkosSparse_CrsMatrix.hpp
+++ b/src/sparse/KokkosSparse_CrsMatrix.hpp
@@ -506,11 +506,11 @@ public:
   {}
 
   /// \brief Constructor that copies raw arrays of host data in
-  ///   coordinate format.
+  ///   3-array CRS (compresed row storage) format.
   ///
-  /// On input, each entry of the sparse matrix is stored in val[k],
-  /// with row index rows[k] and column index cols[k].  We assume that
-  /// the entries are sorted in increasing order by row index.
+  /// On input, the entries must be sorted by row. \c rowmap determines where each row begins
+  /// and ends. For each entry k (0 <= k < annz), \c cols[k] gives the adjacent column,
+  /// and \c val[k] gives the corresponding matrix value.
   ///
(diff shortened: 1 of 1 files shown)


src/sparse/KokkosSparse_CrsMatrix.hpp
```diff
@@ -519,27 +519,21 @@ class CrsMatrix {
...
   CrsMatrix (const std::string &label,
              OrdinalType nrows,
              OrdinalType ncols,
              size_type annz,
              ScalarType* val,
-             OrdinalType* rows,
-             OrdinalType* cols,
-             bool pad = false)
+             OrdinalType* rowmap,
+             OrdinalType* cols)
```
srajama1, line 534: You are sure Tpetra doesn't call this constructor w/ pads set ?

brian-kelley: @srajama1 I'll find out. But the pad argument never had any effect, so I doubt it. We can also catch and fix that easily during promotion if anyone did pass in something.

brian-kelley:
@srajama1 I did an all-package Trilinos build with the parameter removed and it was OK, so I think this is ready.

brian-kelley:
Oops, MueLu did pass in pad in one place. I'll get rid of it and explain that the parameter has no  effect and a default value.

srajama1:
Yeah, that is what I was worried about. We need to fix MueLU when this gets into Trilinos then. Technically, this is a backwards incompatible change, so we shouldn't do it. However, this is such an obscure future we might be ok.

brian-kelley:
@srajama1 Just opened https://github.com/trilinos/Trilinos/pull/7280 to remove it in MueLu, so that this won't come up when we put together the next snapshot.

srajama1:
Thanks, @brian-kelley !

PR #706, opened 2020-04-28
Fix705: load module environment in reproducer script

brian-kelley:
On each machine, add the command(s) to load the full module environment (e.g. loading the local modulefile, running SEMS scripts) to the ``reload_modules.sh`` script for reproducing. This way, the reproducer instructions should work no matter what state the environment is in.

PR #709, opened 2020-04-29
Fix module environment setup in cm_test_all_sandia

brian-kelley:
Needed to use "eval $MODULE_ENVIRONMENT" instead of "$MODULE_ENVIRONMENT" to get the right behavior with compound commands, like "commandA ; commandB". Spot check is running now on kokkos-dev2 from a clean environment (which #706 broke, and it passed because the environment was already set up).


brian-kelley:
@ndellingwood OK, kokkos-dev-2 spot check was good, and I was able to do a fresh login and go through the new reproducer instructions and that worked.

PR #718, opened 2020-05-04
Add "module purge" to reproducer reload_modules.sh

brian-kelley:
This is just a one-line change to cm_test_all_sandia. It fixes the reproducer instructions for bowman, which was saying "module intel/18... conflicts with loaded module...". This matches the actual behavior of the testing script, which purges module before each build.

### DIFF
--- a/scripts/cm_test_all_sandia
+++ b/scripts/cm_test_all_sandia
@@ -1142,6 +1142,7 @@ single_build_and_test() {
   then
     echo "        $MODULE_ENVIRONMENT" &>> reload_modules.sh
   fi
+  echo "        module purge" &>> reload_modules.sh
   echo "        module load $compiler_modules_list" &>> reload_modules.sh
   echo "" &>> reload_modules.sh
   chmod +x reload_modules.sh


scripts/cm_test_all_sandia
```diff
@@ -1142,6 +1142,7 @@ single_build_and_test() {
   then
     echo "        $MODULE_ENVIRONMENT" &>> reload_modules.sh
   fi
+  echo "        module purge" &>> reload_modules.sh
```
ndellingwood, line 1145: @brian-kelley shouldn't this be placed before the `$MODULE_ENVIRONMENT` call? Otherwise I think this will blow away modules for some of the systems.

brian-kelley: @ndellingwood Are you concerned that``purge`` will undo commands in ``$MODULE_ENVIRONMENT``? I was worried about that too, but before my changes (before Apr 28) the script also did the machine-specific setup once and then purged before each build.

``sems-env`` is the first in the list of modules for the builds that need it, and I just made sure that ``module use`` isn't undone by purge.

ndellingwood: Thanks for checking, merge when you're ready

PR #722, opened 2020-05-21
update testing scripts

ndellingwood:
Fix ARM arch codes, build list for gcc/9.1 on kokkos-dev-2


scripts/cm_test_all_sandia
```diff
@@ -1243,7 +1243,7 @@ single_build_and_test() {
...
+    echo "  #     \$KOKKOSKERNELS_PATH/cm_generate_makefile.bash --with-devices=$LOCAL_KOKKOS_DEVICES $ARCH_FLAG --compiler=$(which $compiler_exe) --cxxflags=\"$cxxflags\" --cxxstandard=\"$cxx_standard\" --ldflags=\"$ldflags\" $CUDA_ENABLE_CMD --kokkos-path=\$KOKKOS_PATH --kokkoskernels-path=\$KOKKOSKERNELS_PATH --with-scalars=$kk_scalars --with-ordinals=${KOKKOSKERNELS_ORDINALS} --with-offsets=${KOKKOSKERNELS_OFFSETS} --with-layouts=${KOKKOSKERNELS_LAYOUTS} ${KOKKOSKERNELS_ENABLE_TPL_CMD} ${KOKKOSKERNELS_TPL_PATH_CMD} ${KOKKOSKERNELS_TPL_LIBS_CMD} ${KOKKOSKERNELS_EXTRA_LINKER_FLAGS_CMD} --with-options=${KOKKOS_OPTIONS} --with-cuda-options=${KOKKOS_CUDA_OPTIONS} --no-examples $extra_args" &> call_generate_makefile_genericpath.sh
```
e10harvey, line 1246: This is no longer appending to cm_call_generate_makegraph

ndellingwood: Correct, I intended this to overwrite the file each run when storing the generate_makefile incantation with a generic path vars (easier copy-paste for reporting issues etc), the append was copy-pasta from the line that stores the specific path of the run, which is used by the reproducer script.

PR #729, opened 2020-05-28
Possible interface fix for batched macros

DavidPoliakoff:
Hey folks, I was building Trilinos when I ran into a compilation error centered around `KokkosBatched::TeamVectorTrsvInternalLower` being given template arguments, when it needed to have no template arguments. The following patch fixed it.

I don't really know what I'm doing in Kokkos Kernels, so this should be read really carefully to make sure I'm not missing anything


kyungjoo-kim:
This should be fixed on the other side. Algorithm tag is required for users overriding and testing capability. I will submit another PR for fixing this.

DavidPoliakoff:
@kyungjoo-kim , thanks, I'll close this one

PR #731, opened 2020-05-29
First iteration of HashmapAccumulator cleanup

### LINKED ISSUE #508: HashmapAccumulator has several unused members, misnamed parameters
The HashmapAccumulator has three members (``hash_key_size``, ``max_value_size``, and ``used_size``) that are not used internally, but they are set by the constructors. ``used_size_`` and ``max_value_size_`` (with trailing underscores) are instead passed to the insert() routines. 

``hash_key_size_`` is not passed to any insert routines. It is the number of possible hash outputs, aka 1 plus the bit mask that forms the hash function: ``hash(value) = value & (hash_key_size - 1)``.

I think these changes would make sense and make things a lot more readable and debuggable:

- ``max_value_size`` is the length of the ``keys`` and ``nexts`` arrays, so it should be an immutable member. ``max_value_size_`` should not be passed into the insert routines. In fact, what SPGEMM symbolic does now is read out the member from the hashmap and pass it in.
- ``max_value_size`` should really be called ``capacity`` or something similar - it's meaning is the length of the two arrays (keys, nexts).
- ``used_size`` is an atomically incremented counter that measures how full the hashmap is. It is updated in parallel, so it usually lives in shared/global memory. It should either not be a member and just be passed by ref to insert, OR kept as a member, but be a pointer.
- 3 ways to deal with ``hash_key_size``:
  - Remove the member, since it is unused
  - Keep it as an immutable member, and have ``KOKKOS_FORCEINLINE_FUNCTION key_type hash()`` be a member of Hashmap. The SPGEMM code wouldn't need to keep the hash mask as a local variable anymore.
  - Keep as immutable member, and instead of passing the both the key and the hash to insert(), just pass the key, and compute the hash internally.

e10harvey:
The following changes were made to the HashmapAccumulator class:
1. `max_value_size` changed to `private` member `__max_value_size`.
2. `hash_key_size` was removed.
3. `used_size` was removed.
4. Hashes are now computed within the `HashmapAccumulator` insertion routines:
    - `__hashOpRHS` was added as a `private` member. 
    - `__compute_hash` was added as a `private` member. This function is selected at compile time via a templated argument to the `HashmapAccumulator` constructor.
    - `vector_atomic_insert_into_hash_mergeAdd_with_team_level_list_length` does not compute hashes internally due to the special use-case in `KokkosSparse_spgemm_impl_speed.hpp:operator GPUTag`.

Fixes #508.

## spot-checks
# 1
```
<snip>
WARNING!! THE FOLLOWING CHANGES ARE UNCOMMITTED!! :
?? build/
?? testing/

KokkosKernels Repository Status:  7ff59541f124a9756cabd2c6949758500db58afe common/HashmapAccumulator: cleanup insert fn declarations more

Kokkos Repository Status:  33f730c475802bc226fc17e42e58fe7612d86b41 Merge pull request #3073 from masterleinad/enable_travis_compiler_warnings


Going to test compilers:  gcc/6.4.0 gcc/7.2.0 ibm/16.1.1 cuda/9.2.88 cuda/10.1.105
<snip>
PASSED TESTS
cuda-10.1.105-Cuda_OpenMP-release build_time=647 run_time=170
cuda-10.1.105-Cuda_Serial-release build_time=587 run_time=186
cuda-9.2.88-Cuda_OpenMP-release build_time=664 run_time=185
cuda-9.2.88-Cuda_Serial-release build_time=650 run_time=201
gcc-6.4.0-OpenMP_Serial-release build_time=228 run_time=139
gcc-7.2.0-OpenMP-release build_time=158 run_time=61
gcc-7.2.0-OpenMP_Serial-release build_time=220 run_time=138
gcc-7.2.0-Serial-release build_time=141 run_time=71
ibm-16.1.1-Serial-release build_time=1012 run_time=74
```
# 2
```
KokkosKernels Repository Status:  7ff59541f124a9756cabd2c6949758500db58afe common/HashmapAccumulator: cleanup insert fn declarations more

Kokkos Repository Status:  33f730c475802bc226fc17e42e58fe7612d86b41 Merge pull request #3073 from masterleinad/enable_travis_compiler_warnings


Going to test compilers:  cuda/9.2
Testing compiler cuda/9.2
  Starting job cuda-9.2-Cuda_OpenMP-release
kokkos devices: Cuda,OpenMP
kokkos arch: Kepler35
kokkos options: 
kokkos cuda options: force_uvm
kokkos cxxflags: -O3 -Wall -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wuninitialized 
extra_args: 
kokkoskernels scalars: 'double,complex_double'
kokkoskernels ordinals: int
kokkoskernels offsets: int,size_t
kokkoskernels layouts: LayoutLeft
[eharvey@kokkos-dev testing]$ tail -f do-test-issue-508-05-29-2020.out 
kokkos devices: Cuda,OpenMP
kokkos arch: Kepler35
kokkos options: 
kokkos cuda options: force_uvm
kokkos cxxflags: -O3 -Wall -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wuninitialized 
extra_args: 
kokkoskernels scalars: 'double,complex_double'
kokkoskernels ordinals: int
kokkoskernels offsets: int,size_t
kokkoskernels layouts: LayoutLeft
  PASSED cuda-9.2-Cuda_OpenMP-release
PASSED TESTS
cuda-9.2-Cuda_OpenMP-release build_time=1095 run_time=264
```
# 3
```
<snip>
WARNING!! THE FOLLOWING CHANGES ARE UNCOMMITTED!! :
?? build-issue-508/
?? build.750fe245/
?? build/
?? do-cmake.sh
?? do-test.sh
?? issue-727/
?? testing/

KokkosKernels Repository Status:  7ff59541f124a9756cabd2c6949758500db58afe common/HashmapAccumulator: cleanup insert fn declarations more

Kokkos Repository Status:  cb9727fae308ce7ae2248dbb8168c430d958bc32 core/src/impl: Conditionally define get_gpu in Kokkos_Core
<snip>
PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=687 run_time=159
clang-8.0-Pthread_Serial-release build_time=207 run_time=110
clang-9.0.0-Pthread-release build_time=126 run_time=61
clang-9.0.0-Serial-release build_time=207 run_time=56
cuda-10.1-Cuda_OpenMP-release build_time=806 run_time=149
cuda-9.2-Cuda_Serial-release build_time=753 run_time=180
gcc-4.8.4-OpenMP-release build_time=126 run_time=58
gcc-7.3.0-OpenMP-release build_time=136 run_time=57
gcc-7.3.0-Pthread-release build_time=111 run_time=52
gcc-8.3.0-Serial-release build_time=135 run_time=58
gcc-9.1-OpenMP-release build_time=170 run_time=58
gcc-9.1-Serial-release build_time=155 run_time=62
intel-17.0.1-Serial-release build_time=262 run_time=65
intel-18.0.5-OpenMP-release build_time=338 run_time=53
intel-19.0.5-Pthread-release build_time=469 run_time=56
<snip>
WARNING!! THE FOLLOWING CHANGES ARE UNCOMMITTED!! :
?? build-issue-508/
?? build.750fe245/
?? build/
?? do-cmake.sh
?? do-test.sh
?? issue-727/
?? testing/

KokkosKernels Repository Status:  7ff59541f124a9756cabd2c6949758500db58afe common/HashmapAccumulator: cleanup insert fn declarations more

Kokkos Repository Status:  33f730c475802bc226fc17e42e58fe7612d86b41 Merge pull request #3073 from masterleinad/enable_travis_compiler_warnings


Going to test compilers:  gcc/7.3.0 gcc/8.3.0 gcc/9.1 gcc/4.8.4 intel/17.0.1 intel/18.0.5 intel/19.0.5 clang/8.0 clang/9.0.0 cuda/10.1
<snip>
PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=606 run_time=153
clang-8.0-Pthread_Serial-release build_time=197 run_time=111
clang-9.0.0-Pthread-release build_time=115 run_time=53
clang-9.0.0-Serial-release build_time=126 run_time=57
cuda-10.1-Cuda_OpenMP-release build_time=837 run_time=151
gcc-4.8.4-OpenMP-release build_time=116 run_time=60
gcc-7.3.0-OpenMP-release build_time=138 run_time=59
gcc-7.3.0-Pthread-release build_time=110 run_time=53
gcc-8.3.0-Serial-release build_time=136 run_time=59
gcc-9.1-OpenMP-release build_time=169 run_time=60
gcc-9.1-Serial-release build_time=153 run_time=59
intel-17.0.1-Serial-release build_time=245 run_time=56
intel-18.0.5-OpenMP-release build_time=363 run_time=53
intel-19.0.5-Pthread-release build_time=442 run_time=47
```

## When enabling all scalars
`$ ./KokkosKernels_sparse_cuda --gtest_filter=cuda.sparse_spmv_*` still results in:
```
[==========] 32 tests from 1 test case ran. (233294 ms total)
[  PASSED  ] 20 tests.
[  FAILED  ] 12 tests, listed below:
[  FAILED  ] cuda.sparse_spmv_float_int_int_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_struct_float_int_int_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_float_int_size_t_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_struct_float_int_size_t_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_kokkos_complex_float_int_int_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_struct_kokkos_complex_float_int_int_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_kokkos_complex_float_int_size_t_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_struct_kokkos_complex_float_int_size_t_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_float_int_int_LayoutLeft_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_float_int_size_t_LayoutLeft_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_kokkos_complex_float_int_int_LayoutLeft_TestExecSpace
[  FAILED  ] cuda.sparse_spmv_mv_kokkos_complex_float_int_size_t_LayoutLeft_TestExecSpace

12 FAILED TESTS
```
and `$ ./KokkosKernels_sparse_cuda --gtest_filter=cuda.sparse_spgemm_*` still results in:
```
[==========] 16 tests from 1 test case ran. (127993 ms total)
[  PASSED  ] 12 tests.
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] cuda.sparse_spgemm_jacobi_float_int_int_TestExecSpace
[  FAILED  ] cuda.sparse_spgemm_jacobi_float_int_size_t_TestExecSpace
[  FAILED  ] cuda.sparse_spgemm_jacobi_kokkos_complex_float_int_int_TestExecSpace
[  FAILED  ] cuda.sparse_spgemm_jacobi_kokkos_complex_float_int_size_t_TestExecSpace

 4 FAILED TESTS
```

### DIFF
--- a/src/common/KokkosKernels_HashmapAccumulator.hpp
+++ b/src/common/KokkosKernels_HashmapAccumulator.hpp
@@ -174,14 +174,13 @@ struct HashmapAccumulator
   //Insertion is sequential, no race condition for the insertion.
   KOKKOS_INLINE_FUNCTION
   int sequential_insert_into_hash_mergeOr_TrackHashes (
-      size_type hash,
       key_type key,
       value_type value,
       size_type *used_size_,
       size_type *used_hash_size,
       size_type *used_hashes)
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    size_type hash = __compute_hash(key, __hashOpRHS);
 
     size_type i = hash_begins[hash];
     for (; i != -1; i = hash_nexts[i]) {
@@ -212,7 +211,6 @@ struct HashmapAccumulator
   //Assume that there are 2 values for triangle count.
   KOKKOS_INLINE_FUNCTION
   int sequential_insert_into_hash_mergeOr_TriangleCount_TrackHashes (
-      size_type hash,
       key_type key,
       value_type value,
       value_type *values2,
@@ -220,7 +218,7 @@ struct HashmapAccumulator
       size_type *used_hash_size,
       size_type *used_hashes)
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    size_type hash = __compute_hash(key, __hashOpRHS);
 
     size_type i = hash_begins[hash];
     for (; i != -1; i = hash_nexts[i]) {
@@ -251,7 +249,6 @@ struct HashmapAccumulator
   //L x Incidence
   KOKKOS_INLINE_FUNCTION
   int sequential_insert_into_hash_mergeAnd_TriangleCount_TrackHashes (
-      size_type hash,
       key_type key,
       value_type value,
       value_type *values2,
@@ -259,7 +256,7 @@ struct HashmapAccumulator
       size_type *used_hash_size,
       size_type *used_hashes)
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    size_type hash = __compute_hash(key, __hashOpRHS);
 
     //this function will only try to do an AND operation with
     //existing keys. If the key is not there, returns __insert_full.
@@ -279,11 +276,10 @@ struct HashmapAccumulator
   //this is used in LxL or Incidence^T x L
   KOKKOS_INLINE_FUNCTION
   value_type sequential_insert_into_hash_mergeAnd_TriangleCount_TrackHashes (
-      size_type hash,
       key_type key,
       value_type value)
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    size_type hash = __compute_hash(key, __hashOpRHS);
 
     //this function will only try to do an AND operation with
     //existing keys. If the key is not there, returns __insert_full.
@@ -302,7 +298,6 @@ struct HashmapAccumulator
   //L x Incidence
   KOKKOS_INLINE_FUNCTION
   int sequential_insert_into_hash_TriangleCount_TrackHashes (
-      size_type hash,
       key_type key,
       value_type value,
       value_type *values2,
@@ -310,7 +305,7 @@ struct HashmapAccumulator
       size_type *used_hash_size,
       size_type *used_hashes)
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    size_type hash = __compute_hash(key, __hashOpRHS);
 
     //this function will directly insert, won't check if it exists already.
     if (*used_size_ >= __max_value_size) return __insert_full;
@@ -334,14 +329,13 @@ struct HashmapAccumulator
   //this is used in LxL or Incidence^T x L
   KOKKOS_INLINE_FUNCTION
   int sequential_insert_into_hash_TriangleCount_TrackHashes (
-      size_type hash,
       key_type key,
       value_type value,
       size_type *used_size_,
       size_type *used_hash_size,
       size_type *used_hashes) // TODO figure out what this "used_hashes" is for
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    size_type hash = __compute_hash(key, __hashOpRHS);
 
     //this function will directly insert, won't check if it exists already.
     if (*used_size_ >= __max_value_size) return __insert_full;
@@ -366,14 +360,13 @@ struct HashmapAccumulator
   //the mergeadd used in the numeric of KKMEM.
   KOKKOS_INLINE_FUNCTION
   int sequential_insert_into_hash_mergeAdd_TrackHashes (
-      size_type hash,
       key_type key,
       value_type value,
       size_type *used_size_,
       size_type *used_hash_size,
       size_type *used_hashes)
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    size_type hash = __compute_hash(key, __hashOpRHS);
 
     // issue-508, TODO: ensure that i < __max_value_size, but
     // need information about length of keys, values, and hash_nexts first!
@@ -404,13 +397,12 @@ struct HashmapAccumulator
   //also used in the symbolic of spgemm if no compression is applied.
   KOKKOS_INLINE_FUNCTION
   int sequential_insert_into_hash_TrackHashes (
-      size_type hash,
       key_type key,
       size_type *used_size_,
       size_type *used_hash_size,
       size_type *used_hashes)
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    size_type hash = __compute_hash(key, __hashOpRHS);
 
     size_type i = hash_begins[hash];
     for (; i != -1; i = hash_nexts[i]) {
@@ -439,13 +431,8 @@ struct HashmapAccumulator
   //insertions will have the same key.
   //Insertion is simulteanous for the vector lanes of a thread.
   //used_size should be a shared pointer among the thread vectors
-  template <typename team_member_t>
   KOKKOS_INLINE_FUNCTION
   int vector_atomic_insert_into_hash_mergeAdd_TrackHashes (
-      const team_member_t & /* teamMember */,
-      const int /* vector_size */,
-
-      size_type hash,
       const key_type key,
       const value_type value,
       volatile size_type *used_size_,
@@ -453,7 +440,7 @@ struct HashmapAccumulator
       size_type *used_hashes
       )
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    size_type hash = __compute_hash(key, __hashOpRHS);
 
     if (hash != -1) {
       size_type i = hash_begins[hash];
@@ -513,7 +500,8 @@ struct HashmapAccumulator
       volatile size_type *used_size_,
       const size_type max_value_size_)
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    // Cannot compute hash here due to impl_speed use-case
+    //hash = __compute_hash(key, __hashOpRHS);
 
     if (hash != -1) {
       size_type i = hash_begins[hash];
@@ -568,19 +556,15 @@ struct HashmapAccumulator
   //insertions will have the same key.
   //Insertion is simulteanous for the vector lanes of a thread.
   //used_size should be a shared pointer among the thread vectors
-  template <typename team_member_t>
   KOKKOS_INLINE_FUNCTION
   int vector_atomic_insert_into_hash_mergeAdd (
-      const team_member_t & /* teamMember */,
-      const int /* vector_size */,
-      size_type &hash,
       const key_type key,
       const value_type value,
       volatile size_type *used_size_)
   {
     return vector_atomic_insert_into_hash_mergeAdd_with_team_level_list_length(nullptr,
                                                                                0,
-                                                                               hash,
+                                                                               __compute_hash(key, __hashOpRHS),
                                                                                key,
                                                                                value,
                                                                                used_size_,
@@ -588,27 +572,19 @@ struct HashmapAccumulator
   }
 
   //used in symbolic of kkmem if the compression is not applied.
-  template <typename team_member_t>
   KOKKOS_INLINE_FUNCTION
   int vector_atomic_insert_into_hash (
-      const team_member_t & /* teamMember */,
-      const int &/* vector_size */,
-      size_type hash,
       const key_type &key,
       volatile size_type *used_size_
       )
   {
-    hash = __compute_hash(key, __hashOpRHS);
+    size_type hash = __compute_hash(key, __hashOpRHS);
 
-    if (hash != -1) {
-      size_type i = hash_begins[hash];
-      for (; i != -1; i = hash_nexts[i]) {
-        if (keys[i] == key) {
-          return __insert_success;
-        }
-      }
-    } else {
+    size_type i = hash_begins[hash];
+    for (; i != -1; i = hash_nexts[i]) {
+      if (keys[i] == key) {
         return __insert_success;
+      }
     }
 
     size_type my_write_index = Kokkos::atomic_fetch_add(used_size_, size_type(1));
--- a/src/common/KokkosKernels_HashmapAccumulator.hpp
+++ b/src/common/KokkosKernels_HashmapAccumulator.hpp
@@ -791,6 +791,8 @@ struct HashmapAccumulator
       #ifdef HASHMAPACCUMULATOR_ASSERT_ENABLED
         if (hash == -1)
           Kokkos::abort("__compute_hash: hash = -1");
+        if (key == -1)
+          Kokkos::abort("__compute_hash: key = -1");
       #endif // HASHMAPACCUMULATOR_ASSERT_ENABLED
       return hash;
     }
@@ -803,6 +805,8 @@ struct HashmapAccumulator
       #ifdef HASHMAPACCUMULATOR_ASSERT_ENABLED
         if (hash == -1)
           Kokkos::abort("__compute_hash: hash = -1");
+        if (key == -1)
+          Kokkos::abort("__compute_hash: key = -1");
       #endif // HASHMAPACCUMULATOR_ASSERT_ENABLED
       return hash;
     }
@@ -815,6 +819,8 @@ struct HashmapAccumulator
       #ifdef HASHMAPACCUMULATOR_ASSERT_ENABLED
         if (hash == -1)
           Kokkos::abort("__compute_hash: hash = -1");
+        if (key == -1)
+          Kokkos::abort("__compute_hash: key = -1");
       #endif // HASHMAPACCUMULATOR_ASSERT_ENABLED
       return hash;
     }
--- a/src/common/KokkosKernels_HashmapAccumulator.hpp
+++ b/src/common/KokkosKernels_HashmapAccumulator.hpp
@@ -50,20 +50,34 @@ namespace KokkosKernels {
 
 namespace Experimental {
 
+template <typename size_type, typename key_type, typename value_type>
 /**
- * HashmapAccumulator
- *
+ * \brief HashmapAccumulator class
  * The use of this is described in the paper:
  *   "Performance-portable sparse matrix-matrix multiplication for many-core architectures"
  *   ( https://ieeexplore.ieee.org/abstract/document/7965111/ ) in section III.D
- *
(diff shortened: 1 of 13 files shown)


seheracer:
Which machine are these spot-check results from?

src/sparse/impl/KokkosSparse_spgemm_impl_symbolic.hpp
```diff
@@ -229,21 +229,23 @@ struct KokkosSPGEMM
...
     }
 
     //set first to globally used hash indices.
     nnz_lno_t *globally_used_hash_indices = (nnz_lno_t *) tmp;
     tmp += pow2_hash_size;
 
     //create hashmap accumulator.
-    KokkosKernels::Experimental::HashmapAccumulator<nnz_lno_t,nnz_lno_t,nnz_lno_t> hm2;
+    KokkosKernels::Experimental::HashmapAccumulator<nnz_lno_t,nnz_lno_t,nnz_lno_t,KokkosKernels::Experimental::HashOpType::bitwiseAnd> 
+    hm2(chunk_size, pow2_hash_func, nullptr, nullptr, nullptr, nullptr);
```
brian-kelley, line 248: Does this work now, with chunk_size = 0? I'm worried that the new insert routines will compare against that and return __insert_full immediately.

src/sparse/impl/KokkosSparse_spgemm_impl_symbolic.hpp
```diff
@@ -688,11 +683,11 @@ struct KokkosSPGEMM
    * \param entriesA_: col indices of A
    * \param row_ptr_begins_B_: beginning of the rows of B
    * \param row_ptr_ends_B_:end of the rows of B
-   * \param entriesSetIndicesB_: column set indices of B
-   * \param entriesSetsB_: columns sets of B
+   * \param entriesSetIndicesB_: column set indices of B [CSI]
```
brian-kelley, line 686: Thanks for adding the array names from the paper.

src/sparse/impl/KokkosSparse_spgemm_impl_symbolic.hpp
brian-kelley: For the first arg, you should just be able to pass INT_MAX (or the equivalent for whatever type) because the chunk size is computed to always be big enough for the longest row, so insertion should never return __insert_full.

src/sparse/impl/KokkosSparse_spgemm_impl_triangle.hpp
```diff
@@ -1123,21 +1111,23 @@ struct KokkosSPGEMM
...
     }
 
     //set first to globally used hash indices.
     nnz_lno_t *globally_used_hash_indices = (nnz_lno_t *) tmp;
     tmp += pow2_hash_size;
 
     //create hashmap accumulator.
-    KokkosKernels::Experimental::HashmapAccumulator<nnz_lno_t,nnz_lno_t,nnz_lno_t> hm2;
+    KokkosKernels::Experimental::HashmapAccumulator<nnz_lno_t,nnz_lno_t,nnz_lno_t,KokkosKernels::Experimental::HashOpType::bitwiseAnd> 
+    hm2(chunk_size, pow2_hash_func, nullptr, nullptr, nullptr, nullptr);
```
brian-kelley, line 1130: Same question as above about chunk_size=0

src/common/KokkosKernels_HashmapAccumulator.hpp
```diff
@@ -729,15 +811,63 @@ struct HashmapAccumulator
...
+    size_type __hashOpRHS;
+    const int __insert_success;
+    const int __insert_full;
+
+    template<typename U = hash_type, typename std::enable_if<std::is_same<U, HashOpType::bitwiseAnd>::value, std::size_t>::type = 0>
+    KOKKOS_INLINE_FUNCTION
+    int __compute_hash(size_type key, size_type bitmask) {
+      //std::cout << "bitwiseAnd" << std::endl;
+      size_type hash = key & bitmask;
+      #ifdef HASHMAPACCUMULATOR_ASSERT_ENABLED
```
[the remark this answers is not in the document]
e10harvey: @brian-kelley: I want to double check on this with you. `Kokkos::abort` runs on the GPU and generates a core file. It could be useful for debugging to leave these asserts in. Should we remove the debug blocks?

src/sparse/impl/KokkosSparse_spgemm_impl_symbolic.hpp
e10harvey: Good catch! `chunk_size` should be `MaxRoughNonZero` (see https://github.com/kokkos/kokkos-kernels/pull/731/files/ebab3bc86c0e9aebf28b0cc3cdaaaa96419bf058#diff-bba5fd8c6962fac594fa0e5918d87e88L261)!

src/sparse/impl/KokkosSparse_spgemm_impl_triangle.hpp
e10harvey: Good catch! chunk_size should be MaxRoughNonZero (see https://github.com/kokkos/kokkos-kernels/pull/731/files/ebab3bc86c0e9aebf28b0cc3cdaaaa96419bf058#diff-12c0beaaa61cf811b978be6f4dc44d77L1160)!

src/common/KokkosKernels_HashmapAccumulator.hpp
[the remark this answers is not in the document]
brian-kelley: @e10harvey Ah, forget I said anything :)
Didn't know abort could be called from device.

src/sparse/impl/KokkosSparse_spgemm_impl_symbolic.hpp
brian-kelley: @e10harvey Good, that is a better value to use than INT_MAX. So did this not get caught by sparse unit tests?

e10harvey: It did not get caught by the sparse unit tests. I'll track that down and update the tests.

src/sparse/impl/KokkosSparse_spgemm_impl_triangle.hpp
e10harvey: Same as https://github.com/kokkos/kokkos-kernels/pull/731#discussion_r440438891.

src/common/KokkosKernels_HashmapAccumulator.hpp
```diff
@@ -492,44 +572,68 @@ struct HashmapAccumulator
       //hash_nexts is updated second time as below.
       //but this is okay for spgemm,
       //because no two keys will be inserted into hashmap at the same time, as rows have unique columns.
+      
       hash_nexts[my_write_index] = hash_begins[hash];
```
e10harvey, line 576: @brian-kelley: Why is this here? Is this masking a cache consistency issue? When we are in this insertion routine, we are _supposed_ to be running from a single kokkos thread but in parallel from multiple vector lanes, right? So we can think of each of the kokkos thread partition of `keys`, `values`, hash_begins` or `hash_nexts` as shared among vector lanes but not accessed concurrently among kokkos threads, right? Is the loop above reading `keys` and `values` supposed to be protected from reading a stale/uninitialized element of `keys` or `values` through `hash_begins` and `hash_nexts`? Neither `keys`, `values`, `hash_begins` or `hash_nexts` are read from atomically. So, what if the write has not landed? Or, what if the write to `hash_nexts` has landed but the writes to `values` and `keys` have not landed? Does `key_type`, `value_type`, and `size_type` guarantee atomic reads and where are those types enforced?

e10harvey: If a write to `keys` or `values` doesn't land before it's read, what guarantees that we do not get into the block starting at https://github.com/kokkos/kokkos-kernels/pull/731/files#diff-31dd0387f00dc52299083b47296b28b0R561? And, couldn't `values` at https://github.com/kokkos/kokkos-kernels/pull/731/files#diff-31dd0387f00dc52299083b47296b28b0R543 become corrupted in this scenario?

brian-kelley: @e10harvey That line is to make sure there is no break in the linked list observed by vector lane A, if vector lane B is in between lines 583 and 584 (in your branch). In that situation, lane B has replaced ``hash_begins[hash]`` by atomic exchange, but it hasn't yet made the link from ``my_write_index`` to ``hashbeginning``. Without line 576, at the instant that B is between 583 and 584, the list only contain's B's new node ``my_write_index`` (it's the hash begin, but its next is -1). Because of the assumption that we're not inserting the same key on multiple lanes at the same time, the invariant that matters is that walking the list gives you all keys that were previously inserted.

brian-kelley: If we didn't have this assumption, this insertion code wouldn't be correct at all. It would be necessary to insert at the end instead of the beginning, and it needs an atomic compare-and-swap, not an exchange.

src/sparse/impl/KokkosSparse_spgemm_impl_symbolic.hpp
e10harvey: @brian-kelley: I moved this to a separate issue: https://github.com/kokkos/kokkos-kernels/issues/740.

src/common/KokkosKernels_HashmapAccumulator.hpp
e10harvey: @brian-kelley I'm still trying to understand. Why are we not performing: `Kokkos::atmoic_store(hash_nexts+my_write_index, Kokkos::atomic_exchange(hash_begins+hash, my_write_index))` to ensure the linked list is not broken?

brian-kelley: @e10harvey That's not equivalent to doing atomic_exchange. Say the list was
```
begin -> 1
```
and vector lane A is inserting 2, and B is inserting 3 at the same time. After the first line is executed, the links are:
```
begin -> 1
2 -> 1
3 -> 1
```
then after the 2nd line, depending on the order that the 2nd atomic transaction happens, it could be either:
```
begin -> 2 -> 1
3 -> 1
```
or
```
begin -> 3 -> 1
2 -> 1
```
so one of the insertions fails.

In the same scenario, here's what the actual code does. After line 576:
```
begin -> 1
2 -> 1
3 ->1
```
Then one of them does the atomic exchange (without loss of generality, say it's A):
```
begin -> 2 -> 1, and hashbeginning is 1 on lane A
3 -> 1
```
Now, either B does the atomic exchange next, and then A and B both update nexts:
```
begin -> 3 -> 1. Hashbeginning is 1 on lane A, and 2 on lane B
2 -> 1
```
```
begin -> 3 -> 2 -> 1
```
OR, A updates its next first (doesn't change anything, since 2's next was already 1), and then B does the exchange followed by the next update:
```
begin -> 3 -> 1, and hashbeginning is 2 on lane B
2 -> 1
```
```
begin -> 3 -> 2 -> 1
```

So in any case, at all points in time, walking the list from begin will give you at least the keys that were in the list before (1). There are other ways the 3 lines could interleave, but I convinced myself that they all end up with 2 and 3 inserted, while always keeping 1 in the list. No matter what happens, the original list (starting at 1) is either reachable directly from begin, or from one of the new nodes. A new node can't replace begin until its next reaches the original list.

brian-kelley: I talked with Christian about what ``atomic_store`` does, and it's basically a volatile write (a normal write, but the compiler can't optimize away loads and stores involving that address across the atomic). There is no execution barrier before or after, so different threads/vector lanes can interleave like I described above

e10harvey: Thanks Brian. What I am trying to ask is whether we should be performing:
```C
Kokkos::atomic_store(hash_nexts+my_write_index, Kokkos::atomic_exchange(hash_begins+hash, my_write_index));
```

In other words, is it OK for line 576 and 583 not to be atomic? I'm having trouble understanding but I *think* there could be a data race when updating hash_nexts here. I'll try to come up with a data race scenario.

brian-kelley: Lines 576 and 584 can't data race because they are both writing to ``hash_nexts[my_write_index]``, and ``my_write_index`` is guaranteed to be unique across all vector lanes and iterations, because it's counted up atomically.

e10harvey:
@ndellingwood, @srajama1: This is ready to merge.

ndellingwood:
Thanks @e10harvey !

PR #743, opened 2020-06-17
Simplified SpADD/SpGEMM interface

lucbv:
@srajama1 @seheracer 
I would like to provide a simplified SpGEMM interface, it only requires the CrsMatrices (A, B and C), the kernel handle and flags for the potential transposition of A and/or B.
The proposed interface wraps both symbolic and numeric calls as it is a fairly common use case (typically need to do so at least the first time around).
I am also adding an example that I want to put in the wiki once/if this PR merges, it simplifies the example and actually compiles and run.
Finally I am not sure how we want to test this? Right now I only have the wiki example as this is simply an interface that does not add any algorithmic component but I could add something in the SpGEMM unit test too?

### DIFF
--- a/src/sparse/KokkosSparse_spgemm_numeric.hpp
+++ b/src/sparse/KokkosSparse_spgemm_numeric.hpp
@@ -188,9 +188,9 @@ void spgemm_numeric(
       UniformDevice_t, //       typename bscalar_nnz_view_t_::device_type,
       Kokkos::MemoryTraits<Kokkos::Unmanaged> > Internal_bscalar_nnz_view_t_;
 
-  //static assert clno_row_view_t_ cannot be const type.
+  //static assert clno_row_view_t_ can be const type (row map is fixed after symbolic phase).
   typedef Kokkos::View<
-      typename clno_row_view_t_::non_const_value_type*,
+      typename clno_row_view_t_::value_type*,
       typename KokkosKernels::Impl::GetUnifiedLayout<clno_row_view_t_>::array_layout,
       UniformDevice_t, //       typename clno_row_view_t_::device_type,
       Kokkos::MemoryTraits<Kokkos::Unmanaged> > Internal_clno_row_view_t_;
--- a/example/wiki/sparse/CMakeLists.txt
+++ b/example/wiki/sparse/CMakeLists.txt
@@ -14,6 +14,16 @@ KOKKOSKERNELS_ADD_EXECUTABLE_AND_TEST(
   )
 
 KOKKOSKERNELS_ADD_EXECUTABLE_AND_TEST(
+  wiki_spadd
+  SOURCES KokkosSparse_wiki_spadd.cpp
+  )
+
+KOKKOSKERNELS_ADD_EXECUTABLE_AND_TEST(
+  wiki_spgemm
+  SOURCES KokkosSparse_wiki_spgemm.cpp
+  )
+
+KOKKOSKERNELS_ADD_EXECUTABLE_AND_TEST(
   wiki_gauss_seidel
   SOURCES KokkosSparse_wiki_gauss_seidel.cpp
   )
--- a/example/wiki/sparse/KokkosSparse_wiki_spadd.cpp
+++ b/example/wiki/sparse/KokkosSparse_wiki_spadd.cpp
@@ -1,0 +1,69 @@
+#include "Kokkos_Core.hpp"
+
+#include "KokkosKernels_default_types.hpp"
+#include "KokkosSparse_spadd.hpp"
+
+#include "KokkosKernels_Test_Structured_Matrix.hpp"
+
+using Scalar  = default_scalar;
+using Ordinal = default_lno_t;
+using Offset  = default_size_type;
+using Layout  = default_layout;
+
+int main(int argc, char* argv[]) {
+  Kokkos::initialize();
+
+  using device_type = typename Kokkos::Device<
+      Kokkos::DefaultExecutionSpace,
+      typename Kokkos::DefaultExecutionSpace::memory_space>;
+  using execution_space = typename device_type::execution_space;
+  using memory_space    = typename device_type::memory_space;
+  using matrix_type =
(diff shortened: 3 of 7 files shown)


seheracer:
@lucbv, thanks for this simplified interface, it is more readable. I have a question though:

How will the current spgemm users use this new interface for repeated spgemms with the same sparsity structure? I mean, does this new function have the mechanism to reuse the result of a symbolic call for multiple upcoming numeric calls?

lucbv:
@seher no since we do not have a kernel to do that at the moment.
In my mind ideally you add a flag to the current interface, say `reuseGraph = false`.
After that you can a specially kernel that only computes the values.

seheracer:
> @seheracer no since we do not have a kernel to do that at the moment.
> In my mind ideally you add a flag to the current interface, say `reuseGraph = false`.
> After that you can a specially kernel that only computes the values.

Maybe there is a way of using `is_symbolic_called` function of the spgemm handle for this purpose. That is, if we can count on the user using the same spgemm handle for the repeated calls, we can just check if symbolic was already performed and skip it if so. 

An example for calling `is_symbolic_called`:
https://github.com/kokkos/kokkos-kernels/blob/07a60bcc136d4db46f1a05102762a0a17c07049f/src/sparse/impl/KokkosSparse_spgemm_numeric_spec.hpp#L206

lucbv:
Well if the user can set that parameter manually that's fine, if there is no way to set it manually then that is potentially a problem... there are cases in which you can compute efficiently the CrsGraph because you have knowledge regarding the structure of both A and B, multigrid can actually be one such case depending on the operator and coarsening algorithm...

lucbv:
OK, so the symbolic part of spgemm is expecting the row_map of C to be non const which is of course a problem since it should actually not be modified by the symbolic algorithm otherwise what is the point of calling it symbolic?

So we will not have reuse.

lucbv:
Final note on the reuse potential, it does not look like we are testing multiple calls to `spgemm_numeric` without call `spgemm_symbolic`. It is likely that this does not work based on the type of the row_map store in `StaticCrsGraph`.

srajama1:
This worries me. The whole point of having two calls is to separate out the symbolic vs numeric. Vendors went two phase approach after our work. Now we are going single phase ?

lucbv:
@srajam the two phase does not work, I do intend to fix it later but not necessarily before the tutorial. The point here is to make a convenient feature for users though who can still call both separate symbolic and numerical call if they wish to do so.
Typically though I would say that combine the two on the very first call makes sense and afterwards calling the numerical phase only is the good approach.

seheracer:
@srajama1 Do you prefer the two phase call for the tutorial example? @lucbv has an example in this PR for the new simplified call.

srajama1:
I am on the fence. I see @lucbv 's point. However, you give user's an easy interface that is what they would use and that is what everyone will compare against ! I would like them to learn the "correct" or high performance interface. It is just my style. What do you think ?

seheracer:
> I am on the fence. I see @lucbv 's point. However, you give user's an easy interface that is what they would use and that is what everyone will compare against ! I would like them to learn the "correct" or high performance interface. It is just my style. What do you think ?

I see your point. I think we should discuss this thoroughly in the next meeting.

lucbv:
@srajama1 @seheracer @brian-kelley 
I am adding a detailed discussion on why I am doing a combined symbolic+numeric and why I am not yet wrapping both these calls individually in issue #744 . It boils down to me now understanding the interface or detecting a design bug.

lucbv:
Thanks to PR #745 the new simplified interfaces are finalized and the corresponding wiki examples are added. Will run the spot check on this PR now but would like feedback on this @srajama1 .

lucbv:
## kokkos-dev2 --spot-check

PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=608 run_time=160
clang-8.0-Pthread_Serial-release build_time=202 run_time=108
clang-9.0.0-Pthread-release build_time=129 run_time=50
clang-9.0.0-Serial-release build_time=129 run_time=55
cuda-10.1-Cuda_OpenMP-release build_time=789 run_time=158
cuda-9.2-Cuda_Serial-release build_time=766 run_time=186
gcc-4.8.4-OpenMP-release build_time=113 run_time=58
gcc-7.3.0-OpenMP-release build_time=143 run_time=58
gcc-7.3.0-Pthread-release build_time=112 run_time=51
gcc-8.3.0-Serial-release build_time=141 run_time=57
gcc-9.1-OpenMP-release build_time=176 run_time=58
gcc-9.1-Serial-release build_time=171 run_time=56
intel-17.0.1-Serial-release build_time=252 run_time=56
intel-18.0.5-OpenMP-release build_time=353 run_time=53
intel-19.0.5-Pthread-release build_time=427 run_time=48

lucbv:
@srajama1 yes the documentation will be updated as soon as the PR merges (don't want to advertise function calls that do not exist yet x) )

lucbv:
@srajama1 
OK this is ready to merge now, I had a conflict since @brian-kelley and I are both adding new wiki examples.

lucbv:
Thanks @brian-kelley

src/sparse/KokkosSparse_spgemm_numeric.hpp
```diff
@@ -188,9 +188,9 @@ void spgemm_numeric(
       UniformDevice_t, //       typename bscalar_nnz_view_t_::device_type,
       Kokkos::MemoryTraits<Kokkos::Unmanaged> > Internal_bscalar_nnz_view_t_;
 
-  //static assert clno_row_view_t_ cannot be const type.
+  //static assert clno_row_view_t_ can be const type (row map is fixed after symbolic phase).
   typedef Kokkos::View<
-      typename clno_row_view_t_::non_const_value_type*,
+      typename clno_row_view_t_::value_type*,
```
ndellingwood, line 193: @lucbv @seheracer I think this change is impacting builds with mkl enabled, will post more

PR #751, opened 2020-06-24
Fix tplbuilds

ndellingwood:


ndellingwood:
This PR attempts to resolve a few compiler errors and -Werror (shadow warning with gcc-4.8.4) that showed up in the nightly tpl build.

The MKL change is a work-around (and maybe incorrect) that bypassed the compiler error, though I'm not sure how confident to be even though the tests passed.

The MKL error is because the `mkl_apply` routine modifies the row map, which conflicts with changes from https://github.com/kokkos/kokkos-kernels/pull/743
It's possible the offending assignment line in `mkl_apply` can be removed, though someone more familiar with the implementation should confirm/deny.

**Sample reproducer for kokkos-dev-2** (intel/17 here, but occurs with all intel versions and backends):

```bash
module load sems-env sems-cmake/3.12.2 sems-intel/17.0.1

$KOKKOSKERNELS_PATH/cm_generate_makefile.bash --with-devices=Serial --arch=SNB,Volta70 --compiler=/projects/sems/install/rhel7-x86_64/sems/compiler/intel/17.0.1/base/bin/icpc --cxxflags="-O3 -Wall -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wuninitialized " --cxxstandard="11" --ldflags=""  --kokkos-path=$KOKKOS_PATH --kokkoskernels-path=$KOKKOSKERNELS_PATH --with-scalars='double,complex_double' --with-ordinals=int --with-offsets=int,size_t --with-layouts=LayoutLeft --with-tpls=mkl    --with-options= --with-cuda-options= --no-examples
```

**Link to failing nightly**:
https://jenkins-son.sandia.gov/job/KokkosKernels_KokkosDev2_SPOTCHECK_TPLS/109/consoleFull

src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp
```diff
@@ -538,7 +538,13 @@ void mkl_symbolic(
 
 
           //KokkosKernels::Impl::copy_vector<MKL_INT *, typename cin_row_index_view_type::non_const_type, MyExecSpace> (m, rows_start, row_mapC);
-          idx nnz = row_mapC(m) =  rows_end[m - 1];
+          //idx nnz = row_mapC(m) = rows_end[m - 1];
+          idx nnz = rows_end[m - 1];
+          using ptr_type = decltype(row_mapC.data());
+          using rmptr_type = typename std::remove_pointer<ptr_type>::type;
+          using nonconst_rmptr_type = typename std::remove_const<rmptr_type>::type;
```
lucbv, line 545: This can be simplified using the typedefs defined in Kokkos::View.
I suggest doing the following:
```c++
using non_const_size_type = typename cin_row_index_view_type::non_const_value_type
auto* tmpPtr = const_cast<non_const_size_type*>(row_mapC.data() )
```
or is there something I missed?

src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp
```diff
@@ -627,7 +633,14 @@ void mkl_symbolic(
             Kokkos::Impl::Timer copy_time;
 
             //KokkosKernels::Impl::copy_vector<MKL_INT *, typename cin_row_index_view_type::non_const_type, MyExecSpace> (m, rows_start, row_mapC);
-            idx nnz = row_mapC(m) =  rows_end[m - 1];
+            //idx nnz = row_mapC(m) = rows_end[m - 1];
+            idx nnz = rows_end[m - 1];
+            using ptr_type = decltype(row_mapC.data());
+            using rmptr_type = typename std::remove_pointer<ptr_type>::type;
+            using nonconst_rmptr_type = typename std::remove_const<rmptr_type>::type;
+            auto * tmpPtr = const_cast<nonconst_rmptr_type*>(row_mapC.data());
```
lucbv, line 641: Same as above, I think we can simplify this logic a bit.

src/impl/tpls/KokkosBlas1_nrm2_tpl_spec_decl.hpp
```diff
@@ -91,8 +91,8 @@ Kokkos::View<const double*, LAYOUT, Kokkos::Device<ExecSpace, MEMSPACE>, \
     if (numElems < static_cast<size_type> (INT_MAX)) { \
       nrm2_print_specialization<RV,XV>(); \
       int N = numElems; \
-      int one = 1; \
-      R() = HostBlas<double>::nrm2(N,X.data(),one); \
+      int int_one = 1; \
```
lucbv, line 94: We could use `Kokkos::ArithTraits<int>::one()` here but it's a detail.
On the other hand I see above on line 91: `static_cast<size_type>(INT_MAX)`, this would probably benefit even more of `Kokkos::ArithTraits<size_type>::max()`

lucbv:
@ndellingwood 
I only have minor comments so if this PR needs to go as is now it's fine with me but if you have time it can be improved a little.

src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp
ndellingwood: @lucbv I originally did something like that, but then started overthinking with concern about what happens with things like FAD or ensemble types, is `non_const_value_type` the right type to use in all cases? I thought maybe `decltype` etc was a way to let the machine figure it out in each instance, but I'm not fully confident and happy to make the change to what you suggest if you prefer.

lucbv: This is the row_map so the `value_type` should be an integral type. Also on line 345 we throw if the values are not float or double so I think we are fine ; )

ndellingwood: > This is the row_map so the `value_type` should be an integral type

Ah, yep... good catch. Let me push the cleaner change.

ndellingwood:
@lucbv simplified the mkl stuff, running a tpl spot-check

ndellingwood:
kokkos-dev-2 tpl spot check:

```bash
PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=624 run_time=167
clang-8.0-Pthread_Serial-release build_time=209 run_time=129
clang-9.0.0-Pthread-release build_time=129 run_time=52
clang-9.0.0-Serial-release build_time=139 run_time=57
cuda-10.1-Cuda_OpenMP-release build_time=833 run_time=164
gcc-4.8.4-OpenMP-release build_time=116 run_time=59
gcc-7.3.0-OpenMP-release build_time=141 run_time=60
gcc-7.3.0-Pthread-release build_time=116 run_time=53
gcc-8.3.0-Serial-release build_time=139 run_time=59
gcc-9.1-OpenMP-release build_time=192 run_time=60
gcc-9.1-Serial-release build_time=167 run_time=59
intel-17.0.1-Serial-release build_time=250 run_time=55
intel-18.0.5-OpenMP-release build_time=358 run_time=52
intel-19.0.5-Pthread-release build_time=442 run_time=48
```

src/impl/tpls/KokkosBlas1_nrm2_tpl_spec_decl.hpp
ndellingwood: This is a good point, I decided to hold off and approach more systematically later (there will be more occurrences) when time isn't crunched.

lucbv: Agreed, thanks for making this PR to fix the issues with mkl : )

PR #770, opened 2020-07-23
cm_test_all_sandia: increase timeout

ndellingwood:


ndellingwood:
A couple nightlies are intermittently timing out in the sparse unit tests, this PR just increases the time before timeout.

Issue #694, opened 2020-04-16
SpAdd requires output values to be zero-initialized, but this shouldn't be needed


brian-kelley:
Currently, the values of the sum are computed by adding A and B values into the C values:
```
Cvals(i1) += Avals(j);
Cvals(i2) += Bvals(j);
```
but it is possible to change the loop slightly to compute each C value and then write it out:
```
scalar_t sum = 0;
sum += Avals(j);
sum += Bvals(j);
Cvals(i) = sum;
```
Then the user wouldn't have to initialize the C values view.

Issue #685, opened 2020-04-10
SpAdd doesn't merge entries correctly


brian-kelley:
Neither the sorted nor unsorted versions of SpAdd will merge entries if the input is unmerged (has duplicate entries in a row). It produces the correct sum, but it should be able to merge any number of duplicate entries for no performance penalty.

PR #773, opened 2020-07-29
Cleanup spadd (#685, #694)

brian-kelley:
- Remove "max_nnz_in_result" and its getter/setter from SpADD handle. It was really a duplicate of "c_nnz". Max is a bad name for this since it is not an upper bound, it is always exact. This will need to be propagated to Tpetra_MatrixMatrix_def.hpp when we do promotion.
- Remove team size/vector length stuff from SpADD handle, since SpADD only uses range policy. This was just copied from SpGEMM handle without needing it
- No longer require C's scalar values to be zero-initialized when they are passed in (#694)
- If A and/or B have unmerged rows (aka repeated column in a row), the resulting C is now always merged. Before, it would produce the correct sum but could only merge one entry from A and one entry from B. Now, it can merge an arbitrary number of entries from A and B (#685)
- Improve test coverage: test with unmerged inputs (both sorted and unsorted). The test now checks that the result is sorted and merged. Also add a test for a matrix with nonzero dimensions but zero entries.

Note that these are all just enhancements, not bugfixes. I noticed these while trying to track down an EMPIRE issue, where add's MergeEntriesFunctor is crashing after my recent cuSPARSE TPL changes. I think the underlying cause of the bug is somewhere else since cuSPARSE SpMV is completely unrelated to add.

Testing:
PASSED TESTS
clang-8.0-Pthread_Serial-release build_time=336 run_time=173
clang-9.0.0-Pthread-release build_time=136 run_time=77
clang-9.0.0-Serial-release build_time=174 run_time=64
cuda-10.1-Cuda_OpenMP-release build_time=848 run_time=167
cuda-9.2-Cuda_Serial-release build_time=794 run_time=210
gcc-4.8.4-OpenMP-release build_time=117 run_time=64
gcc-7.3.0-OpenMP-release build_time=179 run_time=2532
gcc-7.3.0-Pthread-release build_time=132 run_time=80
gcc-8.3.0-Serial-release build_time=206 run_time=66
gcc-9.1-OpenMP-release build_time=190 run_time=392
gcc-9.1-Serial-release build_time=175 run_time=67
intel-17.0.1-Serial-release build_time=263 run_time=59
intel-18.0.5-OpenMP-release build_time=425 run_time=67
intel-19.0.5-Pthread-release build_time=467 run_time=72
clang-8.0-Cuda_OpenMP-release (test failed) <== sparse_openmp timed out due to machine congestion, but I re-ran it today from the same build and it passed

### DIFF
--- a/src/sparse/KokkosSparse_spadd.hpp
+++ b/src/sparse/KokkosSparse_spadd.hpp
@@ -47,6 +47,7 @@
 
 #include "KokkosKernels_Handle.hpp"
 #include "KokkosKernels_Sorting.hpp"
+#include <limits>
 
 namespace KokkosSparse {
 namespace Experimental {
@@ -84,7 +85,11 @@ struct SortedCountEntries {
         Browptrs(Browptrs_),
         Bcolinds(Bcolinds_),
         Crowcounts(Crowcounts_) {}
-  KOKKOS_INLINE_FUNCTION void operator()(const ordinal_type i) const {
+
+  static constexpr ordinal_type ORDINAL_MAX = std::numeric_limits<ordinal_type>::max();
+
+  KOKKOS_INLINE_FUNCTION void operator()(const ordinal_type i) const
+  {
     // count the union of nonzeros in Arow and Brow
     size_type numEntries = 0;
     size_type ai         = 0;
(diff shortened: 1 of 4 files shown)


src/sparse/KokkosSparse_spadd.hpp
```diff
@@ -489,48 +499,49 @@ struct SortedNumericSumFunctor {
...
+    size_type Browlen    = Browptrs(i + 1) - Browstart;
+    ordinal_type Acol = (Arowlen == 0) ? ORDINAL_MAX : Acolinds(Arowstart);
+    ordinal_type Bcol = (Browlen == 0) ? ORDINAL_MAX : Bcolinds(Browstart);
+    size_type Coffset = Crowptrs(i);
+    while (Acol != ORDINAL_MAX || Bcol != ORDINAL_MAX)
+    {
+      ordinal_type Ccol = (Acol < Bcol) ? Acol : Bcol;
+      //Eat all entries in both A and B which have this column
+      //This also results in Acol/Bcol being updated to following entries for next loop iter
+      CscalarT accum = {0};
```
lucbv, line 520: I would have thought that `CscalarT accumulation = Kokkos::ArithTraits<CscalarT>::zero();` is the correct way to initialize `accum` to handle various complex types and Stokhos defines specializations for UQ_PCE and MP_VECTORS. Does this version work OK for exotic types?

brian-kelley: @lucbv Good point, I agree that I should use ArithTraits here.

brian-kelley:
Still good after making the 1-line change:

PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=658 run_time=170
clang-8.0-Pthread_Serial-release build_time=210 run_time=141
clang-9.0.0-Pthread-release build_time=148 run_time=71
clang-9.0.0-Serial-release build_time=169 run_time=63
cuda-10.1-Cuda_OpenMP-release build_time=942 run_time=186
cuda-9.2-Cuda_Serial-release build_time=901 run_time=212
gcc-4.8.4-OpenMP-release build_time=122 run_time=67
gcc-7.3.0-OpenMP-release build_time=158 run_time=113
gcc-7.3.0-Pthread-release build_time=185 run_time=76
gcc-8.3.0-Serial-release build_time=146 run_time=63
gcc-9.1-OpenMP-release build_time=215 run_time=151
gcc-9.1-Serial-release build_time=168 run_time=61
intel-17.0.1-Serial-release build_time=255 run_time=58
intel-18.0.5-OpenMP-release build_time=353 run_time=60
intel-19.0.5-Pthread-release build_time=452 run_time=67

Issue #777, opened 2020-07-30
cuBLAS Cmake logic wrong?


lucbv:
@jjwilke @ndellingwood 
While working with @cgcgcg on issue #775 on Lassen, I noticed that `cm_generate_makefile.bash` does not lead to a working configuration if I enable cuBLAS: `--with-tpls="cublas,cusparse"`.
If I enable cuSPARSE only things are working fine.
Simply modifying the CMakeLogic in `Cmake/Modules/FindTPLCUBLAS.cmake` to mirror what happens in the corresponding file for cuSPARSE resolves the issue.
I will open a PR to propose that change.

ndellingwood:
Separate from the cmake logic aspects of this issue, the cm_generate_makefile requires also adding host-blas to the tpl list to get a working configuration when using cublas, so the script needs updating with better output messaging to indicate this; attempting to automatically enable a host blas tpl would be problematic on machines that do not have an install on the system path (for example our white testbed, where this issue is handled separately in our testing scripts), so I don't think that would be a good option. I'll put in a PR to update the script, sorry for the frustrations.

Linking comment here regarding the script issue for reference https://github.com/kokkos/kokkos-kernels/pull/778#issuecomment-666513890

PR #778, opened 2020-07-30
CMake: find cublas

lucbv:
As explained in issue #777 I am having issue configuring with cuBLAS TPL support on Lassen.
The proposed changes are fixing my issues.
@jjwilke is there a reason to have different CMake logic for `cuBLAS` and `cuSPARSE`?
I am also a bit curious as to how the script `cm_test_all_sandia --spot-check-tpls` works, since it relies on the same `cm_generate_makefile.bash`?
I am running tests with TPLs on Kokkos-dev2, I'll check the configure output when it gets to the Cuda build.


lucbv:
Looks like I might need to rebase this on develop first?

jjwilke:
What do you mean by "different logic" for cusparse and cublas? Separate find modules - or?

jjwilke:
Changes seem fine - after rebase.

lucbv:
@jjwilke damn, we have a branch called `devlop` which makes it easy to mis-click when targeting `develop`...
I will suggest that we remove it for safety's sake.

Regarding the Cmake logic, I was referring to the fact the cuBLAS was not listed as a dependency and the following line:
```
KOKKOSKERNELS_CREATE_IMPORTED_TPL(CUBLAS INTERFACE
      LINK_LIBRARIES "${CUDA_CUBLAS_LIBRARIES}")
```
which I replaced with:
```
KOKKOSKERNELS_CREATE_IMPORTED_TPL(CUBLAS LIBRARY ${CUDA_cublas_LIBRARY})
```
I am not sure what the `INTERFACE LINK_LIBRARY` means and how it differs from the `LIBRARY` statement.

ndellingwood:
> I am also a bit curious as to how the script `cm_test_all_sandia --spot-check-tpls` works

@lucbv in Cuda builds this option adds cublas and cusparse to the tpl build list in addition to host-blas

lucbv:
@ndellingwood OK, but when I only have cublas and cusparse I get a configuration error, I am not exactly sure why adding host-blas fixes that situation? I am running spot-check-tpls without my modification to see what the configuration command line looks like.

ndellingwood:
@lucbv this worked for me with cm_generate_makefile on kokkos-dev-2:

```bash
module load sems-cmake/3.12.2 sems-cuda/10.1 sems-gcc/7.3.0

../../cm_generate_makefile.bash --with-cuda --with-tpls=blas,cublas,cusparse --arch=Volta70 --compiler=$HOME/kokkos/bin/nvcc_wrapper --kokkos-path=$HOME/kokkos --kokkoskernels-path=$HOME/kokkos-kernels
```

The cmake config picked up cublas, cusparse, blas, and lapack:

```bash
KokkosKernels ETI Types
   Devices:  <Cuda,CudaSpace>;<Cuda,CudaUVMSpace>;<Serial,HostSpace>
   Scalars:  double
   Ordinals: int
   Offsets:  int;size_t
   Layouts:  LayoutLeft

KokkosKernels TPLs
   BLAS:        /usr/lib64/libblas.so
   LAPACK:      /usr/lib64/liblapack.so;/usr/lib64/libblas.so
   CUBLAS:      /projects/sems/install/rhel7-x86_64/sems/compiler/cuda/10.1/base/lib64/libcublas.so
   CUSPARSE:    /projects/sems/install/rhel7-x86_64/sems/compiler/cuda/10.1/base/lib64/libcusparse.so
```

I think the need for enabling the host-blas with cublas comes from tpl layer design in kokkos-kernels (though I may be misremembering), I can look back at this to give a better answer. Script-wise, we can update them to give a better message that it is required to enable host-blas when using cublas, sorry for the irritation running into this.

lucbv:
@ndellingwood with the current `FindTPLCUBLAS` the variable `CUDA_CUBLAS_LIBRARIES` is undefined in my Lassen build whereas `CUDA_cublas_LIBRARY`. If I grep kokkos-kernels for the current variable here is what I get:
```
/home/lberge/kk_lucbv/scripts/cm_test_all_sandia:562:    # "Could NOT find TPLCUBLAS (missing: CUDA_CUBLAS_LIBRARIES)"
/home/lberge/kk_lucbv/cmake/Modules/FindTPLCUBLAS.cmake:12:  IF("${CUDA_CUBLAS_LIBRARIES}" MATCHES "NOTFOUND")
/home/lberge/kk_lucbv/cmake/Modules/FindTPLCUBLAS.cmake:13:    FIND_PACKAGE_HANDLE_STANDARD_ARGS(TPLCUBLAS REQUIRED_VARS CUDA_CUBLAS_LIBRARIES)
/home/lberge/kk_lucbv/cmake/Modules/FindTPLCUBLAS.cmake:16:      LINK_LIBRARIES "${CUDA_CUBLAS_LIBRARIES}")
```
So I am wondering how this variable is defined? Based on that I might be able to understand why it is called differently on Lassen.

ndellingwood:
@lucbv can you post the modules and environment info from your build on Lassen, like cuda version, host compiler + version, and cmake version? I wonder if some older/newer versions of cmake have changes in the cublas module and different case sensitivity for spelling of `CUDA_cublas_LIBRARY`?

lucbv:
Sure here is what I load:
```
module load cmake/3.12.1
module load gcc/7.3.1
module load cuda/10.1.243
module load spectrum-mpi/rolling-release
```
I think this is fairly close to what we use for the kokkos-dev2 spot-check-tpls:
```
modules=sems-env sems-cmake/3.12.2 sems-cuda/10.1 sems-gcc/7.3.0
```

lucbv:
@ndellingwood I can see the following defined in the Cmake version I use on Lassen:
```
set(CUDA_CUBLAS_LIBRARIES ${CUDA_cublas_LIBRARY} ${CUDA_cublas_device_LIBRARY})
```
so it seems that both variables (`CUDA_CUBLAS_LIBRARIES` and `CUDA_cublas_LIBRARY`) are available, although the first one since to concatenate two variables, could that be an issue?

jjwilke:
Sorry, should have been tracking this better today. Distracted by our building move.

The official FindCUDA documentationl lists CUDA_CUBLAS_LIBRARIES as the variable name (and as you figured out if you read the actual FindCUDA.cmake file from CMake that's where it gets set).

I looked through the issues and above. Did you post the actual error somewhere? I'm struggling a bit to understand exactly how this is breaking.

If there is a weird dependence on host blas, I can look at adding some dependency tracking to the build.

jjwilke:
I can also hop on a skype call if you want to actually talk it through.

PR #782, opened 2020-08-05
Add new SpGEMM interfaces to the unit tests

seheracer:
This PR adds the new SpGEMM interfaces to the unit tests in addition to the old interfaces. The new interfaces take three CrsMatrix objects as input while the old ones take up to nine Kokkos Views instead.

This integration required adding new ETI stuff for numeric, because the old numeric interface requires a **non-const** row_mapC, whereas the new interface requires a **const** row_mapC.

SpGEMM tests are passing on kokkos-dev-2 and white.

### DIFF
--- a/unit_test/sparse/Test_Sparse_spgemm.hpp
+++ b/unit_test/sparse/Test_Sparse_spgemm.hpp
@@ -80,7 +80,31 @@ using namespace KokkosKernels::Experimental;
 namespace Test {
 
 template <typename crsMat_t, typename device>
-int run_spgemm(crsMat_t input_mat, crsMat_t input_mat2, KokkosSparse::SPGEMMAlgorithm spgemm_algorithm, crsMat_t &result) {
+int run_spgemm(crsMat_t A, crsMat_t B, KokkosSparse::SPGEMMAlgorithm spgemm_algorithm, crsMat_t &C) {
+
+  typedef typename crsMat_t::size_type size_type;
+  typedef typename crsMat_t::ordinal_type lno_t;
+  typedef typename crsMat_t::value_type scalar_t;
+
+  typedef KokkosKernels::Experimental::KokkosKernelsHandle
+      <size_type, lno_t, scalar_t,
+      typename device::execution_space, typename device::memory_space,typename device::memory_space > KernelHandle;
+
+  KernelHandle kh;
+  kh.set_team_work_size(16);
+  kh.set_dynamic_scheduling(true);
+
+  kh.create_spgemm_handle(spgemm_algorithm);
+
+  KokkosSparse::spgemm_symbolic(kh, A, false, B, false, C);
+  KokkosSparse::spgemm_numeric(kh, A, false, B, false, C);
+  kh.destroy_spgemm_handle();
+
+  return 0;
+}
+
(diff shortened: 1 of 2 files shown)


lucbv:
@ndellingwood this is good to merge in my opinion.
@seheracer is that fine with you?

seheracer:
> @ndellingwood this is good to merge in my opinion.
> @seheracer is that fine with you?

Sure :)

srajama1:
Can you hod please ? I would like to see the new i/f.

ndellingwood:
Sure, @srajama1 can you go ahead and merge when you finish reviewing?

unit_test/sparse/Test_Sparse_spgemm.hpp
```diff
@@ -351,7 +378,10 @@ void test_spgemm(lno_t numRows, size_type nnz, lno_t bandwidth, lno_t row_size_v
     bool failed = false;
     int res = 0;
     try{
-      res = run_spgemm<crsMat_t, device>(input_mat, input_mat, spgemm_algorithm, output_mat);
+      if(oldInterface)
+	res = run_spgemm_old_interface<crsMat_t, device>(input_mat, input_mat, spgemm_algorithm, output_mat);
```
ndellingwood, line 382: @seheracer some compilers will complain (with `-Werror`) about the indentation, can you add a couple spaces here so it is not flucsh with the `if` statement?

unit_test/sparse/Test_Sparse_spgemm.hpp
```diff
@@ -351,7 +378,10 @@ void test_spgemm(lno_t numRows, size_type nnz, lno_t bandwidth, lno_t row_size_v
...
+      else
+	res = run_spgemm<crsMat_t, device>(input_mat, input_mat, spgemm_algorithm, output_mat);
```
ndellingwood, line 384: @seheracer some compilers will complain (with `-Werror`) about the indentation, can you add a couple spaces here so it is not flucsh with the `if` statement

unit_test/sparse/Test_Sparse_spgemm.hpp
seheracer: It looks OK in: https://github.com/seheracer/kokkos-kernels/blob/f923b410d37c77d21a6bc415e67a6fe9c02cea2c/unit_test/sparse/Test_Sparse_spgemm.hpp#L381

Idk why it looks unindented here.

unit_test/sparse/Test_Sparse_spgemm.hpp
seheracer: Same as above

seheracer:
> Can you hod please ? I would like to see the new i/f.

FYI, these are the new interfaces @lucbv wrote right before kokkos tutorials, so they don't involve pure numeric or pure symbolic yet.

lucbv:
@srajama1 this should really be a quick review, the only real change was made to the ETI system for spgemm to allow the row_map of the C matrix to be either const or non_const when calling the numeric phase.
There is also a small addition to the unit_tests but that should be no worry : )

srajama1:
Oh, I thought this was a big thing. Never mind. I will merge it.

PR #790, opened 2020-08-20
cm_test_all_sandia: update testing for c++14, remove old compilers

ndellingwood:
add workaround for intel/17 modules


ndellingwood:
Kokkos will require c++14 after 3.2.00 release. This updates testing scripts to set c++14 as the default standard and removes old unsupported compilers, aligned with kokkos/kokkos#3291 and kokkos/kokkos#3295

PR #801, opened 2020-08-31
Distance-2 maximal independent set.

brian-kelley:
- ``#include "KokkosGraph_MIS2.hpp"``
- Interface: ``auto mis2 = KokkosGraph::Experimental::graph_d2_mis(rowmap, entries, algo);``, and algo defaults to ``KokkosGraph::MIS2_FAST``. Other choice is ``KokkosGraph::MIS2_QUALITY``. This is much slower but does give a larger MIS. No handle is involved.
- Fast version (based on Luby) and high-quality version (based on ECL)
- Unit test, perf test
- Fast version yields roughly 2-5x speedup vs. CUSP MIS2 on a P100 (CUSP is the only other implementation I know of), and gives a similar sized or slightly bigger MIS
- Bitonic sorting functions now take a comparator closure. Not currently used by MIS2 but may be in the future. Backward compatible interface.
- Temporarily disabled the choice of coarsening algorithm on Cluster Gauss-Seidel. Now the only option is Balloon, but soon a MIS2-based coarsening will be added (much faster, but doesn't allow for choosing cluster size). Not testing the other bad coarsening options really speeds up the cluster GS unit test.

Mike Gilbert is currently waiting to use this for some graph coarsening work.
```
PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=615 run_time=118
clang-8.0-Pthread_Serial-release build_time=200 run_time=102
clang-9.0.0-Pthread-release build_time=127 run_time=50
clang-9.0.0-Serial-release build_time=127 run_time=38
cuda-10.1-Cuda_OpenMP-release build_time=806 run_time=119
cuda-9.2-Cuda_Serial-release build_time=784 run_time=168
gcc-4.8.4-OpenMP-release build_time=117 run_time=40
gcc-7.3.0-OpenMP-release build_time=146 run_time=38
gcc-7.3.0-Pthread-release build_time=115 run_time=48
gcc-8.3.0-Serial-release build_time=141 run_time=41
gcc-9.1-OpenMP-release build_time=178 run_time=38
gcc-9.1-Serial-release build_time=158 run_time=39
intel-17.0.1-Serial-release build_time=253 run_time=43
intel-18.0.5-OpenMP-release build_time=358 run_time=39
intel-19.0.5-Pthread-release build_time=401 run_time=48
```


brian-kelley:
Some quick performance results for the fast algorithm so far:

| Matrix                       | KK Time     | CUSP Time  | KK Speedup | KK MIS Size | CUSP MIS Size |
|------------------------------|-------------|------------|------------|-------------|---------------|
| cfd1                         | 0.00189865  | 0.00529712 | 2.8x       | 1604        | 1611          |
| bodyy5                       | 0.000801862 | 0.00197338 | 2.46x      | 1797        | 4257          |
| msc04515                     | 0.000855407 | 0.00165642 | 1.94x      | 155         | 448           |
| audikw_1                     | 0.0469884   | 0.0429685  | 0.91x      | 4308        | 13585         |
| af_shell7                    | 0.0109693   | 0.0245758  | 2.24x      | 10009       | 37024         |
| McAlister Momentum (3M rows) | 0.0160757   | 0.0743423  | 4.62x      | 280684      | 277066        |

It seems that for graphs with fairly regular structure, my algo is faster but also a similar quality. For others, it's still faster but also lower quality. I will see if I can use degree information to improve quality without increasing time too much.

brian-kelley:
Another update: I wrote a standalone utility to verify a distance-2 MIS, and I don't think CUSP is producing a valid output for those 4 matrices where it seemingly got a much higher quality. In all cases, this verification finds many pairs of vertices in the MIS that are within 2 hops of each other. My MIS output passes this verification (it tests that no vertices could be added to the set, and that no conflicts exist).3

brian-kelley:
Third update: I got an alternate implementation of the algorithm from Amg-X/CUSP from Mike Gilbert, and it produces a correct and similar sized MIS to my impl on those 4 matrices. I guess CUSP has a bug in MIS.

src/graph/KokkosGraph_MIS2.hpp
```diff
@@ -0,0 +1,93 @@
...
+#ifndef _KOKKOSGRAPH_DISTANCE2_MIS_HPP
+#define _KOKKOSGRAPH_DISTANCE2_MIS_HPP
+
+#include "KokkosGraph_Distance2MIS_impl.hpp"
+
+namespace KokkosGraph{
+
+enum MIS2_Algorithm
+{
+  MIS2_QUALITY,
```
srajama1, line 54: There are two algorithms listed here, but only one above.

perf_test/graph/KokkosGraph_mis_d2.cpp
```diff
@@ -0,0 +1,381 @@
...
+struct MIS2Parameters
+{
+  int repeat = 1;
+  bool verbose = false;
+  int use_threads = 0;
+  int use_openmp = 0;
+  int use_cuda = 0;
+  int use_serial = 0;
+  const char* mtx_file = NULL;
+  MIS2_Algorithm algo = MIS2_FAST;
```
srajama1, line 80: MIS2_QUALITY is not an option ?

src/graph/impl/KokkosGraph_Distance2MIS_impl.hpp
```diff
@@ -0,0 +1,737 @@
...
+      {
+        lno_t nei = (j == rowEnd) ? i : entries(j);
+        if(nei <= nv)
+        {
+          status_t neiStat = rowStatus(nei);
+          if(neiStat < s)
+            s = neiStat;
+        }
+      }
+      if(s == IN_SET)
```
srajama1, line 190: Doesn't this need a team barrier here ? s could be set by some thread right ?

srajama1: same comment applies in other places below but this may not matter just making sure.

brian-kelley: Everything is RangePolicy currently, so s is local to a single CUDA thread.

perf_test/graph/KokkosGraph_mis_d2.cpp
brian-kelley: The table above is the MIS2_FAST algorithm. You're right, I should add MIS2_QUALITY as an option for the perf test, if I keep it at all (but it's way slower in exchange for not much increase in the mis size). Both are deterministic. I'll give some numbers from the graphs above to compare FAST against QUALITY.

brian-kelley:
@srajama1 Here is data on FAST vs. QUALITY, as well as a hybrid priority function I tried today. Times are averaged over 100 runs.

| Matrix                       | MIS2_FAST Time | MIS2_HYBRID Time | MIS2_QUALITY Time | FAST Set Size | HYBRID Set Size | QUALITY Set Size |
|------------------------------|----------------|------------------|-------------------|---------------|-----------------|------------------|
| cfd1                         | 0.00189642     | 0.00212453       | 0.0255732         | 1604          | 1684            | 2273             |
| bodyy5                       | 0.000767634    | 0.000772913      | 0.0212677         | 1797          | 1903            | 2099             |
| msc04515                     | 0.000801476    | 0.000823509      | 0.00128827        | 155           | 167             | 194              |
| audikw_1                     | 0.0426331      | 0.0404596        | 0.319517          | 4308          | 4469            | 5757             |
| af_shell7                    | 0.00974024     | 0.00902756       | 0.40871           | 10009         | 9934            | 12905            |
| McAlister Momentum (3M rows) | 0.0142756      | 0.015179         | 1.07994           | 280684        | 283075          | 345105           |

brian-kelley:
It seems like there is a continuum of similar algorithms that trade off between iteration count and quality. The more the priority values change between rounds (like Luby), the faster it converges, but it also means high degree nodes might get in the set. If the priorities never change (like ECL), then it converges much more slowly but yields a higher quality.

brian-kelley:
The "Hybrid" set the highest bit of priority based on whether degree was larger than average, and the rest was the same as FAST.

srajama1:
I like having FAST and QUALITY. I am note sure hybrid buys us much.

brian-kelley:
@srajama1 Here's updated testing for the Gauss-Seidel support and the new coarsening test/example.

brian-kelley:
@srajama1 I just added team policy support for when exec_space is Cuda.

| Matrix            | Range      | Team       | Team Speedup |
|-------------------|------------|------------|--------------|
| McAlister Momentum | 0.0104202  | 0.0119231  | 0.873950567  |
| af_shell7         | 0.00646672 | 0.00440477 | 1.468117518  |
| audikw_1          | 0.0234191  | 0.00842605 | 2.779368743  |
| msc04515          | 0.00064653 | 0.00059379 | 1.088828448  |
| StocF-1465        | 0.00652891 | 0.00757888 | 0.86146106   |
| bodyy5            | 0.00062369 | 0.00061153 | 1.019881216  |

Geometric mean of speedup factor: 1.22693935

There's a clear positive correlation between the speedup factor and the average degree. For the best speedup, the cutoff is about 16 (below that, use range policy, and above use teams). With that:

| Matrix            | Range      | Heuristic  | Speedup vs. Range |
|-------------------|------------|------------|-------------------|
| McAlister Momentu | 0.0104202  | 0.0104894  | 0.993402864       |
| af_shell7         | 0.00646672 | 0.00437513 | 1.478063509       |
| audikw_1          | 0.0234191  | 0.00835315 | 2.80362498        |
| msc04515          | 0.00064653 | 0.00057937 | 1.115922868       |
| StocF-1465        | 0.00652891 | 0.00658136 | 0.992030523       |
| bodyy5            | 0.00062369 | 0.00062067 | 1.004872186       |

(there's about +/- 1% noise in the timings since Momentum, StocF and bodyy5 are still being run with RangePolicy). This gives a geomean 29% speedup over RangePolicy, so that's what I'm going to keep. I have to re run spot checks though.

srajama1:
@brian-kelley If you are adding team policy, there were a few places that required a team barrier. I assume you have introduced it.

brian-kelley:
@srajama1 I don't need any team-wide synchronization here, since I'm treating only using vector-level parallelism. I do use ``Kokkos::single(Kokkos::PerThread, ...)`` for applying side effects from each thread, but I don't think anything else is needed.

brian-kelley:
Newest spot check:

PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=636 run_time=135
clang-8.0-Pthread_Serial-release build_time=204 run_time=114
clang-9.0.0-Pthread-release build_time=121 run_time=53
clang-9.0.0-Serial-release build_time=127 run_time=45
cuda-10.1-Cuda_OpenMP-release build_time=848 run_time=135
cuda-9.2-Cuda_Serial-release build_time=820 run_time=188
gcc-7.3.0-OpenMP-release build_time=141 run_time=44
gcc-7.3.0-Pthread-release build_time=120 run_time=55
gcc-8.3.0-Serial-release build_time=144 run_time=48
gcc-9.1-OpenMP-release build_time=180 run_time=44
gcc-9.1-Serial-release build_time=161 run_time=46
intel-17.0.1-Serial-release build_time=280 run_time=53
intel-18.0.5-OpenMP-release build_time=363 run_time=48
intel-19.0.5-Pthread-release build_time=425 run_time=57

srajama1:
@brian-kelley : ok, thanks !  Audi was the only one that had some trouble with CUSP before. I assume now the new algorithm is faster in all test cases used above. Let me know when it is ready, I will approve it.

brian-kelley:
@srajama1 CUSP's output failed my validation for audikw, af_shell, bodyy, and msc04515. But yes, the new algorithm is never slower than RangePolicy version for the matrices above since I used them to choose the heuristic.

It's ready to merge I think.

src/graph/impl/KokkosGraph_Distance2MIS_impl.hpp
```diff
@@ -396,7 +396,7 @@ struct D2_MIS_RandomPriority
     Kokkos::parallel_for(range_pol(0, numVerts), InitWorklistFunctor(colWorklist));
     worklist_t thirdWorklist = Kokkos::subview(allWorklists, Kokkos::ALL(), 2);
     auto execSpaceEnum = KokkosKernels::Impl::kk_get_exec_space_type<exec_space>();
-    bool useTeams = execSpaceEnum == KokkosKernels::Impl::Exec_CUDA;
+    bool useTeams = (execSpaceEnum == KokkosKernels::Impl::Exec_CUDA) && (entries.extent(0) / numVerts >= 16);
```
srajama1, line 399: Average degree >= 16 ? Is this what you want ? It will not use the team policy for several use cases such as problems with 7 nnz per row.

src/graph/impl/KokkosGraph_Distance2MIS_impl.hpp
```diff
@@ -405,14 +405,15 @@ struct D2_MIS_RandomPriority
     int decideSetTeamSize = 0;
     if(useTeams)
     {
+      team_pol temp(1, 1, vectorLength);
```
srajama1, line 408: may need to find a better name than temp ?

srajama1:
> @srajama1 CUSP's output failed my validation for audikw, af_shell, bodyy, and msc04515. But yes, the new algorithm is never slower than RangePolicy version for the matrices above since I used them to choose the heuristic.

Yeah, you told me. my bad. This is good.

src/graph/impl/KokkosGraph_Distance2MIS_impl.hpp
brian-kelley: @srajama1 I believe that is what we want, at least based on the graphs in the table. bodyy, StocF and McAlister Momentum were the graphs in the table with < 16 nnz/row. Using teams was 2% faster, 16% slower, and 15% slower respectively. So overall using RangePolicy for those is faster.

src/graph/impl/KokkosGraph_Distance2MIS_impl.hpp
brian-kelley: @srajama1 I just renamed it to ``dummyPolicy``.

src/graph/impl/KokkosGraph_Distance2MIS_impl.hpp
srajama1: Alright. Thanks for checking !

srajama1 (approved):
Thanks @brian-kelley !

srajama1:
Can you add the numbers to wiki for reproducibility ?

srajama1:
Oh, and create a wiki page for the feature itself.

Issue #807, opened 2020-09-12
HIP: sqrt math function needs to be guarded


lucbv:
The current code in Kokkos_ArithTraits.hpp does not build with hipcc because of the `long double` specialization of some math operations, as initially reported by Junchao Zhang on [slack](https://kokkosteam.slack.com/archives/CMZDYKZ96/p1599755841049900).
This should be fixed as part of the work HIP integration, see #806

PR #809, opened 2020-09-14
HIP: fix ArithTraits to support HIP backend, see issue #807

lucbv:
The goal is to replace undefined calls to math functions with the HIP backend.
Hopefully the pre-processor logic makes sense?

### DIFF
--- a/src/Kokkos_ArithTraits.hpp
+++ b/src/Kokkos_ArithTraits.hpp
@@ -691,13 +691,13 @@ public:
   static KOKKOS_FORCEINLINE_FUNCTION float infinity() { return HUGE_VALF; }
 
   static KOKKOS_FORCEINLINE_FUNCTION bool isInf (const float x) {
-    #ifndef __CUDA_ARCH__
+    #ifdef KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST
     using std::isinf;
     #endif
     return isinf (x);
   }
   static KOKKOS_FORCEINLINE_FUNCTION bool isNan (const float x) {
-    #ifndef __CUDA_ARCH__
+    #ifdef KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST
     using std::isnan;
     #endif
     return isnan (x);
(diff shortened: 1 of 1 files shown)


src/Kokkos_ArithTraits.hpp
```diff
@@ -1126,9 +1128,11 @@ class ArithTraits<double> {
     return ::atan (x);
   }
   static KOKKOS_FORCEINLINE_FUNCTION val_type nan () {
-#ifdef __CUDA_ARCH__
+#if defined(__CUDA_ARCH__)
     return CUDART_NAN;
     //return nan (); // this returns 0 ???
+#elif defined(__HIP_DEVICE_COMPILE__)
+    return nan("");
```
dalg24, line 1135: I get
```
<kokkos-kernels>/src/Kokkos_ArithTraits.hpp:1135:12: error: too many arguments to function call, expected 0, have 1; did you mean '::nan'?
    return nan("");
           ^~~
           ::nan
/opt/rocm-3.5.0/llvm/lib/clang/11.0.0/include/__clang_cuda_math_forward_declares.h:141:19: note: '::nan' declared here
__DEVICE__ double nan(const char *);
```

dalg24:
```suggestion
    return std::nan("");
```

lucbv:
OK got the PR is building correctly on `caraway` and I checked that the spot-check runs fine on kokkos-dev2.
I think this is good for a final check before merging, @srajama1 any thoughts?

srajama1 (approved):
The changes look benign to me. However, I would like feedback from someone who has experience with this software stack. @dalg24 already commented on this PR, if he approves go for it.

src/Kokkos_ArithTraits.hpp
```diff
@@ -1132,7 +1132,7 @@ class ArithTraits<double> {
     return CUDART_NAN;
     //return nan (); // this returns 0 ???
 #elif defined(__HIP_DEVICE_COMPILE__)
-    return nan("");
+    return ::nan("");
```
dalg24, line 1135: Nope you need `std::`

dalg24 (approved):

srajama1:
Merging it. Thanks @lucbv !

PR #818, opened 2020-09-25
cm_generate_makefile: add hip and amd gpu support

ndellingwood:


ndellingwood:
@lucbv do you have a moment to take a look at this PR? Adds support for HIP and AMD GPUs

cm_generate_makefile.bash
```diff
@@ -59,6 +69,24 @@ get_kokkos_cuda_option_list() {
...
+  PARSE_HIP_LST=$(echo $KOKKOS_HIP_OPTIONS | tr "," "\n")
+  for HIP_ in $PARSE_HIP_LST
+  do
+     HIP_OPT_NAME=
+     if  [ "${HIP_}" == "rdc" ]; then
+        HIP_OPT_NAME=HIP_RELOCATABLE_DEVICE_CODE
+     else
+        echo "${HIP_} is not a valid hip option..."
+     fi
+     if [ "${HIP_OPT_NAME}" != "" ]; then
```
lucbv, line 84: What happens if I want to enable HIP without any particular options, i.e. just "-D Kokkos_ENABLE_HIP:BOOL=ON", will "${HIP_OPT_NAME}" be empty and `KOKKOS_HIP_OPTION_CMD` stay empty?

cm_generate_makefile.bash
```diff
@@ -59,6 +69,24 @@ get_kokkos_cuda_option_list() {
...
+        KOKKOS_HIP_OPTION_CMD="-DKokkos_ENABLE_${HIP_OPT_NAME}=ON ${KOKKOS_HIP_OPTION_CMD}"
```
lucbv, line 85: Does this add `-DKokkos_ENABLE_${HIP_OPT_NAME}=ON` at the beginning every time we detect an option?
It feels like it might repeated multiple times if we have multiple options?

lucbv (commented):
I just have a couple questions, it might just be because I am not a great bash wizard x)

cm_generate_makefile.bash
ndellingwood: @lucbv yes, it just stays empty and has no effect

cm_generate_makefile.bash
ndellingwood: This appends options together in a single string with the proper Kokkos' CMake syntax to enable them at configure; in the case of HIP, only rdc seems to be available as an option based on the Kokkos CMake options so if that option is set `KOKKOS_HIP_OPTION_CMD` will translate to `-DKokkos_ENABLE_HIP_RELOCATABLE_DEVICE_CODE=ON ` when cmake is called later in the file to configure Kokkos; we can add more/new options as they become available.
The pattern follows similar to what is done with Cuda which can create a string of the CMake enable options that are then passed on the cmake line to configure Kokkos (e.g., with Cuda if rdc and lambda's are requested to be enabled, the final `KOKKOS_CUDA_OPTION_CMD` variable will translate to `-DKokkos_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE=ON -DKokkos_ENABLE_CUDA_LAMBDA=ON`).

lucbv (approved):
OK, good enough for me : )

ndellingwood:
Thanks for review @lucbv !

PR #822, opened 2020-10-02
CMake: allow to disable unit-test by categories

lucbv:
This feature will allow for easy incremental building and testing of each category for new backend.
Basically this will make life easier for HIP/OpenMPTarget/DPC++ backends.
I tested this by disabling each category one by one and all at once, it worked fine.
When everything is disabled I think that `gtest` is built without any tests and `ctest` does not report any of the unit-tests.

Any opinion on the design or the implementation?


unit_test/CMakeLists.txt
```diff
@@ -38,210 +38,276 @@ IF (KOKKOS_ENABLE_CUDA)
...
-  KOKKOSKERNELS_ADD_UNIT_TEST_EXECUTABLE_AND_TEST(
-    blas_cuda
-    SOURCES
-      Test_Main.cpp
-      ${CUDA_BLAS_SOURCES}
+  IF (KOKKOSKERNELS_DISABLE_BLAS_TESTS)
+    MESSAGE(STATUS "  disabling Cuda BLAS tests")
+  ELSE()
+    APPEND_GLOB(CUDA_BLAS_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/cuda/Test_Cuda_Blas*.cpp)
+    KOKKOSKERNELS_ADD_UNIT_TEST_EXECUTABLE_AND_TEST(
```
jjwilke, line 45: It may be a little bit clearer to modify `KOKKOSKERNELS_ADD_UNIT_TEST_EXECUTABLE_AND_TEST` to do something like:

````
KOKKOSKERNELS_ADD_UNIT_TEST_EXECUTABLE_AND_TEST(
     blas_cuda
      SOURCES
        Test_Main.cpp
        ${CUDA_BLAS_SOURCES}
      )
     COMPONENTS blas
  ENDIF()
````

And have an option like:
````
-DKokkosKernels_COMPONENTS=blas;sparse
````

Then we could also fix the ETI so that only that blas, sparse, etc get built.
````
KOKKOSKERNELS_GENERATE_ETI(Blas1_abs abs
  COMPONENTS blas
  HEADER_LIST ETI_HEADERS
  SOURCE_LIST SOURCES
  TYPE_LISTS FLOATS LAYOUTS DEVICES
)
````

Does this seem cleaner to you? If so, I can push a draft PR.

lucbv: That's better at least because it would avoid the repetition that I did for every backend/component.
I just did not know about that `COMPONENTS` keyword but it seems quite useful to define something akin to sub-modules/sub-packages within a CMake build.
If it's not too much work for you to update my PR or make one of your own it would be nice.
I think at the moment using this approach for tests is more important than for the library itself.
Ultimately it's nice to have both, especially if some users are only doing dense linear algebra, they might not need/want the sparse and graph stuffs built and installed.

jjwilke:
Related to PR #823

lucbv:
This work was superseded by PR #823 so closing it

PR #826, opened 2020-10-08
cm_test_all_sandia: fix determination of CM_ALL_SCRIPT_PATH var

ndellingwood:
Broken for absolute paths, causing update_libs.sh to not be called on some systems


ndellingwood:
This resolves the problem with intel/17 still loading old gcc on some systems, didn't show up when using relative path to the script (which was how I tested previously with dry-run)

e10harvey (approved):

PR #832, opened 2020-10-14
update test scripts

ndellingwood:
add timeout input option
remove bowman
update intel/17 old gcc module swaps


lucbv (approved):
Good update, the timeout flag might come handy and thanks for fixing the gcc toolchain with intel-17!

ndellingwood:
Thanks @lucbv !

PR #835, opened 2020-10-21
Allow backends that only enable a subset of components

jjwilke:
1. Hooks are now given to describe which components are actually enabled by each backend
2. If multiple backends are enabled, it determines which intersection of components are allowed
3. If the user requests an unsupported component, configure exists gracefully with an error message


lucbv:
This has been implemented using a different mechanism in PR #1504

PR #844, opened 2020-11-03
New issue with HIP and Kokkos_ArithTraits

lucbv:
`isinf` and `isnan` are not defined for long double with HIP.
The input parameter is casted to avoid ambiguous function call.
The function is also defined as __device __ so it needs to be decorated with KOKKOS_FORCE_INLINE_FUNCTION.

@crtrott @brian-kelley any idea why this issue would only appear on Tulip?
It is using rocm-3.8.0 so I am not sure why this would not be an issue on caraway?
Maybe because of the gcc toolchain?

### DIFF
--- a/src/Kokkos_ArithTraits.hpp
+++ b/src/Kokkos_ArithTraits.hpp
@@ -1385,17 +1385,25 @@ public:
   static constexpr bool has_infinity = true;
   static KOKKOS_FORCEINLINE_FUNCTION long double infinity() { return HUGE_VALL; }
 
-  static bool isInf (const val_type& x) {
+  static KOKKOS_FORCEINLINE_FUNCTION bool isInf (const val_type& x) {
     #ifdef KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST
     using std::isinf;
     #endif
+#if defined(KOKKOS_ENABLE_HIP)
+    return isinf (static_cast<double>(x));
+#else
     return isinf (x);
+#endif
   }
-  static bool isNan (const val_type& x) {
+  static KOKKOS_FORCEINLINE_FUNCTION bool isNan (const val_type& x) {
     #ifdef KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST
     using std::isnan;
     #endif
+#if defined(KOKKOS_ENABLE_HIP)
+    return isnan (static_cast<double>(x));
+#else
     return isnan (x);
+#endif
   }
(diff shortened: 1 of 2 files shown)


dalg24:
> `isinf` and `isnan` are not defined for long double with HIP.

So you know, you will have the same issue for all mathematical functions.  This is not new it is also true with CUDA.  And this will be also true for OpenMP Target and SYCL.

lucbv:
OK, in its current form the PR removes the `long double` instantiation of `Kokkos_ArithTraits` and `InnerProductSpaceTraits` using pre-processor macro. This should make it a bit clearer for users that this type is not supported on devices (Nvidia or AMD) at the moment.

## Kokkos-dev2 spot-check
PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=695 run_time=149
clang-8.0-Pthread_Serial-release build_time=233 run_time=132
clang-9.0.0-Pthread-release build_time=153 run_time=62
clang-9.0.0-Serial-release build_time=158 run_time=48
cuda-10.1-Cuda_OpenMP-release build_time=846 run_time=150
cuda-11.0-Cuda_OpenMP-release build_time=864 run_time=148
cuda-9.2-Cuda_Serial-release build_time=818 run_time=220
gcc-7.3.0-OpenMP-release build_time=175 run_time=50
gcc-7.3.0-Pthread-release build_time=127 run_time=63
gcc-8.3.0-Serial-release build_time=158 run_time=51
gcc-9.1-OpenMP-release build_time=224 run_time=52
gcc-9.1-Serial-release build_time=184 run_time=49
intel-17.0.1-Serial-release build_time=396 run_time=53
intel-18.0.5-OpenMP-release build_time=691 run_time=50
intel-19.0.5-Pthread-release build_time=365 run_time=66

caraway is down at the moment but I got a clean build on Tulip!

@brian-kelley @dalg24 any suggestions or comment, I would like this to be closed before the release so I can build Kokkos Kernels with Spack on EAs : )

src/Kokkos_ArithTraits.hpp
```diff
@@ -1367,9 +1367,10 @@ class ArithTraits<double> {
 };
 
 
-// CUDA does not support long double in device functions, so none of
-// the class methods in this specialization are marked as device
-// functions.
+// CUDA and HIP do not support long double in device functions,
```
dalg24, line 1370: This will be true for SYCL and OpenMPTarget as well

dalg24 (approved):

brian-kelley (approved):

PR #863, opened 2020-12-18
Setup nightly HIP builds on ONRL's Jenkins

dalg24:
The job is setup [here](https://cloud.cees.ornl.gov/jenkins-ci/job/KokkosKernels/). I erased the trial builds I did when testing.  It is currently configured to run once a day between 2 and 5am ET.  It is tracking my fork but I will update after we merge.

### DIFF
--- a/.jenkins/nightly.groovy
+++ b/.jenkins/nightly.groovy
@@ -1,0 +1,41 @@
+pipeline {
+    agent none
+
+    stages {
+        stage('HIP-ROCm-3.10-C++14') {
+            agent {
+                dockerfile {
+                    filename 'Dockerfile.hip'
+                    dir 'scripts/docker'
+                    additionalBuildArgs '--build-arg BASE=rocm/dev-ubuntu-20.04:3.10'
+                    label 'rocm-docker && vega'
+                    args '-v /tmp/ccache.kokkos:/tmp/ccache --device=/dev/kfd --device=/dev/dri --security-opt seccomp=unconfined --group-add video --env HIP_VISIBLE_DEVICES=$HIP_VISIBLE_DEVICES'
+                }
+            }
+            steps {
+                sh '''rm -rf kokkos &&
+                      git clone -b develop https://github.com/kokkos/kokkos.git && cd kokkos && \
+                      mkdir build && cd build && \
+                      cmake \
+                        -DCMAKE_CXX_COMPILER=hipcc \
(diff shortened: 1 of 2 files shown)


lucbv:
@ndellingwood can you check that the build parameters are similar to what we usually do for our nightly builds?

.jenkins/nightly.groovy
```diff
@@ -0,0 +1,41 @@
...
+                        -DCMAKE_CXX_COMPILER=hipcc \
+                        -DKokkos_ENABLE_HIP=ON \
+                        -DKokkos_ARCH_VEGA906=ON \
+                        .. && \
+                       make -j8 && make install && \
+                       cd ../.. && rm -rf kokkos'''
+                sh '''rm -rf build && mkdir -p build && cd build && \
+                      cmake \
+                        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+                        -DCMAKE_CXX_COMPILER=hipcc \
```
ndellingwood, line 29: Shouldn't there be a `-DKokkos_DIR=<path-to-kokkos-install>` to point to the Kokkos install?

ndellingwood (commented):
Thanks for working on this @dalg24 , I only have the question about how kokkos-kernels is picking up the path to the kokkos install. Since this is running in a docker container, there must be sudo privileges allowing the kokkos install to be somewhere analogous to /usr/lib and cmake will search the location automatically?

.jenkins/nightly.groovy
dalg24: I actually really dislike the using `<PackageName>_DIR` in to stir the search procedure in CMake.  I could have `-DKokkos_ROOT=/opt/kokkos` but I chose to install it there in part because I knew it was going to fall into the set of possible prefixes that CMake would construct for Unix.

dalg24: After chatting with Nathan I realize `-DKokkos_ROOT` wouldn't work because KokkosKernels currently requires a minimum cmake version of 3.10 but this would work when we bump to 3.12+

ndellingwood (approved):

ndellingwood:
@lucbv @dalg24 are any further changes anticipated to the PR, or shall I merge?

dalg24:
As far as I am concerned it is ready.

dalg24:
I updated on the Jenkins side to point to kokkos/kokkos-kernels instead of my fork.

PR #864, opened 2020-12-18
WIP: improved MIS2 PRNG

brian-kelley:
Don't review or merge yet. This has some temporary experiment drivers in perf_test/graph.

Instead of vanilla xorshift, use the PRNG from Kokkos (xorshift with a linear congruential step). This significantly decreases the number of iterations needed by MIS2.


brian-kelley:
@crtrott Kokkos random is perfect for what it does, but for this I'm using it as a digest of some vertex properties rather than scrambling an RNG state. That way MIS2 is deterministic.

Issue #877, opened 2021-01-21
IntelClang compiler: __assumed_aligned is not supported by icpx and dpc++


lucbv:
The new clang based intel compilers are not supporting the `__assume_aligned` function which generates build errors for the OpenMP backend on iris.
A simple fix will be to check if the `__clang__` macro is defined before enabling this part of Kokkos Kernels.
A more long term solution will be the use of `std::assume_aligned<>()` when c++20 is required.

lucbv:
fix merged in develop via PR #878

PR #878, opened 2021-01-21
IntelClang guarding __assume_aligned with !defined(__clang__)

lucbv:
With the new check we are making sure that the IntelClang compilers that do not define this function (icpx/dpc++) are not generating build errors in the OpenMP backend (would likely be a problem for OpenMP Target too since it uses the same compilers).
This fixes an issue in Kokkos Kernels on the iris/yarrow platforms see issue #877 , a more permanent solution will be available with the c++20 standard which defines: `std::assume_aligned<>()`.
This was tested on iris/yarrow, will now run checks on kokkos-dev2.

### DIFF
--- a/perf_test/sparse/spmv/OpenMPSmartStatic_SPMV.hpp
+++ b/perf_test/sparse/spmv/OpenMPSmartStatic_SPMV.hpp
@@ -138,7 +138,7 @@ void openmp_smart_static_matvec(AType A, XType x, YType y) {
 
   #pragma omp parallel
   {
-#ifdef KOKKOS_COMPILER_INTEL
+#if defined(KOKKOS_COMPILER_INTEL) && !defined(__clang__)
     __assume_aligned(x_ptr, 64);
     __assume_aligned(y_ptr, 64);
 #endif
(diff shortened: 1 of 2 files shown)


dalg24 (approved):

dalg24:
Maybe add a comment that if this pattern occurs again we might want to provide something on the Kokkos side.  At the moment our position is we would rather not distinguish native and clang-based versions of the vendor compiler if we can avoid it.

lucbv:
FYI things are good on kokkos-dev2

PASSED TESTS
clang-8.0-Cuda_OpenMP-release build_time=735 run_time=142
clang-8.0-Pthread_Serial-release build_time=245 run_time=137
clang-9.0.0-Pthread-release build_time=147 run_time=62
clang-9.0.0-Serial-release build_time=164 run_time=50
cuda-10.1-Cuda_OpenMP-release build_time=929 run_time=146
cuda-11.0-Cuda_OpenMP-release build_time=937 run_time=134
cuda-9.2-Cuda_Serial-release build_time=894 run_time=225
gcc-7.3.0-OpenMP-release build_time=183 run_time=48
gcc-7.3.0-Pthread-release build_time=164 run_time=77
gcc-8.3.0-Serial-release build_time=169 run_time=52
gcc-9.1-OpenMP-release build_time=215 run_time=49
gcc-9.1-Serial-release build_time=217 run_time=56
intel-17.0.1-Serial-release build_time=445 run_time=74
intel-18.0.5-OpenMP-release build_time=704 run_time=51
intel-19.0.5-Pthread-release build_time=369 run_time=63

PR #949, opened 2021-04-30
Clean up CrsMatrix raw pointer constructor

brian-kelley:
Before, it was converting a rowmap to non-prefix-summed row counts
and passing that to StaticCrsGraph, which had to do the prefix sum
again. This PR just uses unmanaged views instead, with minimal copies and allocations.

PR #971, opened 2021-05-11
Cherry pick #949: fix CrsMatrix raw ptr ctor

brian-kelley:
See #949. This was the only necessary KokkosKernels change to get Tpetra working on HIP (in https://github.com/trilinos/Trilinos/pull/9090).


brian-kelley:
I tested this (rebased onto the 3.4.1 candidate) locally with cuda/openmp/serial (no TPLs) and then a serial build with the usual set of strict warnings enabled, and both of those were all good.

PR #1114, opened 2021-09-23
SpMV: improving performance on HIP backend

lucbv:
Adding specific parameters for improved SpMV performance on HIP backend.
This probably needs further testing and integration with the current launch parameters heuristic.
However this is the algorithm we use for the Kokkos ecosystem paper so it's probably time to bring that in...

### DIFF
--- a/src/sparse/impl/KokkosSparse_spmv_impl.hpp
+++ b/src/sparse/impl/KokkosSparse_spmv_impl.hpp
@@ -271,6 +271,7 @@ int64_t spmv_launch_parameters(int64_t numRows, int64_t nnz, int64_t rows_per_th
     while(vector_length < max_vector_length && vector_length * 6 < nnz_per_row)
       vector_length*=2;
   }
+  vector_length=8;
 
   // Determine rows per thread
   if(rows_per_thread < 1) {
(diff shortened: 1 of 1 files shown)


src/sparse/impl/KokkosSparse_spmv_impl.hpp
```diff
@@ -271,6 +271,7 @@ int64_t spmv_launch_parameters(int64_t numRows, int64_t nnz, int64_t rows_per_th
     while(vector_length < max_vector_length && vector_length * 6 < nnz_per_row)
       vector_length*=2;
   }
+  vector_length=8;
```
brian-kelley, line 274: Is ``vector_length = 8`` meant for all backends? If it is, we can just take out the above logic about max_vector_length and the loop to choose it based on avg nnz per row.

lucbv: No, it was just a value that gave us reasonable results on matrices for the Kokkos Ecosystem paper but it would be better to test this out on a larger set of matrices and get a sense for a heuristic that works well on AMD.
I am not at all sure that 8 is good on other architectures... that's why this is definitely just a draft that I need to work on x)

PR #1167, opened 2021-11-02
Fix bug with mkl impl of spgemm

ndellingwood:


ndellingwood:
Matching fix to trilinos/Trilinos#9891.
This is only needed for master branch, was resolved on develop branch by PR #1135 with the clang-format enhancement which was not included in 3.5.0 release

src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp
```diff
@@ -425,12 +424,6 @@ void mkl_symbolic(
      MKL\n"); return;
       }
   */
-  if (std::is_same<idx, int>::value) {
```
e10harvey, line 428: I don't see what this if-block ends; is there a matching `}` for it? Should the if-block just be converted to a block -- `{`.

e10harvey (changes requested):
THanks, @ndellingwood ! Please see questions below.

src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp
ndellingwood: @e10harvey there is not a matching brace to close this if block, that was what was causing the compilation error. There was an apparent copy+paste error in a past update to this file that duplicated the lines of code that I removed in this PR (and so removing the mismatch in open-closed braces)

ndellingwood: @e10harvey See lines 433-448 of this file which had mistaken duplication to lines 428-431, removed by this PR

e10harvey (approved):

e10harvey:
@ndellingwood, @lucbv: Shall I enable the auto-tester for pull requests targeting `master`?

ndellingwood:
Alright, looks like the issue was never on the develop branch but had to do with problems with a cherry-pick to the release-candidate. The duplication occurred with commit 05986906f , here is the diff https://github.com/kokkos/kokkos-kernels/pull/1149/commits/05986906f4acce3d2091f6181db2b099a6780945

PR #1275, opened 2022-01-20
CONTRIBUTING.md - added rev1

e10harvey:
Added initial CONTRIBUTING.md docs. Please review and make edits by using the GitHub suggest changes feature in this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,73 @@
+# New Developers
+## Coding Style
+We follow google's c++ coding style. See https://google.github.io/styleguide/cppguide.html and https://github.com/kokkos/kokkos-kernels/blob/master/.clang-format for details. 
+### Automate coding style via a pre-commit hook:
+```bash
+cat kokkos-kernels/.git/hooks/pre-commit
+for FILE in $(git diff --cached --name-only | egrep '.*\.cpp$|.*\.hpp$|.*\.h$')
+do
+        clang-format-8 -i -style=file $FILE
+        git add $FILE
+done
+chmod +x kokkos-kernels/.git/hooks/pre-commit
+```
+### Conditionally enable or disable formatting
+```c++
+// clang-format off
+cpp code here
+// clang-format on
+```
+##### Ensure that clang-format-8 is installed.
+###### Mac
+```bash
+brew install clang-format-8
+```
+
+###### Ubuntu
+```bash
+apt install clang-format-8
+```
+
+## Comment Style
+We follow doxygen style comments for both external (API) and internal members. See https://www.doxygen.nl/manual/docblocks.html for details.
+Our documentation can be generated using the `-DKokkosKernels_ENABLE_DOCS:BOOL=ON` cmake flag.
+
+In general, we prefer that the prototype has the doxygen style comment rather than the definition. If there is no prototype, then the definition should have the doxygen style comment.
+### API Doxygen Style Example
+```c++
+/// \brief Blocking wrapper for accessing a Kokkos View.
+/// \tparam ViewValueType The value type (Scalar or Vector) of each view element
+/// \tparam ViewType The view type
+/// \param v The view handle
+/// \param m The requested row index of v
+/// \param n The requested col index of v
+/// \return If m and n are within the extents of v, a valid element of v;
+///         otherwise, the last element of v.
+///
+template <class ViewValueType, class ViewType>
+KOKKOS_INLINE_FUNCTION ViewValueType
+access_view_bounds_check(ViewType v, int m, int n, const BoundsCheck::Yes &);
+```
+
+# Library policies
+
+## Upcasting and downcasting
+TODO
+
+## Blocking and non-blocking interfaces
+All the APIs are non-blocking unless:
+1. A TPL is enabled
+2. The result vector resides on the host and work is offloaded to a device
+
+When a TPL is enabled, we follow the blocking semantics of the TPL interface.
+
+If no TPLs are enabled, callers can avoid blocking calls by using any overload which accepts a result vector type as a template argument.
+
+# TODOs
+- [] Move library policies out of CONTRIBUTING.md?
+- [] Code owners of `blas/{sparse,dense}`, `batched{sparse,dense}` review and report policy violations.
+  - [] blas/sparse
+  - [] blas/dense
+  - [] batched/sparse
+  - [] batched/dense
+- [] Add up/downcosting CI checks?
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,8 +6,10 @@ We follow google's c++ coding style. See https://google.github.io/styleguide/cpp
 cat kokkos-kernels/.git/hooks/pre-commit
 for FILE in $(git diff --cached --name-only | egrep '.*\.cpp$|.*\.hpp$|.*\.h$')
 do
-        clang-format-8 -i -style=file $FILE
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
+# New Developers
```
e10harvey, line 1:
```suggestion
# Contributing
```

jennloe (commented):
Would also be useful to have info on how to add Unit Tests and Performance Tests?  

Maybe at least a link to the wiki page explaining the logic of the file structure? 

Also, somewhere we should document how to add documentation to the Wiki.  But this maybe isn't the right spot.

e10harvey:
CC: @jgfouca

cwpearson:
Right now some instructions for adding/running unit tests is in the internal wiki, but I think it makes sense to move them out here to Github. I can take a first pass at that

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+## Coding Style
+We follow google's c++ coding style. See https://google.github.io/styleguide/cppguide.html and https://github.com/kokkos/kokkos-kernels/blob/master/.clang-format for details. 
+### Automate coding style via a pre-commit hook:
+```bash
+cat kokkos-kernels/.git/hooks/pre-commit
+for FILE in $(git diff --cached --name-only | egrep '.*\.cpp$|.*\.hpp$|.*\.h$')
+do
```
lucbv, line 8: should we update this script to avoid running git diff on deleted files?

lucbv:
@cwpearson we can move them out, ideally I want to add other bits to the documentation we have in the internal wiki regarding development policy but there is just never enough time...
@e10harvey we should also mention that we do not allow headers that are system specific, I believe we have some at the moment and they create problems on Windows machines. Ideally we should try to stick to c++ headers if possible especially since we require specific version of the c++ standard.

CONTRIBUTING.md
```diff
@@ -0,0 +1,78 @@
...
+/// \param m The requested row index of v
+/// \param n The requested col index of v
+/// \return If m and n are within the extents of v, a valid element of v;
+///         otherwise, the last element of v.
+///
+template <class ViewValueType, class ViewType>
+KOKKOS_INLINE_FUNCTION ViewValueType
+access_view_bounds_check(ViewType v, int m, int n, const BoundsCheck::Yes &);
+```
+
```
e10harvey, line 53:
```suggestion

## Generating KokkosKernels documentation
Add `-DKokkosKernels_ENABLE_DOCS:BOOL=ON` to your cmake.
```
make Doxygen
make Sphinx
```

CC: @cwpearson, @jennloe, @jgfouca, @lucbv, @vqd8a, @kliegeois

Comment (jennloe) [reply CONTRIBUTING.md:53]: Note: To build docs on a Mac:
brew install doxygen
pip install sphinx
pip install breathe
pip install sphinx-rtd-theme

File: CONTRIBUTING.md [hunk 4]
```diff
@@ -0,0 +1,78 @@
...
+## Blocking and non-blocking interfaces
+All the APIs are non-blocking unless:
+1. A TPL is enabled
+2. The result vector resides on the host and work is offloaded to a device
+
+When a TPL is enabled, we follow the blocking semantics of the TPL interface.
+
+If no TPLs are enabled, callers can avoid blocking calls by using any overload which accepts a result vector type as a template argument.
+
```
vqd8a, line 70: Move @cwpearson 's stuff from https://gitlab-ex.sandia.gov/kokkos-kernels-internal/kokkos-kernels-internal/-/wikis/adding-unit-tests to here

```suggestion
# Adding Unit-Tests

The unit-tests are used to check that Kokkos Kernels library is building and running correctly on all the platforms we support. They are run for every pull-request and every night in our nightly builds. They should ideally cover every code path possible across each architecture and are typically required when submitting a pull-request that adds a new capability to the library.

## Unit-tests organization

The unit-tests are organized by categories, as follows:
- `common`: for utilities and kernels that do not fit in any particular category
- `blas`: for all dense blas/lapack implementations (Team/Thread/Vector levels)
- `batched`: similar to blas but using a batched interface that provides SIMD like operations
- `sparse`: for sparse linear algebra operations and solvers
- `graph`: for graph algorithms

Unit-tests are located in `kokkos-kernels/unit_test/category` and are implemented in header files. Each category contains a master header file `Test_Category.hpp` which includes all individual test header of that category.

Additionally backends have an associated header file `Test_Backend.hpp` and source files for each (backend, category) pair `Test_Backend_Category.cpp`, these can be found in `kokkos-kernels/unit-test/backend`. Each of the backend - category source file eventually defines an executable `KokkosKernels_category_backend`.

## Implement unit-tests

This design requires each unit-test header to isolate appropriately the functions, classes, struct, etc... that it defines to avoid potential redefinition in other tests.

## Run unit-tests

To enable unit-tests, configure Kokkos Kernels with `cmake .. -DKokkosKernels_ENABLE_TESTS=ON`

You can build all targets, or try for a specific one that's relevant to you: `make KokkosKernels_sparse_cuda`

Kokkos Kernels uses CTest to register and run test binaries, and within each binary, GTest to register and run individual tests.

To list all test binaries, `ctest -N`

To run a single binary, use the `ctest -R` option.
It is also useful to pass `-V`, otherwise CTest will capture stdout and stderr. Example: `ctest -R sparse_cuda -V`

To run a single test within a binary, you need to isolate the test command, then add flags to restrict the tests GTest will run. For example: `ctest -R sparse_cuda -V -N`
Then you can invoke the test command directly, with additional gtest flags if desired:
`--gtest_list_tests` will list the names of all tests without running them:
`--gtest_filter` allows you to pass a regular expression to filter tests by name

File: CONTRIBUTING.md (hunk 3 above)
Comment (e10harvey) [inline CONTRIBUTING.md:53]:
```suggestion


### Struct/Class Doxygen Style Example
```
/// BatchLayout class used to specify where the batch dimension is
/// allocated in the input views for host-level Batched BLAS/LAPACK routines.
struct BatchLayout {
  /// Batch dimension is the leftmost dimension within input views
  struct Left {};
  /// Batch dimension is the rightmost dimension within input views
  struct Right {};
};
```
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,78 @@
+# New Developers
+## Coding Style
+We follow google's c++ coding style. See https://google.github.io/styleguide/cppguide.html and https://github.com/kokkos/kokkos-kernels/blob/master/.clang-format for details. 
+### Automate coding style via a pre-commit hook:
+```bash
+cat kokkos-kernels/.git/hooks/pre-commit
+for FILE in $(git diff --cached --name-only | egrep '.*\.cpp$|.*\.hpp$|.*\.h$')
+do
+  if [ -e $file ]; then
+    clang-format-8 -i -style=file $FILE
```
jgfouca, line 10: Is `clang-format-8` is specific to MacOS? If so, some comments are needed here for Linux users. The SEMS modules for clang are putting `clang-format` in my PATH.

masterleinad: In the end, it's only important that `clang-format` has version 8. Newer `clang-format` version indent slightly differently so that using a specific version is required.

CONTRIBUTING.md
e10harvey: TODO: Add requirements.txt for python dependencies.

e10harvey:
Closed in favor of #1368.

PR #1323, opened 2022-02-15
Fix issue with bfloat16 and matrix market files. 

jennloe:
Fixes crash when reading from a Matrix Market file into bfloat16 format.  

My only changes are here.  The rest is from the formatter, I think. 
```git diff ../../src/common/KokkosKernels_IOUtils.hpp
diff --git a/src/common/KokkosKernels_IOUtils.hpp b/src/common/KokkosKernels_IOUtils.hpp
index 9b2e06c4..249e5ed0 100644
--- a/src/common/KokkosKernels_IOUtils.hpp
+++ b/src/common/KokkosKernels_IOUtils.hpp
@@ -1088,7 +1088,8 @@ int read_mtx(const char *fileName, lno_t *nrows, lno_t *ncols, size_type *ne,
 
   if (fline.find("real") != std::string::npos ||
       fline.find("double") != std::string::npos) {
-    if (std::is_same<scalar_t, Kokkos::Experimental::half_t>::value)
+    if (std::is_same<scalar_t, Kokkos::Experimental::half_t>::value ||
+        std::is_same<scalar_t, Kokkos::Experimental::bhalf_t>::value)
       mtx_field = REAL;
     else {
       if (!std::is_floating_point<scalar_t>::value)
@@ -1109,7 +1110,8 @@ int read_mtx(const char *fileName, lno_t *nrows, lno_t *ncols, size_type *ne,
   } else if (fline.find("integer") != std::string::npos) {
     if (std::is_integral<scalar_t>::value ||
         std::is_floating_point<scalar_t>::value ||
-        std::is_same<scalar_t, Kokkos::Experimental::half_t>::value)
+        std::is_same<scalar_t, Kokkos::Experimental::half_t>::value ||
+        std::is_same<scalar_t, Kokkos::Experimental::bhalf_t>::value)
       mtx_field = INTEGER;
     else
       throw std::runtime_error(
       ```


jennloe:
@ndellingwood This is the one that needs to go in the release.

jennloe:
Ooops.  Just realized I also committed my hackish vector reading utility.  Do you want to keep this, or should I remove it?
```template <typename idx_array_type>
inline void kk_read_MM_vector(idx_array_type &view, const char *filename){

  typedef typename idx_array_type::HostMirror host_type;
  //typedef typename idx_array_type::size_type idx;
  host_type host_view = Kokkos::create_mirror_view (view);
  std::ifstream myFile (filename, std::ios::in );

  std::string line;
  std::getline(myFile, line);  // have line 1
  std::getline(myFile, line);  

  for (size_t i = 0; i < view.extent(0); ++i){
	  myFile >> host_view(i);
  }
  myFile.close();
  Kokkos::deep_copy (view, host_view);
  Kokkos::fence();
}

jennloe:
Rats.  This doesn't even have the changes for bfloat16.  What did I do?
Lemme close this PR and fix it in a new one...

PR #1611, opened 2022-11-30
[WIP] Add Kokkos::Impl::assume_aligned for broader compiler support

cwpearson:
This is a work in progress

* gcc and clang have `__builtin_assume_aligned`
* nvcc added `assume_aligned` in 11.2 
* msvc has `__assume()` which can be used in a similar way
* intel has `__assume_aligned`
- [ ] HIP?
- [ ] What are the right macros to use to detect the compiler and host code vs device code?

If MSVC can't see `__assume` through a function return a pure macro-based approach may be more appropriate.


cwpearson:
Dropping this, not much progress was made and if we have a good use for it we can redo it.

PR #1805, opened 2023-04-24
Blas1: docs update for PR #1803

lucbv:
updating documentation for changes in PR #1803, this PR will be rebased once #1803 is merged.
It mostly includes additional overloads of our BLAS1 functions to accept an execution space instance.
Additional documentation improvement have been made as well.

### DIFF
--- a/blas/src/KokkosBlas1_abs.hpp
+++ b/blas/src/KokkosBlas1_abs.hpp
@@ -25,8 +25,8 @@ namespace KokkosBlas {
 
 /// \brief R(i,j) = abs(X(i,j))
 ///
-/// Non-blocking function to replace each entry in R with the absolute value (magnitude) of the
-/// corresponding entry in X.
+/// Non-blocking function to replace each entry in R with the absolute value
+/// (magnitude) of the corresponding entry in X.
 ///
 /// \tparam execution_space a Kokkos execution space to run the kernels on.
 /// \tparam RMV 1-D or 2-D Kokkos::View specialization.
@@ -35,7 +35,8 @@ namespace KokkosBlas {
 ///   those of RMV.
 ///
 /// \param space [in] an execution_space instance where the kernel will run.
-/// \param R [out] view of type RMV that contains the absolute value X on output.
+/// \param R [out] view of type RMV that contains the absolute value X on
+/// output.
 /// \param X [in] view of type XMV.
 template <class execution_space, class RMV, class XMV>
 void abs(const execution_space& space, const RMV& R, const XMV& X) {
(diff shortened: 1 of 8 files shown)


blas/src/KokkosBlas1_abs.hpp
```diff
@@ -63,24 +79,40 @@ void abs(const RMV& R, const XMV& X) {
...
   XMV_Internal X_internal = X;
 
-  Impl::Abs<RMV_Internal, XMV_Internal>::abs(R_internal, X_internal);
+  Impl::Abs<execution_space, RMV_Internal, XMV_Internal>::abs(space, R_internal,
+                                                              X_internal);
+}
+
+/// \brief R(i,j) = abs(X(i,j))
+///
+/// Replace each entry in R with the absolute value (magnitude) of the
```
e10harvey, line 103:
```suggestion
/// Non-blocking function to replace each entry in R with the absolute value (magnitude) of the
```

blas/src/KokkosBlas1_abs.hpp
```diff
@@ -28,18 +28,34 @@ namespace KokkosBlas {
 /// Replace each entry in R with the absolute value (magnitude) of the
 /// corresponding entry in X.
 ///
+/// \tparam execution_space a Kokkos execution space to run the kernels on.
```
e10harvey, line 31:
```suggestion
/// \tparam execution_space a Kokkos execution space to run the kernels on (default: RMV::execution_space)
```

e10harvey (approved):
Before merging this, I would like the doxygen style comments updated to note:
1. Blocking behavior of each function
2. Default execution space of each function

shall I open a PR into your fork for these changes?

blas/src/KokkosBlas1_abs.hpp
e10harvey: Actually, we should document this in the overload without a `space` parameter.

lucbv:
@e10harvey thanks for approving this although it would be best if you approved the original PR as it needs to merge before this one...

e10harvey (approved):

e10harvey:
> @e10harvey thanks for approving this although it would be best if you approved the original PR as it needs to merge before this one...

Both are approved now. Merging this PR should mark #1803 as merged since these commits are a superset of the commits in 1803.

lucbv:
Actually I believe I have fixed an issue or two with the TPLs yesterday that is not in this branch so this branch needs to be at least rebased.
More importantly this PR contains code changes that are not tested since it is marked as WIP, merging it would be fairly unsafe for the code, again because there can be some diff that appeared while I was writing the documentation here.
In general if there is a code change I **require** that it goes through testing.

e10harvey:
> Actually I believe I have fixed an issue or two with the TPLs yesterday that is not in this branch so this branch needs to be at least rebased.
> More importantly this PR contains code changes that are not tested since it is marked as WIP, merging it would be fairly unsafe for the code, again because there can be some diff that appeared while I was writing the documentation here.
> In general if there is a code change I **require** that it goes through testing.

Okay, note that this PR went from 11 commits to 3 commits after #1803 merged and that these 3 commits only touch c++ comments and .rst files which were tested via github-DOCS in this PR.

PR #1822, opened 2023-05-09
Ger: adding documentation stubs in apidocs

lucbv:
Really not much to say beyond the fact that we forgot to do it while adding the ger function...


e10harvey (approved):
Thank you!

PR #1941, opened 2023-08-14
.github/workflows/docs.yml: Use up-to-date doxygen version

e10harvey:
There seems to be a bug we're hitting in versions of doxygen that are < 1.9.7.

Brian has been hitting this in https://github.com/kokkos/kokkos-kernels/pull/1932 while trying to document the spmv overloads.

To address this blocking issue, let us use doxygen 1.9.7, which is available via brew on the macos VMs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -11,15 +11,16 @@ permissions:
 
 jobs:
   docs-check:
-    runs-on: ubuntu-latest
+    runs-on: [macos-latest]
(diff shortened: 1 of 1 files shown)


brian-kelley (approved):

e10harvey:
Using label 'AT: WIP' to avoid running all our CI checks on this diff.

e10harvey:
Merging since github-DOCS passed.

PR #2078, opened 2023-12-19
SpGEMM oneAPI: adding TPL interface with oneAPI MKL

lucbv:
@jczhang07 
Starting some work to interface with oneapi sparse::matmat. This will help for multigrid and domain decomposition algorithms.


lucbv:
Slowly getting started on this, so far only the TPL avail layer is added. I am building/testing on Aurora.
Hopefully I can make some progress on the actual interface before the break, this will probably need some eyes on it as this kernel has a fair bit of setup which could lead to sub-optimal performance... basically feel free to criticize things as they move forward!

PR #2210, opened 2024-05-23
Enable 3 at2 builds

e10harvey:


lucbv (approved):
We might want to make some changes to the configs to tune them but for now it's good to go

PR #2221, opened 2024-05-29
Kokkos Kernels: adding SHA for github actions

lucbv:
This would improve our security scorecard thing see here: https://securityscorecards.dev/viewer/?uri=github.com/kokkos/kokkos-kernels

### DIFF
--- a/.github/workflows/bdw.yml
+++ b/.github/workflows/bdw.yml
@@ -188,12 +188,12 @@ jobs:
     
     steps:
       - name: checkout_kokkos_kernels
-        uses: actions/checkout@v3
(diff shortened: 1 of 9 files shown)


lucbv:
Adding `AT: WIP` because I do not want the auto-tester running this one but everything else that changed for the github actions is passing so the changes should be good : )

e10harvey (approved):
Nice! I would just confirm that the github ci .yml changes still run with these changes before merging.

lucbv:
It does, I confirmed it before that last commit for the Docker files, all was running just fine : )
I can remove the `AT: WIP` and re-run the workflows though

PR #2243, opened 2024-06-12
.github/workflows: Auto trigger AT2 ci upon review

e10harvey:
Luc, this is ready for review.

Summary of changes:
- Every time a PR review is submitted, the BDW, H100, and MI210 checks will be triggered. These checks should only run upon PR review submission if they have not already passed on the head commit of the source branch in the given PR.

### DIFF
--- a/.github/workflows/bdw.yml
+++ b/.github/workflows/bdw.yml
@@ -9,6 +9,9 @@ on:
     - '**/*.py'
     - 'docs/**'
     types: [ opened, reopened, synchronize ]
+  pull_request_review:
+    types:
+      - submitted
 
 permissions:
   contents: none
--- a/.github/workflows/h100.yml
+++ b/.github/workflows/h100.yml
@@ -10,6 +10,9 @@ on:
     - '**/*.py'
     - 'docs/**'
     types: [ opened, reopened, synchronize ]
+  pull_request_review:
+    types:
+      - submitted
 
 permissions:
   contents: none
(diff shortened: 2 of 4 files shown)


cwpearson:
> These checks should only run upon PR review submission if they have not already passed on the head commit of the source branch in the given PR.

Is this part of the semantics of `pull_request_review` or in conjunction with the other events that trigger a workflow run?

lucbv:
Okay so as long as we only have Sandians in the approved list to perform reviews this check should be reasonable correct? If so I can have this merged

e10harvey:
> > 
> 
> Is this part of the semantics of `pull_request_review` or in conjunction with the other events that trigger a workflow run?

Solely `pull_request_review` semantics. See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_review.

e10harvey:
> Okay so as long as we only have Sandians in the approved list to perform reviews this check should be reasonable correct? If so I can have this merged

That's correct. The review authors must be from the approved list.

e10harvey (commented):
Test trigger.

lucbv:
So I am still worried that this means that a blocking review being issued will lead to running the tests?
Am I interpreting this wrongly?
If not I suspect that this will lead to undesirable code running after blocking reviews.

lucbv:
Would this work?
It would ensure that when a PR is requiring changes it does not run, comments and approvals should run.
What do you think?

```
on:
  pull_request_review:
    types: [submitted]

jobs:
  not_rejected:
    if: (github.event.review.state != 'BLOCKED')
    runs-on: ubuntu-latest
    steps:
      - run: echo "This PR was approved"
```

lucbv (approved):

PR #2329, opened 2024-09-06
AT-2: adding non-TPL build for HIP backend

lucbv:
Clean-up a few things in the mi210 workflow as well should make it a little bit more readable and maintainable.

### DIFF
--- a/.github/workflows/mi210.yml
+++ b/.github/workflows/mi210.yml
@@ -107,12 +107,12 @@ jobs:
           mkdir -p kokkos/{build,install}
           cd kokkos/build
           HIPCC=$(which hipcc)
-          cmake -S /kokkos \
-            -B kokkos/build \
+          cmake -S "$GITHUB_WORKSPACE/kokkos" \
+            -B "$GITHUB_WORKSPACE/kokkos/build" \
             -D CMAKE_CXX_COMPILER=$HIPCC \
             -D CMAKE_CXX_FLAGS=-O3 \
             -D CMAKE_EXE_LINKER_FLAGS= \
-            -D CMAKE_INSTALL_PREFIX=kokkos/install \
+            -D CMAKE_INSTALL_PREFIX="$GITHUB_WORKSPACE/kokkos/install" \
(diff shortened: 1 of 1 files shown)


lucbv:
@ndellingwood @cwpearson the new mi210 without TPLs build is doing fine. Let me know if you want any changes otherwise we should be good to merge this

.github/workflows/mi210.yml
```diff
@@ -4,90 +4,89 @@ on:
...
+        uses: actions/checkout@v4
+        with:
+          repository: kokkos/kokkos
+          ref: ${{ github.base_ref }}
+          path: kokkos
+
+      - name: configure_kokkos
+        run: |
+          echo "GITHUB_WORKSPACE = $GITHUB_WORKSPACE"
+          mkdir "$GITHUB_WORKSPACE/kokkos/build"
```
ndellingwood, line 27: Thanks for the work on this @lucbv !
When running this job through GA, I just wanted to check that the `GITHUB_WORKSPACE` directory is unique to each job (the name includes for example a sha, timestamp, or contains a PR number)?
If not (and if there is not clean up of previous jobs) then the `mkdir` lines could trigger an error if the directory already exists (adding a `-p` would avoid the error). I'm assuming this should not occur, but wanted to check for my own understanding

ndellingwood (approved):

.github/workflows/mi210.yml
lucbv: Good question, each jobs runs in its own container image, here it happens to be the same but it gets reloaded each time. Also since we have two nodes to test on caraway these might actually run on two different compute nodes concurrently. I saw them running at the same time so in this particular PR that's definitely what happened.
The downside of this approach is that you cannot go back and see the artifacts from the container, you have to rerun the pipeline in an interactive container yourself to observe the error... we could save the image after the container is done running the tests and save it somewhere but I think we would use a lot of space so we are not doing that for now...

ndellingwood: Ah right, thanks! Yeah, this is within the container, so no worries 👍

PR #2354, opened 2024-10-02
`kk_is_gpu_exec_space()` -> `is_gpu_exec_space_v`

cwpearson:
* `kk` prefix is unnecessary because of the `KokkosKernels::Impl` namespace
* replace function with `inline constexpr bool` for simplicity.


lucbv (approved):
Okay, this is a bit more inline with how the standard would define it.

PR #2366, opened 2024-10-08
workflow: update the documentation workflow to use pinned python modules

lucbv:


ndellingwood (approved):

lucbv:
Okay since the `github-DOCS` action has not died in that PR let's merge!

PR #2368, opened 2024-10-08
`OpenMPSmartStatic_SPMV.hpp`: throw if posix_memalign fails

cwpearson:


lucbv (approved):

cwpearson:
@lucbv it looks like the failing autotest is because one test is not being excluded properly, unrelated to this. Okay to merge?\

lucbv:
@cwpearson this is fine to merge.
Also my position is the failing test should not be excluded it should be fixed and then we can promote the build to `Required`!

Issue #2005, opened 2023-10-18
Nightly Blas unit test failures with llvm/14, llvm/15


ndellingwood:
The following Blas unit tests fail in Serial builds with debugging when compiled with llvm/15.0.7, openblas/0.3.23 

**KokkosKernels_blas_serial**
```
./blas/unit_test/KokkosKernels_blas_serial --gtest_filter=serial.asum_complex_double
[ RUN      ] serial.asum_complex_double
/home/ndellin/kokkos-kernels/test_common/KokkosKernels_TestUtils.hpp:159: Failure
Expected: ((double)AT1::abs(val1 - val2)) <= ((double)AT3::abs(tol)), actual: 119.541 vs 1.19541e-05
/home/ndellin/kokkos-kernels/test_common/KokkosKernels_TestUtils.hpp:159: Failure
Expected: ((double)AT1::abs(val1 - val2)) <= ((double)AT3::abs(tol)), actual: 119.541 vs 1.19541e-05
[  FAILED  ] serial.asum_complex_double (54 ms)

...

./blas/unit_test/KokkosKernels_blas_serial --gtest_filter=serial.nrm1_complex_double
[ RUN      ] serial.nrm1_complex_double
/home/ndellin/kokkos-kernels/test_common/KokkosKernels_TestUtils.hpp:159: Failure
Expected: ((double)AT1::abs(val1 - val2)) <= ((double)AT3::abs(tol)), actual: 119.541 vs 1.19541e-05
/home/ndellin/kokkos-kernels/test_common/KokkosKernels_TestUtils.hpp:159: Failure
Expected: ((double)AT1::abs(val1 - val2)) <= ((double)AT3::abs(tol)), actual: 119.541 vs 1.19541e-05
[  FAILED  ] serial.nrm1_complex_double (8 ms)
```

**Reproducer** (blake "all" queue):
```bash
salloc -N 1 -p all

module purge
module load cmake llvm openblas/0.3.23

$KOKKOSKERNELS_PATH/cm_generate_makefile.bash --with-serial --compiler=clang++ --with-tpls=blas,lapack --debug --with-scalars=complex_double,double
```

llvm/15.0.7 is the only version available on blake, but I'll test with different versions on another machine

ndellingwood:
I tested and reproduced this with the following in debug builds:

* llvm/15.0.7 + openblas/0.3.23 (blake)
* llvm/15.0.7 + openblas/0.3.20 (blake)
[...]

PR #2376, opened 2024-10-13
Block Sptrsv fixes

jgfouca:
Go back to the vector reduce implementation and fix block divide to have the right order of operations. With these changes, RBILUK+KK produces the same results as RBILUK (without KK).

These changes unfortunately cause a problem I saw earlier to reemerge. The problem is unique to Clang+CUDA+Complex+Release, so I strongly suspect it's a compiler bug. If you use GCC, it works. If you turn off -03 (-O0 and -O1 work), it works. To get around this, I disabled the test for this specific case.

### DIFF
--- a/sparse/impl/KokkosSparse_sptrsv_solve_impl.hpp
+++ b/sparse/impl/KokkosSparse_sptrsv_solve_impl.hpp
@@ -369,7 +369,8 @@ struct SptrsvWrap {
     KOKKOS_INLINE_FUNCTION
     static void multiply_subtract(const CBlock &A, const CVector &b, Vector &c) {
       // Use gemv. alpha is hardcoded to -1, beta hardcoded to 1
-      KokkosBlas::SerialGemv<KokkosBlas::Trans::NoTranspose, KokkosBlas::Algo::Gemv::Blocked>::invoke(-1.0, A, b, 1.0, c);
+      KokkosBlas::SerialGemv<KokkosBlas::Trans::NoTranspose, KokkosBlas::Algo::Gemv::Blocked>::invoke(-1.0, A, b, 1.0,
+                                                                                                      c);
     }
 
     KOKKOS_INLINE_FUNCTION
--- a/sparse/impl/KokkosSparse_sptrsv_solve_impl.hpp
+++ b/sparse/impl/KokkosSparse_sptrsv_solve_impl.hpp
@@ -324,9 +324,9 @@ struct SptrsvWrap {
       team.team_barrier();
       KokkosBatched::TeamLU<member_type, KokkosBatched::Algo::LU::Blocked>::invoke(team, LU);
 
-      // A = LU
-      // A^-1 = U^-1 * L^-1
-      // b = (b * U^-1) * L^-1, so do U trsv first
+      // Ax = LUx = Lz = b, we use the change of variable z = U*x
+      // z = L^-1 * b, first we solve for z, storing the result back into b
+      // x = U^-1 * z, second we solve for x, again storing the result back into b
       team.team_barrier();
       KokkosBatched::TeamTrsv<member_type, KokkosBatched::Uplo::Lower, KokkosBatched::Trans::NoTranspose,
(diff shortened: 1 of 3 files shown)


cwpearson (changes requested):
I suggest:

1. detect when we're in the scenario that's buggy ( compiler + optimization + algorithm choice, if I understand correctly) and throw if someone tries to call.
2. Change the test so it's explicitly GTEST_SKIP in that case with a comment.

jgfouca:
Thanks, @cwpearson . I added some code to the handle to address your first point. I'm not sure I can do the second point because GTEST_SKIP is for when you are skipping an entire TEST; in this case, we are just skipping a small piece of the overall test.

sparse/impl/KokkosSparse_sptrsv_solve_impl.hpp
```diff
@@ -260,14 +327,14 @@ struct SptrsvWrap {
       // A = LU
       // A^-1 = U^-1 * L^-1
```
lucbv, line 328: The comments here are a bit confusing, especially the line below where you apply the matrix on the left side of `b` which is wrong.
I would say the following:
// Ax = LUx = Lz = b, we use the change of variable z = U*x
// z = L^-1 * b, first we solve for z, storing the result back into b
// x = U^-1 * z, second we solve for x, again storing the result back into b

sparse/impl/KokkosSparse_sptrsv_solve_impl.hpp
```diff
@@ -278,21 +345,44 @@ struct SptrsvWrap {
...
-      KK_KERNEL_REQUIRE_MSG(block_size_ <= MAX_VEC_SIZE,
-                            "Max supported block size for range-policy is 16. Use team-policy alg if you need more.");
-
       Block LU(&buff[0], block_size_, block_size_);
       assign(LU, A);
       KokkosBatched::SerialLU<KokkosBatched::Algo::LU::Blocked>::invoke(LU);
 
       // A = LU
       // A^-1 = U^-1 * L^-1
       // b = (b * U^-1) * L^-1, so do U trsv first
```
lucbv, line 354: Again this is misleading at best.

lucbv (commented):
It looks correct but it would be nice to update the comments so they match the code.

cwpearson:
> Thanks, @cwpearson . I added some code to the handle to address your first point. I'm not sure I can do the second point because GTEST_SKIP is for when you are skipping an entire TEST; in this case, we are just skipping a small piece of the overall test.

Then I can live with printing a message to stderr that says what's going on and why before `continue`

lucbv:
@cwpearson I do not believe that the version of gtest snapshotted in our repo (1.7.0) supports gtest_skip?
We could move to a new gtest as kokkos core is using 1.11.0 and Trilinos is using 1.10.0

jgfouca:
@cwpearson , @nellingwood , I see the one CI job failed:

```
/home/runner/_work/kokkos-kernels/kokkos-kernels/kokkos-kernels/test_common/KokkosKernels_TestUtils.hpp:140: Failure
Expected: ((double)AT1::abs(val1 - val2)) <= ((double)AT3::abs(tol)), actual: 119.541 vs 1.19541e-05
/home/runner/_work/kokkos-kernels/kokkos-kernels/kokkos-kernels/test_common/KokkosKernels_TestUtils.hpp:140: Failure
Expected: ((double)AT1::abs(val1 - val2)) <= ((double)AT3::abs(tol)), actual: 119.541 vs 1.19541e-05
[  FAILED  ] serial.asum_complex_double (0 ms)
```

This is not code that I touched in this PR; is this an expected fail on develop? Maybe it's intermittent?

cwpearson:
The failures are unrelated: https://github.com/kokkos/kokkos-kernels/issues/2005

cwpearson (approved):

jgfouca:
Everything looks good now. @lucbv , would you mind merging this?

lucbv (approved):
Yep, looks good to me

Issue #2379, opened 2024-10-15
`KOKKOSKERNELS_ENABLE_COMPONENT_{whatever}` is always defined


cwpearson:
This causes problems e.g. here:

https://github.com/kokkos/kokkos-kernels/blob/130631f240bc840489f0963dc7379ca09a878fd3/common/src/KokkosKernels_EagerInitialize.cpp#L22

The error is here, the variables are set to a string like "KokkosKernels_ENABLE_COMPONENT_{whatever}" instead of the *value* of that variable.

https://github.com/kokkos/kokkos-kernels/blob/130631f240bc840489f0963dc7379ca09a878fd3/cmake/kokkoskernels_components.cmake#L108-L113

PR #2380, opened 2024-10-15
set `KokkosKernels_ENABLE_COMPONENT` variables to value instead of variable name

cwpearson:
Fix for https://github.com/kokkos/kokkos-kernels/issues/2379


lucbv (approved):
Yeah, that's a good catch

cwpearson:
@lucbv we can probably merge this one?

PR #2383, opened 2024-10-15
Prefer `atomic_assign(ptr, val) -> atomic_store(ptr, val)`

dalg24:
We are looking at deprecating atomic_assign() https://github.com/kokkos/kokkos/issues/7449 Use atomic_store() instead.


cwpearson (approved):
Approving so we can run CI

lucbv (approved):

PR #2384, opened 2024-10-16
Blas1 asum: workaround for openblas error with short vectors

cwpearson:
Fix for https://github.com/kokkos/kokkos-kernels/issues/2005


lucbv:
So you really don't think the idea of directly implementing asum instead of allocating, filling and dispatching?

cwpearson:
Let's do it your better way.

lucbv (approved):
Nice, lgtm!

PR #2388, opened 2024-10-17
Reduce duplicated code in trsv

jgfouca:
A slightly different for loop can avoid having to separate out the last iteration in all the uppertri functions.

### DIFF
--- a/sparse/impl/KokkosSparse_trsv_impl.hpp
+++ b/sparse/impl/KokkosSparse_trsv_impl.hpp
@@ -261,10 +261,8 @@ struct TrsvWrap {
       return;
     }
 
-    // Don't use r >= 0 as the test, because that fails if
-    // lno_t is unsigned.  We do r == 0 (last
-    // iteration) below.
-    for (lno_t r = numRows - 1; r != 0; --r) {
+    // Iterate backwards with care due to potentially unsigned type
+    for (lno_t r = numRows; r-- > 0;) {
       const offset_type beg = ptr(r);
       const offset_type end = ptr(r + 1);
       for (offset_type k = beg; k < end; ++k) {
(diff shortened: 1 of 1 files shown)


cwpearson (approved):

cwpearson (changes requested):
Is `upperTriSolveCscConj` right? Copy-paste error from the previous?

jgfouca:
@cwpearson , I think it's right. The c==0 part I removed assumed there was only one item, which should hold true in both cases.

cwpearson:
@jgfouca we merged a fix for the failing atomics, I think if you rebase on develop the current failures will go away

sparse/impl/KokkosSparse_trsv_impl.hpp
```diff
@@ -261,10 +261,8 @@ struct TrsvWrap {
       return;
     }
 
-    // Don't use r >= 0 as the test, because that fails if
-    // lno_t is unsigned.  We do r == 0 (last
-    // iteration) below.
-    for (lno_t r = numRows - 1; r != 0; --r) {
+    // Iterate backwards with care due to potentially unsigned type
+    for (lno_t r = numRows; r-- > 0;) {
```
lucbv, line 265: I am not a big fan of `r-- > 0` because it is a bit harder to understand but does not really make the code much smaller or more performant. Also is the loop now starting at `numRows` instead of `numRows - 1`?

lucbv (changes requested):
Reducing the amount of code is a good thing for maintenance, but I would rather make the loop control flow a little simpler to avoid misunderstanding by next developer who maintains the code.

jgfouca:
@lucbv , so do you want me to just close this PR?

sparse/impl/KokkosSparse_trsv_impl.hpp
jgfouca: > Also is the loop now starting at numRows instead of numRows - 1?

Yes, because `r` gets `--`'d before it gets used in the loop body.

lucbv: So what would be the issue with just rewriting this as: `for (lno_t r = numRows; r > 0; --r)` and then just index with `r - 1`?
Does it lead to the same issue with the case `r == 0`

jgfouca: @lucbv , I think that would work too but you'd need to change all the indexes. Did you see my static_cast option below?

jgfouca:
@lucbv , see if you like the current version better.

lucbv (approved):

lucbv:
@cwpearson are you okay with this, can we merge?

cwpearson (approved):

PR #2389, opened 2024-10-18
workflows: adding variable to set kokkos version to test against

lucbv:
The variable is set directly in at2.yml so it can be uniformly applied to all the "sub-workflows" making maintenance when a new release of kokkos comes out easier.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -19,7 +19,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  KOKKOS_VERSION: 4.4.01
+  kokkos_version: 4.4.01
 
 jobs:
   check-pr-labels:
@@ -61,7 +61,7 @@ jobs:
         uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
         with:
           repository: kokkos/kokkos
-          ref: $KOKKOS_VERSION
+          ref: ${{ env.kokkos_version }}
           path: kokkos
 
       - name: configure_kokkos
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -18,6 +18,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  KOKKOS_VERSION: 4.4.01
+
 jobs:
   check-pr-labels:
     runs-on: [ubuntu-latest]
@@ -58,7 +61,7 @@ jobs:
         uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
         with:
           repository: kokkos/kokkos
-          ref: 4.3.01
+          ref: $KOKKOS_VERSION
           path: kokkos
 
       - name: configure_kokkos
--- a/.github/workflows/at2.yml
+++ b/.github/workflows/at2.yml
@@ -18,14 +18,27 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  KOKKOS_VERSION: 4.4.01
+
 jobs:
   mi210:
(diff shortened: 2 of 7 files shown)


cwpearson (approved):
This is fine with me for the vast majority of our CI. It might be good to have one serial build against the previous release too if we want to support that.

cwpearson:
I think this variable is not propagating properly. If you look at old actions the checkout part has a `ref:` field that is now missing, so we're pulling kokkos develop everywhere

cwpearson:
May need to do something like this:

```yaml
jobs:
  call-workflow:
    uses: ./.github/workflows/reusable-workflow.yml
    with:
      my_var: ${{ vars.MY_VAR }}
```

lucbv:
Yeah, I am looking at the documentation to get the appropriate github action syntax to propagate the variable correctly...

lucbv:
@cwpearson it took me a minute to catch on what you were saying in your comment, I eventually read the documentation and figured out the syntax. Once small thing that remains unsatisfactory: we need to change the version in both `at2.yml` and `osx.yml`. I don't think I want to fix that now though...

lucbv:
Ah and feel free to double check that everything does indeed build against kokkos 4.4.1.
Thanks!

ndellingwood (approved):
Thanks @lucbv ! I thought the version was already pinned, but I guess that was for some the jobs

ndellingwood:
The osx-ci jobs look unhappy about fetching the version,
```
  /opt/homebrew/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin +refs/heads/$KOKKOS_VERSION*:refs/remotes/origin/$KOKKOS_VERSION* +refs/tags/$KOKKOS_VERSION*:refs/tags/$KOKKOS_VERSION*
  The process '/opt/homebrew/bin/git' failed with exit code 1
  Waiting 17 seconds before trying again
  /opt/homebrew/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin +refs/heads/$KOKKOS_VERSION*:refs/remotes/origin/$KOKKOS_VERSION* +refs/tags/$KOKKOS_VERSION*:refs/tags/$KOKKOS_VERSION*
  The process '/opt/homebrew/bin/git' failed with exit code 1
  Waiting 15 seconds before trying again
  /opt/homebrew/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin +refs/heads/$KOKKOS_VERSION*:refs/remotes/origin/$KOKKOS_VERSION* +refs/tags/$KOKKOS_VERSION*:refs/tags/$KOKKOS_VERSION*
  Error: The process '/opt/homebrew/bin/git' failed with exit code 1
```

.github/workflows/osx.yml
```diff
@@ -58,7 +61,7 @@ jobs:
         uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
         with:
           repository: kokkos/kokkos
-          ref: 4.3.01
+          ref: $KOKKOS_VERSION
```
ndellingwood, line 64: osx-ci doesn't like this variable, maybe braces are needed?
```suggestion
          ref: ${KOKKOS_VERSION}
```

lucbv: Yep that looks reasonable!

lucbv: I'll do a separate commit as I am not sure how to sign-off otherwise...

ndellingwood: I think since the checkout is performed by an action this made including the context necessary (if it was handled on the runner then the Linux shell syntax would've been fine), [using-the-env-context-to-access-environment-variable-values](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#using-the-env-context-to-access-environment-variable-values)

ndellingwood:
@lucbv osx builds look happier now :)
I checked the output on one of the jobs to confirm it was using kokkos@4.4.01 as expected

Issue #2374, opened 2024-10-10
KokkosBatched_Util.hpp: macro `__KOKKOSBATCHED_PROMOTION__` uses a reserved name


cwpearson:
https://github.com/kokkos/kokkos-kernels/blob/c7df61d536c74effb3e4466b011f67f24e85f314/batched/KokkosBatched_Util.hpp#L22

Potential Trilinos integration issues with renaming this?

Issue #2364, opened 2024-10-07
Avoid reserved identifiers


dalg24:
* [x] Identifiers starting with two consecutive underscores are reserved in C++ (meaning user-defined program are not allowed to use them)
  e.g. 
  ```
  batched/dense/impl/KokkosBatched_Axpy_Impl.hpp:16:#ifndef __KOKKOSBATCHED_AXPY_IMPL_HPP__
  batched/dense/impl/KokkosBatched_Axpy_Impl.hpp:17:#define __KOKKOSBATCHED_AXPY_IMPL_HPP__
  ```
  or
  ```
  batched/dense/impl/KokkosBatched_HostLevel_Gemm_Armpl_Impl.hpp:74:  HandleType *const __handle;
  batched/dense/impl/KokkosBatched_HostLevel_Gemm_Armpl_Impl.hpp:80:  AViewType __A;
  batched/dense/impl/KokkosBatched_HostLevel_Gemm_Armpl_Impl.hpp:81:  avt *__Adp = nullptr;
  batched/dense/impl/KokkosBatched_HostLevel_Gemm_Armpl_Impl.hpp:82:  armpl_int_t __Ajstrd, __Aistrd, __Abstrd;
  batched/dense/impl/KokkosBatched_HostLevel_Gemm_Armpl_Impl.hpp:84:  BViewType __B;
  ```
* [x] Identifiers starting with an underscore followed by an uppercase letter are reserved in C++
  For example
  ```
  batched/dense/impl/KokkosBatched_HostLevel_Gemm_Serial_Impl.hpp:131:  BatchedSerialGemm(ScalarType _alpha, AViewType _A, BViewType _B, ScalarType _beta, CViewType _C)
  batched/dense/impl/KokkosBatched_HostLevel_Gemm_Serial_Impl.hpp:132:      : A(_A), B(_B), C(_C), alpha(_alpha), beta(_beta) {}
  ```
* [x] Avoid defining macros starting with `KOKKOS_` (this includes header guards)
  ```
  blas/impl/KokkosBlas_util.hpp:18:#define KOKKOS_BLAS_UTIL_HPP
  blas/tpls/KokkosBlas_tpl_spec.hpp:89:#define KOKKOS_CUBLAS_SAFE_CALL_IMPL(call) KokkosBlas::Impl::cublas_internal_safe_call(call, #call, __FILE__, __LINE__)
  ```
[...]

PR #2392, opened 2024-10-21
deprecate `__KOKKOSBATCHED_PROMOTION__`

cwpearson:
Fix for https://github.com/kokkos/kokkos-kernels/issues/2374: this macro is a reserved identifier, so it needs to go. See also https://github.com/kokkos/kokkos-kernels/issues/2364

It has no uses in Trilinos, but I'm deprecating this rather than outright removing it because the comment hints at some trilinos integration thing.

### DIFF
--- a/batched/KokkosBatched_Util.hpp
+++ b/batched/KokkosBatched_Util.hpp
@@ -19,7 +19,19 @@
 /// \author Kyungjoo Kim (kyukim@sandia.gov)
 
 // no experimental name space guard for trilinos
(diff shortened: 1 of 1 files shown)


lucbv (commented):
I see a lot more underscore than I am comfortable with, can we name the new `__KOKKOSBATCHED_PROMOTION___DEPRECATED_MACRO` something simpler and with less underscores?

ndellingwood (approved):

lucbv (approved):

PR #2394, opened 2024-10-21
workflow: Actually using the KOKKOS_VERSION variable

lucbv:
Currently the version is still hardcoded for each job in at2.yml Now switching to use the control variable so all the jobs can be in sync and switched at once for easier maintenance.

### DIFF
--- a/.github/workflows/at2.yml
+++ b/.github/workflows/at2.yml
@@ -25,20 +25,20 @@ jobs:
   mi210:
     uses: ./.github/workflows/mi210.yml
     with:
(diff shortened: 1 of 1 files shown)


cwpearson (approved):

ndellingwood (approved):

lucbv:
Unfortunately the work around to get this working is quite more cumbersome than I would have expected see [here](https://github.com/orgs/community/discussions/26671). At this point I think we should keep the current workflow that does not use a variable but simply hard codes the value of the Kokkos version to test against. : (

PR #2395, opened 2024-10-23
Implement batched serial laswp

yasahi-hpc:
This PR implements [laswp](https://www.netlib.org/lapack/explore-html/d1/d7e/group__laswp_ga5d3ea3e3cb61e32750bf062a2446aa33.html) function, which is needed for [getrf PR](https://github.com/kokkos/kokkos-kernels/pull/2331).

Following files are added:
1. `KokkosBatched_Laswp_Serial_Impl.hpp`: Internal interfaces with implementation details
2. `KokkosBatched_Laswp.hpp`: APIs
3. `Test_Batched_SerialLaswp.hpp`: Unit tests for that

## Detailed description
This performs a series of row interchanges on a general rectangular matrix.
- `A`: `(batch_count, m, n)` or `(batch_count, m)`
  On entry, the M-by-N matrix or the length M vector. The row interchanges will be applied to the matrix of column dimension N. On exit, the permuted matrix or vector.
- `IPIV`: `(batch_count, m)`
  The pivot indices; for `0 <= i < m`, row `i` of the matrix was interchanged with row `IPIV(i)`.

```C++
Kokkos::parallel_for('laswp', 
    [=](const int k) {
        auto sub_ipiv = Kokkos::subview(m_ipiv, k, Kokkos::ALL());
        auto sub_a = Kokkos::subview(m_a, k, Kokkos::ALL(), Kokkos::ALL());
        KokkosBatched::SerialLaswp<typename ParamTagType::direct>::invoke(sub_ipiv, sub_a);
    });
```

## Tests
1.  Make a random matrix from random `A` and `ipiv`. Apply `Laswp` to `A` while preparing a reference `Ref` from `A` which is permuted by `ipiv`. Then confirm `A` == `Ref`.

### DIFF
--- a/batched/dense/impl/KokkosBatched_Laswp_Serial_Internal.hpp
+++ b/batched/dense/impl/KokkosBatched_Laswp_Serial_Internal.hpp
@@ -53,18 +53,35 @@ struct SerialLaswpMatrixForwardInternal {
   template <typename IntType, typename ValueType>
   KOKKOS_INLINE_FUNCTION static int invoke(const int n, const int plen, const IntType *KOKKOS_RESTRICT p, const int ps0,
                                            /* */ ValueType *KOKKOS_RESTRICT A, const int as0, const int as1) {
-    for (int j = 0; j < n; j++) {
-      ValueType *KOKKOS_RESTRICT A_at_j = A + j * as1;
+    if (as0 <= as1) {
+      // LayoutLeft like
+      for (int j = 0; j < n; j++) {
+        ValueType *KOKKOS_RESTRICT A_at_j = A + j * as1;
+        for (int i = 0; i < plen; ++i) {
+          const int piv = p[i * ps0];
+          if (piv != i) {
+            const int idx_i = i * as0, idx_p = piv * as0;
+            const ValueType tmp = A_at_j[idx_i];
+            A_at_j[idx_i]       = A_at_j[idx_p];
+            A_at_j[idx_p]       = tmp;
+          }
+        }
+      }
+    } else {
+      // LayoutRight like
       for (int i = 0; i < plen; ++i) {
         const int piv = p[i * ps0];
         if (piv != i) {
           const int idx_i = i * as0, idx_p = piv * as0;
-          const ValueType tmp = A_at_j[idx_i];
-          A_at_j[idx_i]       = A_at_j[idx_p];
-          A_at_j[idx_p]       = tmp;
+          for (int j = 0; j < n; j++) {
+            ValueType *KOKKOS_RESTRICT A_at_j = A + j * as1;
+            const ValueType tmp               = A_at_j[idx_i];
+            A_at_j[idx_i]                     = A_at_j[idx_p];
+            A_at_j[idx_p]                     = tmp;
+          }
         }
       }
-    };
+    }
     return 0;
   }
 };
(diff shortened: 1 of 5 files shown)


batched/dense/impl/KokkosBatched_Laswp_Serial_Internal.hpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+    return 0;
+  }
+};
+
+struct SerialLaswpMatrixForwardInternal {
+  template <typename IntType, typename ValueType>
+  KOKKOS_INLINE_FUNCTION static int invoke(const int n, const int plen, const IntType *KOKKOS_RESTRICT p, const int ps0,
+                                           /* */ ValueType *KOKKOS_RESTRICT A, const int as0, const int as1) {
+    for (int j = 0; j < n; j++) {
```
lucbv, line 56: Are we assuming layout left? Otherwise it would make sense for a layout right to inverse these two loops.

lucbv (commented):
Looks good overall, I would have implemented two more tests. 

1. As usual an analytical test were we hard code the original vector and its permutation and check the output of `laswp` is correct.
2. You could compute ipiv_inv, defined as `ipiv_inv(ipiv(i))=i;`, then perform laswap with ipiv followed by laswap with ipiv_inv and check that this returns the original vector. Can be done by simply keeping a copy of the original vector...

batched/dense/impl/KokkosBatched_Laswp_Serial_Internal.hpp
yasahi-hpc: Actually, the expected layout is `LayoutStrided` since `A` is supposed to be a subview of batches of `A`. Based on your comments, I have changed the loop order by comparing the strides of `A` (assuming that `A` is originated from `LayoutLeft` or `LayoutRight`).

batched/dense/unit_test/Test_Batched_SerialLaswp.hpp
```diff
@@ -0,0 +1,509 @@
...
+    }
+  }
+
+  inline int run() {
+    using value_type = typename AViewType::non_const_value_type;
+    std::string name_region("KokkosBatched::Test::SerialLaswp");
+    const std::string name_value_type = Test::value_type_name<value_type>();
+    std::string name                  = name_region + name_value_type;
+    int info_sum                      = 0;
+    Kokkos::Profiling::pushRegion(name.c_str());
```
cwpearson, line 57: This profiling region can be removed (matches the inner `parallel_reduce` exactly)

batched/dense/unit_test/Test_Batched_SerialLaswp.hpp
```diff
@@ -0,0 +1,509 @@
...
+  info0 = Functor_BatchedSerialLaswp<DeviceType, PivView2DType, View2DType, InvDirect>(ipiv0, A0_identity).run();
+  info1 = Functor_BatchedSerialLaswp<DeviceType, PivView2DType, View3DType, InvDirect>(ipiv1, A1_identity).run();
+  info2 = Functor_BatchedSerialLaswp<DeviceType, PivView2DType, View3DType, InvDirect>(ipiv2, A2_identity).run();
+
+  Kokkos::fence();
+  EXPECT_EQ(info0, 0);
+  EXPECT_EQ(info1, 0);
+  EXPECT_EQ(info2, 0);
+
+  // this eps is about 10^-14
```
cwpearson, line 228: I think it depends on the scalar type 😄

cwpearson (approved):
A couple comments, not enough to block merging.

lucbv (approved):

PR #2397, opened 2024-10-24
Fixing potential overflow issue in inner product trait

lucbv:
When result type is double and inputs are floats, one input has to be cast to double so the multiplication operator for double is used instead of the float multiplication operator that could overflow for valid double values.

@csiefer2 this should take care of the security issue you reported.

### DIFF
--- a/batched/dense/impl/KokkosBatched_HostLevel_Gemm_DblBuf_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_HostLevel_Gemm_DblBuf_Impl.hpp
@@ -249,8 +249,7 @@ class BatchedDblBufGemm {
     CViewType __C;
     ScalarType __alpha, __beta;
     int __k;
-    size_t __n_sub_tiles;
-    unsigned __tiles_per_col, __tiles_per_row;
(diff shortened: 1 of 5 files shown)


yasahi-hpc (approved):
LGTM
Reasonable to avoid a potential overflow.

Oh, I have one question about complex overload.

```C++
KOKKOS_INLINE_FUNCTION void updateDot(Kokkos::complex<double>& sum, const Kokkos::complex<float> x,
                                      const Kokkos::complex<float> y) {
  const auto tmp = Kokkos::conj(x) * y;
  sum += Kokkos::complex<double>(tmp.real(), tmp.imag());
}
```

Do you need to apply casting for this case like

```C++
KOKKOS_INLINE_FUNCTION void updateDot(Kokkos::complex<double>& sum, const Kokkos::complex<float> x,
                                      const Kokkos::complex<float> y) {
  const auto tmp_x = Kokkos::complex<double>(x.real(), x.imag());
  const auto tmp_y = Kokkos::complex<double>(y.real(), y.imag());
  sum += Kokkos::conj(tmp_x) * tmp_y;
}
```

brian-kelley (approved):

PR #2398, opened 2024-10-25
worflows: increasing concurrency for builds on modern CPUs

lucbv:
Trying to increase concurrency to see if that improves the build time for mi210 actions


lucbv:
This is not really improving performance in a meaningful way, there is probably something else going on in the MI210 build. Will close this and try something different next...

PR #2399, opened 2024-10-27
implement batched serial iamax

yasahi-hpc:
This PR implements [iamax](https://netlib.org/lapack/explore-html/dd/d52/group__iamax_gacec03c5109f531c06b4fb301cf1a2d7a.html#gacec03c5109f531c06b4fb301cf1a2d7a) function, which is needed for [getrf PR](https://github.com/kokkos/kokkos-kernels/pull/2331).

Following files are added:
1. `KokkosBatched_Iamax_Serial_Impl.hpp`: Internal interfaces with implementation details
2. `KokkosBatched_Iamax.hpp`: APIs
3. `Test_Batched_SerialIamax.hpp`: Unit tests for that

## Detailed description
This returns the index of the first element having maximum absolute value.
- `X`: `(batch_count, n)`
  The length N vector.

```C++
Kokkos::parallel_for('iamax', 
    [=](const int k) {
        auto iamax = KokkosBatched::SerialIamax::invoke(sub_x);
        m_r(k)     = iamax;
    });
```

## Tests
1.  Analytical tests for length 3 vectors 
```C++
A0: [1, 2, 0] -> 1
A1: [-5, 4, 3] -> 0
A2: [0, 0, 0] -> 0
A3: [0, -1, -1] -> 1
```
2. Make a random vector from random `X`. Compare the return value with the index of the first element having maximum absolute value.

### DIFF
--- a/batched/dense/impl/KokkosBatched_Iamax_Serial_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Iamax_Serial_Impl.hpp
@@ -1,0 +1,35 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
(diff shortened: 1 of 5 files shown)


lucbv (approved):
Looks fine, the only small question would be: "Are we worried about views containing more than 2B elements with an int return type?"

yasahi-hpc:
> Looks fine, the only small question would be: "Are we worried about views containing more than 2B elements with an int return type?"

Assuming that this is a `KokkosBatched` functionality, return value should be smaller than 2B elements for a practical use case. However, if you are thinking of security issue like https://github.com/kokkos/kokkos-kernels/pull/2397, it makes sense to use `in64_t`. If you prefer, I can fix it accordingly.

yasahi-hpc:
Sure.
I will rely on the `index_type` of a View.

lucbv (approved):
Looks pretty good to me!

PR #2404, opened 2024-10-28
Fix batched serial laswp

yasahi-hpc:
This PR aims at fixing some issues related to #2395 

- [x] Fix include path in `batched/dense/src/KokkosBatched_Laswp.hpp`
- [x] Add tests for incx argument
- [x] Fix some comments in tests


lucbv (approved):
Okay, the header was obviously a problem, should have caught it in the last PR.
The unit test improvements for strided layouts are good.

PR #2405, opened 2024-10-28
Mult result conversion

lucbv:


graph/impl/KokkosGraph_Distance2Color_impl.hpp
```diff
@@ -616,7 +616,7 @@ class GraphColorDistance2 {
     const lno_t numVerts = this->nr;
     const lno_t numCols  = this->nc;
     // note: relying on forbidden and colors_out being initialized to 0
-    forbidden_view forbidden("Forbidden", batch * numCols);
+    forbidden_view forbidden("Forbidden", static_cast<typename forbidden_view::size_type>(batch) * numCols);
```
brian-kelley, line 619: Getting the size_type for each specific view is correct but really hurts readability. What about just using size_t for these?
```suggestion
    forbidden_view forbidden("Forbidden", size_t(batch) * numCols);
```

sparse/src/KokkosSparse_IOUtils.hpp
```diff
@@ -73,7 +73,8 @@ void kk_sparseMatrix_generate(OrdinalType nrows, OrdinalType ncols, SizeType &nn
   }
   // Sample each value from uniform (-50, 50) for real types, or (-50 - 50i, 50
   // + 50i) for complex types.
-  Kokkos::View<ScalarType *, Kokkos::HostSpace> valuesView(values, nnz * block_elem_count);
+  Kokkos::View<ScalarType *, Kokkos::HostSpace> valuesView(
+      values, nnz * static_cast<typename Kokkos::View<ScalarType *, Kokkos::HostSpace>::size_type>(block_elem_count));
```
brian-kelley, line 77:
```suggestion
  Kokkos::View<ScalarType *, Kokkos::HostSpace> valuesView(
      values, size_t(nnz) * block_elem_count);
```
?

graph/impl/KokkosGraph_Distance2Color_impl.hpp
lucbv: Are we sure that `size_t` is always the right thing to do? I do not think that in practice there are architectures that provide `long long int` that are really larger than `size_t`?
For this function we could locally define `using size_type = typename forbidden_view::size_type` to make the cast a bit more readable. Regarding the c-style vs. cpp-style cast, I don't think that it's fair to favor one over the other unless we make it a rule for the library?

sparse/src/KokkosSparse_IOUtils.hpp
lucbv: Here we could actually try to use the `SizeType` template parameter but that could lead to issues if the user picks something like `int` for that type? We could also define an `impl_size_type` that is not visible to the user and picks the right type?

graph/impl/KokkosGraph_Distance2Color_impl.hpp
brian-kelley: size_t will always be OK when computing anything related to memory, which these examples are. Kokkos hardcodes size_t for extents/strides already (even if a view's size_type is different). It makes sense because size_t can always represent the size of any block of memory.

I don't care that much about ``type(x)`` vs ``(type) x`` vs ``static_cast<type>(x)``, though static_cast is a little verbose and cluttered.

brian-kelley: I asked @crtrott about this and it turns out ``size_type`` of a view is intended for iterating over a view efficiently, so it may be smaller than ``size_t`` but won't ever be larger.

cwpearson (approved):

graph/impl/KokkosGraph_Distance2Color_impl.hpp
lucbv: Okay switching to `static_cast<size_t>()` for more clarity

cwpearson (approved):

PR #2407, opened 2024-10-29
Free allocated `MatrixPrec`

cwpearson:
Let's set a good example in our examples


ndellingwood (approved):

lucbv:
Actually that should be something we can catch with the memory checks CI : )

lucbv (approved):

PR #2408, opened 2024-10-29
CI: `address` sanitizer and most of `undefined` sanitizer

cwpearson:
Exclude vptr due to Preconditioner visibility.
Exclude signed integer overflow because we do this all over the place (future work?)

Needs https://github.com/kokkos/kokkos-kernels/pull/2407 before it will pass.


.github/workflows/linux.yml
```diff
@@ -0,0 +1,110 @@
...
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: docker://agilepathway/pull-request-label-checker@sha256:ee57b0e1aedab22063ce6467a6e6358e254a9204693ca20d8a16b2d891db8d5f  # v1.6.32
+        with:
+          none_of: 'AT: WIP'
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+  sanitizers:
+    needs: check-pr-labels
+    # TODO: allow re-run via retest label if: ${{ github.event.label.name == 'AT: RETEST' }}
+    name: osx-ci
```
lucbv, line 35: Probably a bad name considering that it's not running on osx?

ndellingwood, line 35: Looks like the name needs an update, something indicative of asan check on ubuntu?

ndellingwood (approved):
Thanks for working on getting this cleaned up and ready to run @cwpearson , I think only an update to the name is needed. Added tentative approval so the PR is not blocked on approval

.github/workflows/linux.yml
```diff
@@ -0,0 +1,110 @@
...
+    runs-on: [ubuntu-latest]
+
+    strategy:
+      matrix:
+        include:
+          - backend: "SERIAL"
```
lucbv, line 41: Do we plan to expand this further? Otherwise maybe the matrix part can be removed and the variable associated with it can be hard coded directly in the configuration options?

.github/workflows/linux.yml
ndellingwood, line 35:
```suggestion
    name: ubuntu-asan-ci
```
Maybe this?

ndellingwood:
This will need the AT2-SPECIAL-APPROVAL tag when ready to test (post-merge of https://github.com/kokkos/kokkos-kernels/pull/2407)

.github/workflows/linux.yml
```diff
@@ -0,0 +1,110 @@
...
+          -DKokkos_ENABLE_COMPILER_WARNINGS=ON \
+          -DKokkos_ENABLE_DEBUG_BOUNDS_CHECK:BOOL=${{ matrix.debug_bounds_check }} \
+          -DKokkos_ENABLE_DEPRECATED_CODE_3=OFF \
+          -DKokkos_ENABLE_TESTS=OFF \
+          -DKokkos_ENABLE_DEPRECATED_CODE_4=OFF \
+          -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+          -DCMAKE_CXX_FLAGS="-fsanitize=address -fsanitize=undefined -fno-sanitize=signed-integer-overflow -fno-sanitize=vptr -fno-sanitize-recover=all" \
+          -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address -fsanitize=undefined -fno-sanitize=signed-integer-overflow -fno-sanitize=vptr -fno-sanitize-recover=all" \
+          -DCMAKE_INSTALL_PREFIX=$PWD/../install \
+          ..
```
lucbv, line 75: I like the -S and -B options better if that's no trouble.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: build_and_install_kokkos
+        working-directory: kokkos/build
+        run: make -j$(nproc) install
```
lucbv, line 79: How about `cmake --build /path/to/build/dir -j$(nproc)` that way if we switch to ninja one day it will work out of the box : )
It would require a second line for `cmake --install /path/to/build/dir --prefix /path/to/install/dir` I believe the --prefix is only used if you want to override the `CMAKE_INSTALL_PREFIX` option

.github/workflows/linux.yml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: configure_kokkos_kernels
+        run: |
+          mkdir -p kokkos-kernels/{build,install}
+          cd kokkos-kernels/build
+          cmake \
+          -DKokkos_DIR=$PWD/../../kokkos/install/lib/cmake/Kokkos \
```
lucbv, line 86: Please prefer `Kokkos_ROOT` otherwise we could run into trouble with the lib vs. lib64 path depending on platforms.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,110 @@
...
+          -DKokkosKernels_ENABLE_EXAMPLES:BOOL=ON \
+          -DKokkosKernels_INST_COMPLEX_DOUBLE=ON \
+          -DKokkosKernels_INST_DOUBLE=ON \
+          -DKokkosKernels_INST_COMPLEX_FLOAT=ON \
+          -DKokkosKernels_INST_FLOAT=ON \
+          -DKokkosKernels_INST_LAYOUTLEFT:BOOL=ON \
+          -DKokkosKernels_INST_LAYOUTRIGHT:BOOL=ON \
+          -DKokkosKernels_INST_OFFSET_INT=ON \
+          -DKokkosKernels_INST_OFFSET_SIZE_T=ON \
+          -DCMAKE_CXX_FLAGS="-fsanitize=address -fsanitize=undefined -fno-sanitize=signed-integer-overflow -fno-sanitize=vptr -fno-sanitize-recover=all" \
```
lucbv, line 100: Does this override or add to the previously defined `CMAKE_CXX_FLAGS` on line 88

.github/workflows/linux.yml
```diff
@@ -0,0 +1,110 @@
...
+          -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address -fsanitize=undefined -fno-sanitize=signed-integer-overflow -fno-sanitize=vptr -fno-sanitize-recover=all" \
+          ..
+
+      - name: build_kokkos_kernels
+        working-directory: kokkos-kernels/build
+        run: make -j$(nproc)
```
lucbv, line 106: Same remark as the one for kokkos core build/install

lucbv (changes requested):
Looks good overall but left some comments for portability improvements.

lucbv:
We do not really need to set the special approval here as we do not really need to run the AT2 CI, I can merge when ready.

.github/workflows/linux.yml
cwpearson: I agree in spirit but I chose a different name since it's asan and ubsan

.github/workflows/linux.yml
cwpearson: Maybe a similar build just with clang instead? I think they have different implementations but the same interface of these sanitizers.

.github/workflows/linux.yml
cwpearson: yes. I'll just drop the warning flags since this build is focused on runtime checks.

lucbv (approved):
Looks good to me.

cwpearson:
I signed off on all the commits

lucbv (approved):
Okay this latest version looks good, let's just wait for CI to finish and then merge.

PR #2410, opened 2024-10-30
SerialInverseLU: fix overflow in integer multiplication

lucbv:
Last one of a series of fixes to clean-up the CodeQL safety issues, after that we should be all clean!

### DIFF
--- a/batched/dense/unit_test/Test_Batched_SerialInverseLU.hpp
+++ b/batched/dense/unit_test/Test_Batched_SerialInverseLU.hpp
@@ -143,7 +143,7 @@ void impl_test_batched_inverselu(const int N, const int BlkSize) {
   /// randomized input testing views
   AViewType a0("a0", N, BlkSize, BlkSize);
   AViewType a1("a1", N, BlkSize, BlkSize);
-  WViewType w("w", N, BlkSize * BlkSize);
+  WViewType w("w", N, BlkSize * static_cast<size_t>(BlkSize));
   AViewType c0("c0", N, BlkSize, BlkSize);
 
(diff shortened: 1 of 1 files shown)


batched/dense/unit_test/Test_Batched_SerialInverseLU.hpp
```diff
@@ -143,7 +143,7 @@ void impl_test_batched_inverselu(const int N, const int BlkSize) {
   /// randomized input testing views
   AViewType a0("a0", N, BlkSize, BlkSize);
   AViewType a1("a1", N, BlkSize, BlkSize);
-  WViewType w("w", N, BlkSize * BlkSize);
+  WViewType w("w", N, BlkSize * static_cast<size_t>(BlkSize));
```
vqd8a, line 146: @lucbv Why having a fix for only ```w```? How about ```a0```, ```a1```?

lucbv: That one actually does `BlkSize * BlkSize` and since both are `int`, the temporary result is also `int` but the input parameter for the view is `size_t` so there is an overflow potential before we store the result. For the other views there is no multiplication so implicit casting is working just fine.
Does that make sense?

vqd8a: @Thanks. Sorry I did not see the commas in a0 and a1 :-)

lucbv: No worries!

ndellingwood (approved):

PR #2415, opened 2024-11-05
remove unused `FillSymmetricLowerEdgesHashMap::lower_only`

cwpearson:
This field seems to be both unset and unused.

Possible fix for something caught in UB sanitizer.


lucbv (approved):
Reasonable

cwpearson:
It seems like this did fix one UB sanitizer error!

PR #2416, opened 2024-11-05
Use fused gtest v1.11.0 from Kokkos

cwpearson:
I have included each change Kokkos made in its own commit, without comment or justification:

* https://github.com/kokkos/kokkos/pull/4230
* https://github.com/kokkos/kokkos/pull/4406
* https://github.com/kokkos/kokkos/pull/4599
* https://github.com/kokkos/kokkos/pull/4811
* https://github.com/kokkos/kokkos/pull/6328

Looks like this version is a bit pickier about types in some comparisons.


cwpearson:
@lucbv our format check just applies clang-format to every changed file, not a list of files in the repo, so it's expected to fail here since we don't format this vendored code.

lucbv (approved):

ndellingwood:
This will need one follow up to resolve some -Werror in Trilinos nightly integration builds, e.g.

```
/scratch/trilinos/workspace/Nightly/Trilinos_nightly_pipeline/Trilinos/kokkos-kernels/tpls/gtest/gtest/gtest.h:11427:11: error: comparison of integer expressions of different signedness: 'const long unsigned int' and 'const long int' [-Werror=sign-compare]
   if (lhs == rhs) {
```

https://trilinos-cdash.sandia.gov/viewBuildError.php?buildid=1952053

PR #2420, opened 2024-11-06
Two-stage GS: initialize functor bool members

brian-kelley:
Follow on to #2415 and #2417 - fix one more UBsan error. This one happened in the CI build, but not my local build.
It's the same kind of error, where bool members of a struct were left uninitialized.


lucbv (approved):
Good initialization values.

cwpearson:
Same thing for me, this does not happen in my local build.

brian-kelley:
Sweet, ubuntu-asan-ubsan-ci is clean now.

lucbv:
Down the line we may want to run these sanitizers on more configurations like complex numbers for instance, maybe also with clang in case that changes something?
I see only one of the HIP builds is still running, will keep an eye on this and merge when it completes.
Thanks for cleaning this up @brian-kelley !

PR #2421, opened 2024-11-07
gtest 1.15.2 from cwpearson/fused-gtest

cwpearson:


cwpearson:
Or we could jump all the way to 1.15.2 if you want. If you definitely don't want to, feel free to close this.

Issue #2371, opened 2024-10-10
`KOKKOS_CUSPARSE_SAFE_CALL` has `KOKKOS_` prefix


cwpearson:
https://github.com/kokkos/kokkos-kernels/blob/c7df61d536c74effb3e4466b011f67f24e85f314/sparse/src/KokkosSparse_Utils_cusparse.hpp#L67-L69

Let's be good ecosystem partners and prefix this with `KOKKOSSPARSE_` instead.

I personally don't think this should be part of our interface. Is this intentional? If so, let's use `KOKKOSSPARSE_CUSPARSE_SAFE_CALL`, otherwise let's deprecate it and use `KOKKOSSPARSE_IMPL_CUSPPARSE_SAFE_CALL`.

lucbv:
I think the last option makes sense: `KOKKOSSPARSE_IMPL_CUSPARSE_SAFE_CALL`, eventually we should review what we are doing for other libraries like rocSPARSE, rocBLAS, etc...

cwpearson:
Interestingly those other ones are either not `KOKKOS_` or they have `IMPL` in the name, in which case I feel happy to just rename them.

PR #2426, opened 2024-11-11
deprecate KOKKOS_CUSPARSE_SAFE_CALL -> KOKKOSPARSE_IMPL_CUSPARSE_SAFE_CALL

cwpearson:
Fix for https://github.com/kokkos/kokkos-kernels/issues/2371

### DIFF
--- a/sparse/src/KokkosSparse_Utils_cusparse.hpp
+++ b/sparse/src/KokkosSparse_Utils_cusparse.hpp
@@ -64,9 +64,23 @@ inline void cusparse_internal_safe_call(cusparseStatus_t cusparseStatus, const c
   }
 }
 
-// The macro below defines is the public interface for the safe cusparse calls.
-// The functions themselves are protected by impl namespace.
-#define KOKKOS_CUSPARSE_SAFE_CALL(call) KokkosSparse::Impl::cusparse_internal_safe_call(call, #call, __FILE__, __LINE__)
+#define KOKKOSSPARSE_IMPL_CUSPARSE_SAFE_CALL(call) \
+  KokkosSparse::Impl::cusparse_internal_safe_call(call, #call, __FILE__, __LINE__)
+
+// Deprecated public interface for the cuSparse safe calls
+#if defined(KOKKOS_COMPILER_MSVC)
+#define KOKKOS_CUSPARSE_SAFE_CALL(call)                                                                          \
+  (__pragma(message("warning: KOKKOS_CUSPARSE_SAFE_CALL is deprecated and will be removed in a future version")) \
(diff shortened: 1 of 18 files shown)


sparse/impl/KokkosSparse_sptrsv_solve_impl.hpp
```diff
@@ -220,13 +220,13 @@ struct SptrsvWrap {
...
-                  m_data[i] = rhs_.m_data[i];
-              }
+      ArrayType &operator=(const ArrayType &rhs_) {
+        if (this != &rhs_) {
+          for (size_type i = 0; i < MAX_VEC_SIZE; ++i) {
+            m_data[i] = rhs_.m_data[i];
           }
-          return *this;
+        }
+        return *this;
```
JBludau, line 229: what happened here?

JBludau (commented):
can you paste the command you used for the search and replace?

lucbv:
Something strange going on in the mi210 without TPLs, the tests are all reporting a segfault.
This might be a machine/workflow issue rather than an actual bug...

lucbv (approved):
Okay, this looks reasonable to me.
As mentioned by Jacob, have the command line/script in the PR description would be a good idea : )

sparse/impl/KokkosSparse_sptrsv_solve_impl.hpp
cwpearson: yes this is formatting related, I’m not sure why it wasn’t formatted before.

cwpearson:
I used Visual Studio Code’s global search/replace, except for the area where I implemented the deprecation macro.

sparse/src/KokkosSparse_Utils_cusparse.hpp
```diff
@@ -66,7 +66,8 @@ inline void cusparse_internal_safe_call(cusparseStatus_t cusparseStatus, const c
 
 // The macro below defines is the public interface for the safe cusparse calls.
 // The functions themselves are protected by impl namespace.
```
JBludau, line 68: I think the comment should be revised.
I could not find any documentation of the macro as part of the public interface. Do you know of anyone using it? Do you want to have the old one around but emit a deprecation warning instead of removing it?

cwpearson:
Looks like Trilinos has one use in Ifpack2:

https://github.com/trilinos/Trilinos/blob/8bded3a266b3aa18b5f9fd4f380b5136982b3c23/packages/ifpack2/src/Ifpack2_LocalSparseTriangularSolver_def.hpp#L678-L679

lucbv (approved):
Good point, the impl namespace is more appropriate here

cwpearson:
@lucbv I think we can merge this one?

PR #2433, opened 2024-11-19
gtest: add cast to remove sign-compare comparison warning

ndellingwood:
Resolve `error: comparison of integer expressions of different signedness: 'const long unsigned int' and 'const long int' [-Werror=sign-compare]`

The warning is triggering -Werror in Trilinos nightly integration builds

### DIFF
--- a/tpls/gtest/gtest/gtest.h
+++ b/tpls/gtest/gtest/gtest.h
@@ -11424,7 +11424,7 @@ AssertionResult CmpHelperEQ(const char* lhs_expression,
                             const char* rhs_expression,
                             const T1& lhs,
                             const T2& rhs) {
-  if (lhs == rhs) {
+  if (lhs == T1(rhs)) {
     return AssertionSuccess();
   }
 


ndellingwood:
The clang-format check wants to reformat the entire gtest.h file, should gtest.h be excluded from the check?

ndellingwood:
The AT2 h100 check is failing with 
```
 Unable to determine the device handle for GPU0000:B8:00.0: Unknown Error
Error: Process completed with exit code 255.
```
I'm not sure what that indicates, probably something I should forward to sys admins?

cwpearson:
@ndellingwood let's not format gtest. The way the format check is set up right now is that it tries to format all changed files, rather than specific files.

cwpearson (changes requested):
I'd rather put the cast in the test itself rather than modify gtest.

lucbv (changes requested):
Same as Carl, can you let us know which test triggers this compiler error?
Also I am somewhat surprised that this does not appear in the Kokkos Kernels CI directly? Let us know what combination of compiler flags and KK configuration triggers this failure so we can cover it better.

ndellingwood:
I'll close this, for reference this is the -Werror reported on cdash:
```
In file included from ../kokkos-kernels/test_common/Test_Serial.hpp:19,
                 from ../kokkos-kernels/sparse/unit_test/backends/Test_Serial_Sparse.cpp:19:
../kokkos-kernels/tpls/gtest/gtest/gtest.h: In instantiation of 'testing::AssertionResult testing::internal::CmpHelperEQ(const char*, const char*, const T1&, const T2&) [with T1 = long unsigned int; T2 = long int]':
../kokkos-kernels/tpls/gtest/gtest/gtest.h:11446:23:   required from 'static testing::AssertionResult testing::internal::EqHelper::Compare(const char*, const char*, const T1&, const T2&) [with T1 = long unsigned int; T2 = long int; typename std::enable_if<((! std::is_integral<_Tp>::value) || (! std::is_pointer<_Dp>::value))>::type* <anonymous> = 0]'
../kokkos-kernels/sparse/unit_test/Test_Sparse_TestUtils_RandCsMat.hpp:44:3:   required from 'void Test::doCsMat(size_t, size_t, ScalarType, ScalarType) [with ScalarType = float; LayoutType = Kokkos::LayoutLeft; ExeSpaceType = Kokkos::Serial; size_t = long unsigned int]'
../kokkos-kernels/sparse/unit_test/Test_Sparse_TestUtils_RandCsMat.hpp:64:51:   required from 'void Test::doAllCsMat(size_t, size_t) [with ExeSpaceType = Kokkos::Serial; size_t = long unsigned int]'
../kokkos-kernels/sparse/unit_test/Test_Sparse_TestUtils_RandCsMat.hpp:78:74:   required from here
../kokkos-kernels/tpls/gtest/gtest/gtest.h:11427:11: error: comparison of integer expressions of different signedness: 'const long unsigned int' and 'const long int' [-Werror=sign-compare]
   if (lhs == rhs) {
       ~~~~^~~~~~
[CTest: warning matched] cc1plus: some warnings being treated as errors
```

https://trilinos-cdash.sandia.gov/viewBuildError.php?buildid=1994256

cwpearson:
The offending line is 

https://github.com/kokkos/kokkos-kernels/blob/9121db79044bca19de722f13b7b8d615df006330/sparse/unit_test/Test_Sparse_TestUtils_RandCsMat.hpp#L44

Where lhs is a `long unsigned int` (`size_t` in this case) and rhs is a `long int`, which comes from `RandCsMatrix::ordinal_type`, which is defaulted to `int64_t` in this scenario.

This strikes me as a common case, I'm not sure why we don't see it in our testing

Issue #2424, opened 2024-11-09
CMake Config forces launch_compiler


rbberger:
Projects that use Kokkos and KokkosKernels can't disable the Kokkos `launch_compiler` via `COMPONENTS separable_compilation`. KokkosKernels always forces the use of `launch_compiler`.

```cmake
find_package(Kokkos COMPONENTS separable_compilation)
find_package(KokkosKernels)
```

still leads to the global `launch_compiler` rule to get activated due to `find_dependency(Kokkos)` in the KokkosKernels Config module.

JBludau:
`separable compilation` is not working as expected in kokkos core. We are currently working on it, for starters see https://github.com/kokkos/kokkos/pull/7525.
Will take a look at KokkosKernels as soon as it is fixed in core.

dalg24:
I am torn on this.  I feel like it is a "components" misuse.

lucbv:
@JBludau @dalg24 
I read the strings of PRs that came out of kokkos [#7581](https://github.com/kokkos/kokkos/pull/7581) and it seems that we decided to not pursue a resolution, at least not before Kitware/CMake offers better options to treat this problem.
Is that a fair summation of this effort?

JBludau:
> [@JBludau](https://github.com/JBludau) [@dalg24](https://github.com/dalg24) I read the strings of PRs that came out of kokkos [#7581](https://github.com/kokkos/kokkos/pull/7581) and it seems that we decided to not pursue a resolution, at least not before Kitware/CMake offers better options to treat this problem. Is that a fair summation of this effort?

The main problem was satisfying spack, cmake, and "it has always been this way" advocates at the same time.
The current proposed solution as of today (needs updating) is https://github.com/kokkos/kokkos/pull/8698, but the general battle has been pretty tiring.
For people that are willing to manually enable/disable the launcher we can provide a solution. For an automatic way that only applies the launcher to translation units that need it, we need changes in Kitware/CMake to propagate a launcher through a dependency

PR #2435, opened 2024-11-21
Only search for kokkos if it is not already found

JBludau:
This proposes a partial fix for #2424 Nevertheless, I am not yet sure if we are actually want to go that way

What is the problem:
Kokkos misuses the `component` feature of CMake to specify if the compiler Kokkos defines as `CXX_COMPILER` will be launched globally. The default is a global set on everything in the directory. But Kokkos allows to specify `separable_compilation` as a required component which disables this behavior. Nevertheless, because of the logic that CMake uses to aggregate `find_package/dependency` commands (see [here](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html)), we overwrite the user choice if `separable_compilation` was specified.

If we guard for an already found Kokkos we allow the user to use the separable compilation feature like:
```
find_package(Kokkos COMPONENTS separable_compilation)
find_package(KokkosKernels)

add_executable(example_with_kokkos example_with_kokkos.cpp)
kokkos_compilation(TARGET example_with_kokkos) #this sets the compiler for the target to the appropriate one
target_link_libraries(example_with_kokkos PRIVATE Kokkos::kokkos Kokkos::kokkoskernels)

add_executable(example_without_kokkos foobar.cpp)
...

```

Alternative solution:
Kokkos sets the compiler only for the current scope ... so putting your kokkos files in a subdirectory with its own targets and link Kokkos in `PRIVATE` mode will allow to compile all other directories how the user prefers. Nevertheless this will not work if Kokkos is exposed in the other directories.

What do we need to discuss:

- [ ] does this work with kokkos/kokkos-kernels as part of trilinos
- [ ] This is a misuse of `components` in CMake ... Kokkos core should think about if this should stay this way.

### DIFF
--- a/cmake/KokkosKernelsConfig.cmake.in
+++ b/cmake/KokkosKernelsConfig.cmake.in
@@ -7,7 +7,9 @@ include(CMakeFindDependencyMacro)
 
 @KOKKOSKERNELS_TPL_EXPORTS@
 
-find_dependency(Kokkos HINTS @Kokkos_DIR@)
+IF(NOT Kokkos_FOUND)
+  find_dependency(Kokkos HINTS @Kokkos_DIR@)
+ENDIF()
 
 INCLUDE("${KokkosKernels_CMAKE_DIR}/KokkosKernelsTargets.cmake")
 


rbberger:
Can this workaround at least be merged for a future release while you figure out a long term solution?

JBludau:
@lucbv @cwpearson what do you think? Would you consider this workaround for now? I found a way to get rid of the problems with the components and am currently pushing it into core but it will take some time to convince some people ...

lucbv:
Done via PR #2511

PR #2439, opened 2024-11-25
ci: Disable PR tests which ran on `sogpu01` at Sandia

cwpearson:
This machine has been repurposed.

### DIFF
--- a/.github/workflows/at2.yml
+++ b/.github/workflows/at2.yml
@@ -38,7 +38,8 @@ jobs:
     uses: ./.github/workflows/spr.yml
     with:
       kokkos_version: 4.5.00
-  volta70:
-    uses: ./.github/workflows/volta70.yml
-    with:
-      kokkos_version: 4.5.00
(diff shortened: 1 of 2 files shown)


JBludau:
why not delete the lines instead of commenting out?

jgfouca (approved):

jgfouca:
I think commenting out makes sense since we will likely be re-enabling v100 testing on some other machine soon.

lucbv (changes requested):
Actually looking at PR #2441 the Clang CI build seems to still be running correctly. We should keep it if possible.

cwpearson:
It was an oversight that that test continued to work. It was running on sogpu01 and I have disabled it since that machine will be used for something else.

lucbv:
Okay, if you can do a quick rebase for the update to kokkos 4.5.00 as version to test against then I can approve and we can merge this.

lucbv (approved):

PR #2441, opened 2024-11-25
Update version, changelog, checks for 4.5

ndellingwood:
Depends on merge of #2440


CMakeLists.txt
```diff
@@ -136,13 +136,13 @@ ELSE()
         SET(CMAKE_HIP_ARCHITECTURES ${Kokkos_HIP_ARCHITECTURES})
       ENDIF()
     ENDIF()
-    IF(${Kokkos_VERSION} VERSION_GREATER_EQUAL "4.3.01")
+    IF(${Kokkos_VERSION} VERSION_GREATER_EQUAL "4.4.01")
```
lucbv, line 139: This should be 4.4.00 to allow two the support of two version of Kokkos.

CMakeLists.txt
```diff
@@ -136,13 +136,13 @@ ELSE()
...
       MESSAGE(STATUS "Found Kokkos version ${Kokkos_VERSION} at ${Kokkos_DIR}")
-      IF((${Kokkos_VERSION} VERSION_GREATER "4.4.99"))
+      IF((${Kokkos_VERSION} VERSION_GREATER "4.5.99"))
         MESSAGE(WARNING "Configuring with Kokkos ${Kokkos_VERSION} which is newer than the expected develop branch - version check may need update")
       ENDIF()
     ELSE()
-      MESSAGE(FATAL_ERROR "Kokkos Kernels ${KokkosKernels_VERSION} requires Kokkos 4.3.01 or greater (found ${Kokkos_VERSION})")
+      MESSAGE(FATAL_ERROR "Kokkos Kernels ${KokkosKernels_VERSION} requires Kokkos 4.4.01 or greater (found ${Kokkos_VERSION})")
```
lucbv, line 145: Same as above, the new minimum should be 4.4.00

lucbv (changes requested):
Ugh, I greped for "4.4.01" in ou workflow files thinking that would get us up to date but clearly that would not catch these "4.3.01" stragglers...

CMakeLists.txt
ndellingwood: We used 4.4.01 for the release and merge to master though, shouldn't we stay consistent?

CMakeLists.txt
lucbv: My bad, you are correct that would be 4.4.01 and 4.5.0

CMakeLists.txt
lucbv: Yes that's correct

lucbv (approved):
Looks good to me

PR #2451, opened 2024-12-03
__KOKKOSBATCHED_ENABLE_AVX__ -> KOKKOSBATCHED_IMPL_ENABLE_AVX

cwpearson:
Ordinarily I'd deprecate this, but as far as I know, there's no good way to deprecate a macro when its only use is checking if the macro is defined.

https://github.com/kokkos/kokkos-kernels/issues/2364


cwpearson:
@lucbv how do you feel about this one? Is there a better way?

masterleinad:
Might be conflicting with https://github.com/kokkos/kokkos-kernels/pull/2510.

masterleinad (approved):

lucbv (approved):

PR #2460, opened 2024-12-13
Fix the package version

### LINKED ISSUE #2459: KokkosKernels PACKAGE_VERSION is not set
The `PACKAGE_VERSION` does not seem to always be correctly set. The installation of Kokkos Kernels fails to get the correct version, I get the following error 
```
CMake Error at CMakeLists.txt:5 (find_package):
  Could not find a configuration file for package "KokkosKernels" that is
  compatible with requested version "4.5".

  The following configuration files were considered but not accepted:

    /opt/lib/cmake/KokkosKernels/KokkosKernelsConfig.cmake, version: ..
```

This can be reproduced on the develop branch with the following `CMakeLists.txt`
```cmake
cmake_minimum_required(VERSION 3.22)

project(test_kokkoskernels_version LANGUAGES CXX)

find_package(KokkosKernels 4.5 REQUIRED)
```

It seems to me it is related to the call to the `project` command without a version that erases the version variables previously defined https://github.com/kokkos/kokkos-kernels/blob/d47cadbb67fb1ad6f6ee3aa16747c8ea79005cd1/CMakeLists.txt#L23-L37

We can either pass the version in `project` or move this block of code before the definition of the version variables.

tpadioleau:
Fixes #2459


lucbv:
@tpadioleau why not set the version parameter in the call to `PROJECT()` on line 35?
Would that fix the issue you are observing?

tpadioleau:
> @tpadioleau why not set the version parameter in the call to `PROJECT()` on line 35? Would that fix the issue you are observing?

Yes it is a solution, I have updated accordingly

tpadioleau:
Could it be part of the patch release 4.5.01 ?

lucbv:
Thanks, I tried locally on my laptop as well and it looks good so we should go for that. We can definitely ship it in 4.5.01 then. My sense is we can merge it today : )

lucbv (approved):
Thanks for diagnosing and helping us fix this

lucbv:
Ah you would need the DCO thing to make things perfect but it all tests well I could give you a pass on that.

lucbv (approved):
lgtm

PR #2469, opened 2024-12-19
Improve batched serial gemm

yasahi-hpc:
This PR aims at improving the implementation and testing of serial Gemm.

 - [x] Moving implementation details into `Impl` namespace.
 - [x] Moving `OpID` and `OpConj` into `blas/impl/KokkosBlas_util.hpp`
 - [x] Add all the specializations of the combinations of `Non-Trans/Trans/ConjTrans`
 - [x] Add a check function of input to force A, B, C are rank 0, 1, 2 `Kokkos::View`
 - [x] Deprecated warnings for the older implementation details under `KokkosBatched` namespace

### DIFF
--- a/batched/dense/unit_test/Test_Batched_SerialGemm.hpp
+++ b/batched/dense/unit_test/Test_Batched_SerialGemm.hpp
@@ -34,95 +34,98 @@ namespace Gemm {
 
 template <typename TA, typename TB>
 struct ParamTag {
-  typedef TA transA;
-  typedef TB transB;
+  using transA = TA;
+  using transB = TB;
 };
 
 template <typename DeviceType, typename ViewType, typename ScalarType, typename ParamTagType, typename AlgoTagType>
 struct Functor_TestBatchedSerialGemm {
   using execution_space = typename DeviceType::execution_space;
-  ViewType _a, _b, _c;
-
-  ScalarType _alpha, _beta;
+  ViewType m_a, m_b, m_c;
+  ScalarType m_alpha, m_beta;
 
   KOKKOS_INLINE_FUNCTION
   Functor_TestBatchedSerialGemm(const ScalarType alpha, const ViewType &a, const ViewType &b, const ScalarType beta,
                                 const ViewType &c)
-      : _a(a), _b(b), _c(c), _alpha(alpha), _beta(beta) {}
+      : m_a(a), m_b(b), m_c(c), m_alpha(alpha), m_beta(beta) {}
 
   KOKKOS_INLINE_FUNCTION
   void operator()(const ParamTagType &, const int k) const {
-    auto aa = Kokkos::subview(_a, k, Kokkos::ALL(), Kokkos::ALL());
-    auto bb = Kokkos::subview(_b, k, Kokkos::ALL(), Kokkos::ALL());
-    auto cc = Kokkos::subview(_c, k, Kokkos::ALL(), Kokkos::ALL());
+    auto aa = Kokkos::subview(m_a, k, Kokkos::ALL(), Kokkos::ALL());
+    auto bb = Kokkos::subview(m_b, k, Kokkos::ALL(), Kokkos::ALL());
+    auto cc = Kokkos::subview(m_c, k, Kokkos::ALL(), Kokkos::ALL());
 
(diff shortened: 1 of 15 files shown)


common/src/KokkosKernels_BlockUtils.hpp
```diff
@@ -52,13 +52,13 @@ KOKKOS_INLINE_FUNCTION void kk_block_add(const size_type block_dim, value_type *
...
 template <typename size_type, typename value_type,
-          typename DGEMM = KokkosBatched::SerialGemmInternal<KokkosBatched::Algo::Gemm::Unblocked>>
+          typename DGEMM = KokkosBatched::Impl::SerialGemmInternal<KokkosBatched::Algo::Gemm::Unblocked>>
 KOKKOS_INLINE_FUNCTION void kk_block_dgemm(const size_type block_dim, value_type *dst, const value_type *valA,
                                            const value_type *valB, const bool clear = false) {
   const auto ZERO = static_cast<value_type>(0);
   const auto ONE  = static_cast<value_type>(1);
-  DGEMM::invoke(block_dim, block_dim, block_dim, ONE, valA, block_dim, 1, valB, block_dim, 1, clear ? ZERO : ONE, dst,
-                block_dim, 1);
+  DGEMM::invoke(KokkosBlas::Impl::OpID(), KokkosBlas::Impl::OpID(), block_dim, block_dim, block_dim, ONE, valA,
```
cwpearson, line 60: I'm confused about how calling `OpID()` here with no arguments works. It looks like it must take a value.

cwpearson (changes requested):

common/src/KokkosKernels_BlockUtils.hpp
yasahi-hpc: Thank you for the comments. 
I am doing exactly the same thing as in gemv. Basically, I am just passing a temporary object of `OpID()` class.

https://github.com/kokkos/kokkos-kernels/blob/85bbf1ff4aba86881c6bd65e8b00ec2fd21d6e9e/blas/impl/KokkosBlas2_serial_gemv_impl.hpp#L91-L107

yasahi-hpc:
@cwpearson I have removed the global namespace following your suggestions for #2331 
I have also confirmed that Trilinos does not rely on the implementation details directly. 
I am worried about the potential usage of the APIs with `DynRankViews` though

common/src/KokkosKernels_BlockUtils.hpp
lucbv: To be clear, in that spot Yuuichi is calling the constructor not `operator()` so that should work fine.

batched/dense/unit_test/Test_Batched_SerialGemm.hpp
```diff
@@ -14,133 +14,133 @@
...
+
+  // check C = C_ref
   using mag_type = typename ats::mag_type;
   mag_type sum(1), diff(0);
 
   mag_type eps = ats::epsilon();
-
-  eps *= std::is_same<value_type, Kokkos::Experimental::half_t>::value ||
-                 std::is_same<value_type, Kokkos::Experimental::bhalf_t>::value
+  eps *= std::is_same_v<ValueType, Kokkos::Experimental::half_t> ||
```
lucbv, line 134: This definition of epsilon worries me a bit, we should revisit to actually compute the expected accuracy based on the compute parameters chosen for the test...

lucbv (approved):
The changes look fine to me, let's run this through the auto-tester.

cwpearson:
> @cwpearson I have removed the global namespace following your suggestions for #2331 I have also confirmed that Trilinos does not rely on the implementation details directly. I am worried about the potential usage of the APIs with `DynRankViews` though

There's `Kokkos::is_dyn_rank_view_v` if you want to modify the static asserts, or are you worried for another reason?

cwpearson (approved):

yasahi-hpc:
> > @cwpearson I have removed the global namespace following your suggestions for #2331 I have also confirmed that Trilinos does not rely on the implementation details directly. I am worried about the potential usage of the APIs with `DynRankViews` though
> 
> There's `Kokkos::is_dyn_rank_view_v` if you want to modify the static asserts, or are you worried for another reason?

I personally prefer not to allow `DynRankViews` because we cannot check ranks of `DynRankViews` at compile time.

PR #2473, opened 2025-01-06
BLAS - ROT: calling appropriate cuBLAS function

lucbv:
After changing the rot interface from using all real to a mix of real and complex coefficients, the cuBLAS call needs to be made to {Z,C}rot instead of {Zd,Cs}rot.


brian-kelley (approved):
I tried this out locally and it fixed the rot example and blas unit tests.

ndellingwood:
We need this for the nightly cu* TPL and Trilinos builds as well

vqd8a (approved):
Thanks, @lucbv

PR #2478, opened 2025-01-13
Refactor serial tbsv implementation details and tests

yasahi-hpc:
This PR aims at improving the implementation and testing of serial tbsv.

 - [x] Moving implementation details into `Impl` namespace.
 - [x] Use `OpID` and `OpConj` for `conj` operations instead of `do_conj` variable
 - [x] Use `Kokkos::is_view_v<AViewType>` instead of `Kokkos::is_view<AViewType>::value`
 - [x] Remove `using namespace KokkosBatched` from the test code
 - [x] Rename variables starting from underscores in the test code  

Added based on the discussions in #2497
- [X] Add docstring and assertion for `Arg` template parameters

### DIFF
--- a/batched/dense/impl/KokkosBatched_Pbtrs_Serial_Internal.hpp
+++ b/batched/dense/impl/KokkosBatched_Pbtrs_Serial_Internal.hpp
@@ -17,6 +17,7 @@
 #ifndef KOKKOSBATCHED_PBTRS_SERIAL_INTERNAL_HPP_
 #define KOKKOSBATCHED_PBTRS_SERIAL_INTERNAL_HPP_
 
+#include "KokkosBlas_util.hpp"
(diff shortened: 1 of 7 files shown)


lucbv (approved):

lucbv:
I don't think the gesvd error is related to this so we could merge the PR, fine with you @yasahi-hpc ?

yasahi-hpc:
> I don't think the gesvd error is related to this so we could merge the PR, fine with you @yasahi-hpc ?

I agree gesvd error should be unrelated. Based on the discussions in #2497, I added docstrings and assertion about `Arg` parameters. If it is good for you, I think this PR is ready to be merged.

If we find errors in gesvd, I will open an issue and fix that.

lucbv (approved):

PR #2483, opened 2025-01-14
implement batched serial getrs

yasahi-hpc:
This PR implements [getrs](https://www.netlib.org/lapack/explore-html-3.6.1/dd/d9a/group__double_g_ecomputational_ga58e332cb1b8ab770270843221a48296d.html) function.

Following files are added:
1. `KokkosBatched_Getrs_Serial_Impl.hpp`: Internal interfaces
2. `KokkosBatched_Getrs_Serial_Internal.hpp`: Implementation details
3. `KokkosBatched_Getrs.hpp`: APIs
4. `Test_Batched_SerialGetrs.hpp`: Unit tests for that

## Detailed description
It solves a general N-by-N matrix `A` using the LU factorization computed by getrf.
- `A`: `(batch_count, n, n)`  
  The N-by-N factorized matrix by getrf where `A = P * L * U`; the unit diagonal elements of `L` are not stored.
- `IPIV`: `(batch_count, n)`
  The pivot indices from getrf. for `0 <= i < n`, row `i` of the matrix was interchanged with row IPIV(i).

```C++
Kokkos::parallel_for('getrs', 
    [=](const int k) {
        auto ipiv = Kokkos::subview(m_ipiv, k, Kokkos::ALL());
        auto bb   = Kokkos::subview(m_b, k, Kokkos::ALL());

        KokkosBatched::SerialGetrs<Trans, AlgoTagType>::invoke(aa, ipiv, bb);
    });
```

## Tests
1.  Make a random matrix from random `A` and factorize it into `LU` with `ipiv` by `getrf`.
Then, solve `A * x = b` with `getrs` to get `x`, while keeping the original `b` in `x_ref`. Finally, confirm that `A * x` is equal to `b (=x_ref)` using `gem`.
1.  Simple and small analytical test, i.e. choose `A` as follows to confirm `LU` == `A`.
```bash
A: [[1, 1],
    [1, -1]]
b: [2, 0]
x: [1, 1]
```

### DIFF
--- a/batched/dense/impl/KokkosBatched_Getrs_Serial_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Getrs_Serial_Impl.hpp
@@ -1,0 +1,76 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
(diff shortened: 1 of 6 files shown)


lucbv (approved):
Thanks for the update

PR #2491, opened 2025-02-04
implement batched serial ger

yasahi-hpc:
This PR implements [ger](https://www.netlib.org/lapack/explore-html-3.6.1/d7/d15/group__double__blas__level2_ga458222e01b4d348e9b52b9343d52f828.html) function.

Following files are added:
1. `KokkosBatched_Ger_Serial_Impl.hpp`: Internal interfaces
2. `KokkosBatched_Ger_Serial_Internal.hpp`: Implementation details
3. `KokkosBatched_Ger.hpp`: APIs
4. `Test_Batched_SerialGer.hpp`: Unit tests for that

## Detailed description
It performs the rank 1 operation `A:= alpha*x*y**T + A` or `A:= alpha*x*y**H + A`
- `x`: `(batch_count, m)`  
On entry, it contains the m elements of x.
- `y`: `(batch_count, n)`
On entry, it contains the n elements of y.
On entry, the leading m by n part of the array A must contain the matrix of coefficients. 
On exit, A is overwritten by the updated matrix.

```C++
Kokkos::parallel_for('ger', 
    [=](const int k) {
        auto xx = Kokkos::subview(m_x k, Kokkos::ALL());
        auto yy = Kokkos::subview(m_y, k, Kokkos::ALL());

        KokkosBatched::SerialGer<TransType>::invoke(xx, yy, aa);
    });
```

## Tests
1.  Make a random `x`, `y` and `A`, while copying `A` into `A_ref`. The reference `A_ref` is computed by `A_ref:= alpha*x*y**T + A_ref` or `A_ref:= alpha*x*y**H + A_ref` at host.  Finally, we confirm `A` computed by serial ger and `A_ref` are the same. A == 0 case is tested as well.
1.  Simple and small analytical test, i.e. choose `x`, `y`, and `A` as follows to confirm `A` is updated as expected. 3x4 and 5x4 cases are tested as well.
```bash
A = [[1, -3, -2,  0],
     [-1, 1, -3, -2],
     [2, -1,  1, -3],
     [0,  2, -1,  1]]
x = [1, 2, 3, 4]
y:  [0, 1, 2, 3]
Ref = [[ 1.,  -1.5,  1.,   4.5],
       [-1.,   4.,   3.,   7. ],
       [ 2.,   3.5, 10.,  10.5],
       [ 0.,   8.,  11.,  19. ]]
```

### DIFF
--- a/batched/dense/impl/KokkosBatched_Ger_Serial_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Ger_Serial_Impl.hpp
@@ -1,0 +1,118 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOSBATCHED_GER_SERIAL_IMPL_HPP_
+#define KOKKOSBATCHED_GER_SERIAL_IMPL_HPP_
+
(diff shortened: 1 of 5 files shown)


lucbv (approved):
Looks good

PR #2494, opened 2025-02-05
Sparse - Gauss Seidel: moving GS out of experimental namespace

lucbv:


sparse/unit_test/Test_Sparse_block_gauss_seidel.hpp
```diff
@@ -97,29 +97,29 @@ int run_block_gauss_seidel_1(
   const int apply_count   = 100;
 
   if (!skip_symbolic) {
-    KSExp::block_gauss_seidel_symbolic(&kh, num_rows_1, num_cols_1, block_size, input_mat.graph.row_map,
-                                       input_mat.graph.entries, is_symmetric_graph);
+    KokkosSparse::block_gauss_seidel_symbolic(&kh, num_rows_1, num_cols_1, block_size, input_mat.graph.row_map,
```
brian-kelley, line 100: Could now get rid of KSExp namespace alias since it's not used anywhere else.

brian-kelley (approved):
Looks good. There is one line ``namespace KSExp = ...`` in the block GS test that can be cleaned up now

sparse/unit_test/Test_Sparse_block_gauss_seidel.hpp
lucbv: Oh right that makes sense, will go clean that up as well

brian-kelley (approved):

PR #2495, opened 2025-02-06
Iluk deprecate experimental

lucbv:


vqd8a (approved):

vqd8a:
> Correct, I think it is still a bit early for that. Let me know if you want to take it out of experimental?

@lucbv I am okay with letting it in experimental for a little while. Just wanted to make sure.

PR #2499, opened 2025-02-11
implement batched serial lacgv

yasahi-hpc:
This PR implements [lacgv](https://www.netlib.org/lapack/explore-html/d9/d50/group__lacgv_gae42087fcabd33130fcbac2aff031de8b.html) function.

Following files are added:
1. `KokkosBatched_Lacgv_Serial_Impl.hpp`: Internal interfaces
2. `KokkosBatched_Lacgv_Serial_Internal.hpp`: Implementation details
3. `KokkosBatched_Lacgv.hpp`: APIs
4. `Test_Batched_SerialLacgv.hpp`: Unit tests for that

## Detailed description
Conjugates the length n complex vector.
- `x`: `(batch_count, n)`  
On entry, it contains the n elements of `x`.
On exit, `x` is overwritten with `conj(x)`.

No operation is performed if `x` is real vector.

```C++
Kokkos::parallel_for('lacgv', 
    [=](const int k) {
        auto xx = Kokkos::subview(m_x, k, Kokkos::ALL());
        KokkosBatched::SerialLacgv::invoke(xx);
    });
```

## Tests
1.  Make a random `x`, while copying `x` into `x_ref`. The reference `x_ref` is computed by `x_ref:= conj(x)` at host.  Finally, we confirm `x` computed by serial lacgv and `x_ref` are the same.
1.  Simple and small analytical test, i.e. choose `x` as follows to confirm `x` is updated as expected. Both real and complex vectors are tested.
```bash
x = [1 + 1j, -3+2j, -2-2j,  0+1j]
Ref = conj(x) = [1 - 1j, -3-2j, -2+2j,  0-1j]
```

### DIFF
--- a/batched/dense/impl/KokkosBatched_Lacgv_Serial_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Lacgv_Serial_Impl.hpp
@@ -1,0 +1,33 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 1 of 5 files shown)


lucbv (approved):
Looks good to me

PR #2500, opened 2025-02-12
Adding ctest config for cdash submissions

lucbv:
As explained in the title, this will allow us to submit to a public cdash [here](https://my.cdash.org/index.php?project=Kokkos+Kernels) for jobs on machines that permit it.

### DIFF
--- a/CTestConfig.cmake
+++ b/CTestConfig.cmake
@@ -1,0 +1,4 @@
+set(CTEST_PROJECT_NAME Kokkos Kernels)
+set(CTEST_NIGHTLY_START_TIME 01:00:00 UTC)
+set(CTEST_SUBMIT_URL https://my.cdash.org/submit.php?project=Kokkos+Kernels)
+set(CTEST_DROP_SITE_CDASH TRUE)


ndellingwood:
Adding a note here, jobs run from SNL resources/hardware currently cannot post results to my.cdash.org

CTestConfig.cmake
```diff
@@ -0,0 +1,4 @@
+set(CTEST_PROJECT_NAME Kokkos Kernels)
+set(CTEST_NIGHTLY_START_TIME 01:00:00 UTC)
+set(CTEST_SUBMIT_URL https://my.cdash.org/submit.php?project=Kokkos+Kernels)
```
ndellingwood, line 3: Adding a note here, jobs run from SNL resources/hardware currently cannot post results to my.cdash.org - results can only be posted on approved internal server resources

ndellingwood (approved):
Thanks @lucbv

jgfouca (approved):

CTestConfig.cmake
lucbv: That is correct, the jobs planned will be coming from ALCF and OLCF only

PR #2502, opened 2025-02-18
Worflow: refactoring SPR to standard cmake and suppress warnings

lucbv:
The -Wpass-fail warnings are poluting the output of the compiler this does not really tell us anything either so let us suppress these.

### DIFF
--- a/.github/workflows/spr.yml
+++ b/.github/workflows/spr.yml
@@ -30,27 +30,65 @@ jobs:
           ref: ${{ inputs.kokkos_version }}
           path: kokkos
 
-      - name: configure
+      - name: configure_kokkos
         run: |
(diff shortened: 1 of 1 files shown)


lucbv:
Do we want to add `-DCMAKE_EXE_LINKER_FLAGS="-Wno-unused-command-line-argument"` to suppress the linker warnings? I am guessing the warnings are due to some path spack is adding by default?

jgfouca (approved):

ndellingwood:
@lucbv I think it's worth trying the suppression with the linker flags to see if it quiets down the noise in the output

lucbv:
This latest update suppresses the linker warning we use to have, it looks like it's mostly building warning free now

PR #2510, opened 2025-02-25
SYCL - SIMD: preventing AVX512 code to be compiled when targeting GPU

lucbv:
There is a check missing to verify that we are not trying to compile AVX512 intrinsic in GPU code which is obviously not a good idea.

### DIFF
--- a/batched/dense/src/KokkosBatched_Vector_SIMD.hpp
+++ b/batched/dense/src/KokkosBatched_Vector_SIMD.hpp
@@ -22,7 +22,7 @@
 #include <KokkosBatched_Vector.hpp>
 #include "KokkosKernels_Macros.hpp"
 
-#if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)
+#if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__) || defined(SYCL_LANGUAGE_VERSION)
 #undef __KOKKOSBATCHED_ENABLE_AVX__
 #else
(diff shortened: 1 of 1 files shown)


lucbv:
@masterleinad please have a look, this should fix the issue we are having with the AVX512 intrinsics.

batched/dense/src/KokkosBatched_Vector_SIMD.hpp
```diff
@@ -22,7 +22,7 @@
 #include <KokkosBatched_Vector.hpp>
 #include "KokkosKernels_Macros.hpp"
 
-#if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)
+#if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__) || defined(SYCL_LANGUAGE_VERSION)
```
masterleinad, line 25: The corresponding macro would rather be
```suggestion
#if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__) || defined(__SYCL_DEVICE_ONLY__)
```

masterleinad:
Might be conflicting with https://github.com/kokkos/kokkos-kernels/pull/2451.

batched/dense/src/KokkosBatched_Vector_SIMD.hpp
lucbv: Done

brian-kelley (approved):

PR #2523, opened 2025-03-05
docs: pinning github actions by SHA

lucbv:


lucbv:
The github-DOCS action has passed so I think this should be enough for this to be merged : )

jgfouca (approved):

PR #2539, opened 2025-03-14
Implement batched serial gbtrs

yasahi-hpc:
This PR implements [gbtrs](https://www.netlib.org/lapack/explore-html/d1/dd7/group__gbtrs_ga21ed607226c9690cadc991f15435fa74.html) function.

Following files are added:
1. `KokkosBatched_Gbtrs_Serial_Impl.hpp`: Internal interfaces
2. `KokkosBatched_Gbtrs_Serial_Internal.hpp`: Implementation details
3. `KokkosBatched_Gbtrs.hpp`: APIs
4. `Test_Batched_SerialGbtrs.hpp`: Unit tests for that

## Detailed description
It solves a general band matrix `A` using the LU factorization computed by gbtrf.
- `A`: `(batch_count, lab, n)`  
The factorized band matrix by gbtrf. The factors L and U from the factorization where U is stored as an upper triangular band matrix with KL+KU superdiagonals in rows 0 to KL+KU, and the multipliers used during the factorization are stored in rows KL+KU+1 to 2*KL+KU.
- `IPIV`: `(batch_count, n)`
  The pivot indices from gbtrf. for `0 <= i < n`, row `i` of the matrix was interchanged with row IPIV(i).
- kl: The number of subdiagonals within the band of A. kl >= 0
- ku: The number of superdiagonals within the band of A. ku >= 0

```C++
Kokkos::parallel_for('gbtrs', 
    [=](const int k) {
        auto ipiv = Kokkos::subview(m_ipiv, k, Kokkos::ALL());
        auto bb   = Kokkos::subview(m_b, k, Kokkos::ALL());

        KokkosBatched::SerialGbtrs<Trans, AlgoTagType>::invoke(aa, ipiv, bb);
    });
```

## Tests
1.  Make a random band matrix from random `A`. Represent `A` in band storage `AB` and factorize it with `gbtrf`, followed by solving `AB * x = b` by `gbtrs` to get `x`. Note `b` is filled with ones. Finally, confirm that `A * x` is equal to b (ones) using `gemv`.
1.  Simple and small analytical test, i.e. choose `A` as follows to confirm `A * x = b`, where
```bash
A = [[1. -3. -2. 0.],
     [-1. 1 -3 -2],
     [2. -1. 1. -3],
     [0. 2. -1. 1.]]
b: [1, 1, 1, 1]
x: [67/81, 22/81, -40/81, -1/27] (A * x = b) or [-1/27, -40/81, 22/81, 67/81] (A**T * x = b)
```

We confirm this with the factorized matrix `LUB` and pivot given by
```bash
LUB: [[0,       0,    0,    0],
      [0,       0,    0,   -3],
      [0,       0,    1,  1.5],
      [0,      -1, -2.5, -3.2],
      [2,    -2.5,   -3,  5.4],
      [-0.5, -0.2,    1,    0],
      [0.5,  -0.8,    0,    0]]
piv = [2 2 2 3]
```

### DIFF
--- a/batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
@@ -30,12 +30,14 @@ KOKKOS_INLINE_FUNCTION static int checkGbtrsInput([[maybe_unused]] const AViewTy
   static_assert(Kokkos::is_view_v<AViewType>, "KokkosBatched::gbtrs: AViewType is not a Kokkos::View.");
   static_assert(Kokkos::is_view_v<PivViewType>, "KokkosBatched::gbtrs: PivViewType is not a Kokkos::View.");
   static_assert(Kokkos::is_view_v<BViewType>, "KokkosBatched::gbtrs: BViewType is not a Kokkos::View.");
-  static_assert(AViewType::rank == 2, "KokkosBatched::gbtrs: AViewType must have rank 2.");
-  static_assert(PivViewType::rank == 1, "KokkosBatched::gbtrs: PivViewType must have rank 1.");
-  static_assert(BViewType::rank == 1, "KokkosBatched::gbtrs: BViewType must have rank 1.");
+  static_assert(AViewType::rank() == 2, "KokkosBatched::gbtrs: AViewType must have rank 2.");
+  static_assert(PivViewType::rank() == 1, "KokkosBatched::gbtrs: PivViewType must have rank 1.");
+  static_assert(BViewType::rank() == 1, "KokkosBatched::gbtrs: BViewType must have rank 1.");
   using PivValueType = typename PivViewType::non_const_value_type;
   static_assert(std::is_integral_v<PivValueType>,
                 "KokkosBatched::gbtrs: value type of PivViewType must be an integral type.");
+  static_assert(std::is_same_v<typename BViewType::value_type, typename BViewType::non_const_value_type>,
+                "KokkosBatched::gbtrs: BViewType must have non-const value type.");
 #if (KOKKOSKERNELS_DEBUG_LEVEL > 0)
   if (kl < 0) {
     Kokkos::printf(
@@ -58,7 +60,7 @@ KOKKOS_INLINE_FUNCTION static int checkGbtrsInput([[maybe_unused]] const AViewTy
   const int lda = A.extent(0), n = A.extent(1);
   if (lda < (2 * kl + ku + 1)) {
     Kokkos::printf(
-        "KokkosBatched::gbtrs: leading dimension of A must be smaller than 2 * "
+        "KokkosBatched::gbtrs: leading dimension of A must not be smaller than 2 * "
         "kl + ku + 1: "
         "lda = %d, kl = %d, ku = %d\n",
         lda, kl, ku);
@@ -68,7 +70,7 @@ KOKKOS_INLINE_FUNCTION static int checkGbtrsInput([[maybe_unused]] const AViewTy
   const int ldb = b.extent(0);
   if (ldb < Kokkos::max(1, n)) {
     Kokkos::printf(
-        "KokkosBatched::gbtrs: leading dimension of b must be smaller than "
+        "KokkosBatched::gbtrs: leading dimension of b must not be smaller than "
         "max(1, n): "
         "ldb = %d, n = %d\n",
         ldb, n);
--- a/batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
@@ -76,7 +76,7 @@ KOKKOS_INLINE_FUNCTION static int checkGbtrsInput([[maybe_unused]] const AViewTy
     return 1;
   }
 
-  const int npiv = ipiv.extent(0);
+  const int npiv = ipiv.extent_int(0);
   if (npiv != n) {
     Kokkos::printf(
         "KokkosBatched::gbtrs: the dimension of the ipiv array must "
--- a/batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
@@ -33,9 +33,9 @@ KOKKOS_INLINE_FUNCTION static int checkGbtrsInput([[maybe_unused]] const AViewTy
   static_assert(AViewType::rank() == 2, "KokkosBatched::gbtrs: AViewType must have rank 2.");
   static_assert(PivViewType::rank() == 1, "KokkosBatched::gbtrs: PivViewType must have rank 1.");
   static_assert(BViewType::rank() == 1, "KokkosBatched::gbtrs: BViewType must have rank 1.");
-  using PivValueType = typename PivViewType::non_const_value_type;
-  static_assert(std::is_integral_v<PivValueType>,
-                "KokkosBatched::gbtrs: value type of PivViewType must be an integral type.");
+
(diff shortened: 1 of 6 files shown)


batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
```diff
@@ -0,0 +1,108 @@
...
+namespace Impl {
+template <typename AViewType, typename PivViewType, typename BViewType>
+KOKKOS_INLINE_FUNCTION static int checkGbtrsInput([[maybe_unused]] const AViewType &A,
+                                                  [[maybe_unused]] const PivViewType &ipiv,
+                                                  [[maybe_unused]] const BViewType &b, [[maybe_unused]] const int kl,
+                                                  [[maybe_unused]] const int ku) {
+  static_assert(Kokkos::is_view_v<AViewType>, "KokkosBatched::gbtrs: AViewType is not a Kokkos::View.");
+  static_assert(Kokkos::is_view_v<PivViewType>, "KokkosBatched::gbtrs: PivViewType is not a Kokkos::View.");
+  static_assert(Kokkos::is_view_v<BViewType>, "KokkosBatched::gbtrs: BViewType is not a Kokkos::View.");
+  static_assert(AViewType::rank == 2, "KokkosBatched::gbtrs: AViewType must have rank 2.");
```
lucbv, line 33: In general, switch to `View::rank()` instead of `View::rank` as explained in the documentation of [Kokkos::View](https://kokkos.org/kokkos-core-wiki/API/core/view/view.html#enums).

batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
```diff
@@ -0,0 +1,108 @@
...
+  static_assert(PivViewType::rank == 1, "KokkosBatched::gbtrs: PivViewType must have rank 1.");
+  static_assert(BViewType::rank == 1, "KokkosBatched::gbtrs: BViewType must have rank 1.");
+  using PivValueType = typename PivViewType::non_const_value_type;
+  static_assert(std::is_integral_v<PivValueType>,
+                "KokkosBatched::gbtrs: value type of PivViewType must be an integral type.");
```
lucbv, line 38: To be perfectly exhaustive you probably also want to check:
```c++
static_assert(std::is_same_v<typename BViewType::values_type, typename BViewType::non_const_values_type>)
```
as we are overwriting `B` with the solution in this function.

batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
```diff
@@ -0,0 +1,108 @@
...
+}  // namespace Impl
+
+template <typename ArgTrans>
+struct SerialGbtrs<ArgTrans, Algo::Gbtrs::Unblocked> {
+  template <typename AViewType, typename PivViewType, typename BViewType>
+  KOKKOS_INLINE_FUNCTION static int invoke(const AViewType &A, const PivViewType &piv, const BViewType &b, const int kl,
+                                           const int ku) {
+    // quick return if possible
+    if (A.extent(1) == 0) return 0;
+    auto info = Impl::checkGbtrsInput(A, piv, b, kl, ku);
```
lucbv, line 99: I would run the checks before the quick return as this might reveal a mismatch in argument sizes and inform the user on a potential issue that will otherwise be skipped...

lucbv (changes requested):
I have not read the Internal and Test files yet but here are some initial quick remarks that can easily be addressed while I get to the rest of the files.

batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
yasahi-hpc: Sure

batched/dense/src/KokkosBatched_Gbtrs.hpp
```diff
@@ -0,0 +1,68 @@
...
+namespace KokkosBatched {
+
+/// \brief Serial Batched Gbtrs:
+///
+/// Solve a system of linear equations
+///   A * X = B or A**T * X = B or A**H * X = B
+///   with a general band matrix A using the LU factorization computed
+///   by gbtrf.
+/// \tparam ArgTrans: Type indicating whether the A (Trans::NoTranspose), or A**T (Trans::Transpose) or A**H
+/// (Trans::ConjTranspose) is used. \tparam ArgAlgo: Type indicating the blocked (KokkosBatched::Algo::Gbtrs::Blocked)
```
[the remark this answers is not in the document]
yasahi-hpc: Sorry for that. I have fixed accordingly

batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
yasahi-hpc: Actually, there is one check from lapack implementation that can be violated for `n == 0` case, assuming b size is also `0 (ldb == 0)` . If we think `n == 0` is a valid input, then we also need to modify this check.

```C++
const int ldb = b.extent(0);
if (ldb < Kokkos::max(1, n)) {
  Kokkos::printf(
      "KokkosBatched::gbtrs: leading dimension of b must not be smaller than "
      "max(1, n): "
      "ldb = %d, n = %d\n",
      ldb, n);
  return 1;
}
```

yasahi-hpc:
@lucbv 
Thank you for your review. I have fixed accordingly. I also fixed the error messages. 
As for the quick return, I still kept my implementation as commented above. 
But I am open to modify this, if you prefer another solution.

batched/dense/impl/KokkosBatched_Gbtrs_Serial_Impl.hpp
yasahi-hpc: We can modify this check to

```C++
const int ldb = b.extent(0);
if (ldb < Kokkos::max(0, n)) {
  Kokkos::printf(
      "KokkosBatched::gbtrs: leading dimension of b must not be smaller than "
      "max(0, n): "
      "ldb = %d, n = %d\n",
      ldb, n);
  return 1;
}
```

lucbv:
By the way I think the autotester is not running the tests because you are using two different accounts. You might need to trigger the runs with your `yasahi-hpc` account.

batched/dense/unit_test/Test_Batched_SerialGbtrs.hpp
```diff
@@ -0,0 +1,403 @@
...
+
+template <typename DeviceType, typename ScalarType, typename ParamTagType, typename AlgoTagType>
+int test_batched_gbtrs() {
+#if defined(KOKKOSKERNELS_INST_LAYOUTLEFT)
+  {
+    using LayoutType = Kokkos::LayoutLeft;
+    Test::Gbtrs::impl_test_batched_gbtrs_analytical<DeviceType, ScalarType, LayoutType, ParamTagType, AlgoTagType>(1);
+    Test::Gbtrs::impl_test_batched_gbtrs_analytical<DeviceType, ScalarType, LayoutType, ParamTagType, AlgoTagType>(2);
+    for (int i = 0; i < 10; i++) {
+      int k = 1;
```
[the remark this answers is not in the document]
yasahi-hpc: That is true. I will add test cases with multiple `k` sizes.

yasahi-hpc:
> By the way I think the autotester is not running the tests because you are using two different accounts. You might need to trigger the runs with your `yasahi-hpc` account.

Makes sense. I will update git configuration of my environment.
Thank you for the review and remark

yasahi-hpc:
@lucbv An issue has been detected for `ConjTrans` case with larger k, which has been resolved. May I have another review please

lucbv (approved):
Thanks for the updates and glad it revealed an issue in the PR : )

PR #2540, opened 2025-03-14
update develop version to 4.6.99

ndellingwood:


lucbv (approved):
Okay, let's not forget to also update the workflows when we push the release out!

PR #2545, opened 2025-03-17
std::filesystem library for gcc < 9.1 and Clang < 9.0

cwpearson:


lucbv (approved):
Ah the beauty of the non-standard implementations of the c++ library...

Issue #2238, opened 2024-06-10
Issue with installed Kokkos-Kernels on Frontier


Rombur:
I need to install Kokkos-Kernels on Frontier using spack. The installation process went smoothly but I get an error when I try using the installed version. When I call `find_package(KokkosKernels)`, I get the following error:

```
CMake Error at /autofs/nccs-svm1_proj/ums018/spack/opt/spack/linux-sles15-zen/gcc-7.5.0/kokkos-kernels-4.3.00-623m36evj3nkkygvgryninkyklfzlp7k/lib64/cmake/KokkosKernels/KokkosKernelsConfig.cmake:34 (SET_TARGET_P
ROPERTIES):                                                                                                                                                                                                        
  SET_TARGET_PROPERTIES called with incorrect number of arguments.                                                                                                                                                 
Call Stack (most recent call first):                                                                                                                                                                               
  CMakeLists.txt:4 (find_package)                                                                                                                                                                                  
                                                                                                                                                                                                                   
                                                                                                                                                                                                                   
CMake Error at /autofs/nccs-svm1_proj/ums018/spack/opt/spack/linux-sles15-zen/gcc-7.5.0/kokkos-kernels-4.3.00-623m36evj3nkkygvgryninkyklfzlp7k/lib64/cmake/KokkosKernels/KokkosKernelsConfig.cmake:42 (SET_TARGET_PROPERTIES):
  SET_TARGET_PROPERTIES called with incorrect number of arguments.
Call Stack (most recent call first):
  CMakeLists.txt:4 (find_package)


CMake Error at /autofs/nccs-svm1_proj/ums018/spack/opt/spack/linux-sles15-zen/gcc-7.5.0/kokkos-kernels-4.3.00-623m36evj3nkkygvgryninkyklfzlp7k/lib64/cmake/KokkosKernels/KokkosKernelsConfig.cmake:49 (SET_TARGET_PROPERTIES):
  SET_TARGET_PROPERTIES called with incorrect number of arguments.
Call Stack (most recent call first):
  CMakeLists.txt:4 (find_package)
```

Looking at `lib64/cmake/KokkosKernels/KokkosKernelsConfig.cmake`, I can see that there is an issue. The path after `INTERFACE_INCLUDE_DIRECTORIES` is duplicated:

```
IF(NOT TARGET KokkosKernels::ROCBLAS)
ADD_LIBRARY(KokkosKernels::ROCBLAS UNKNOWN IMPORTED)
[...]

PR #2550, opened 2025-03-17
CMake modules: improve finding of ROC TPLs when ROC{package}_ROOT is specified

japlews:
Previously, no information was given to the library finding routine.


cwpearson (approved):

cwpearson:
The failing DCO test is because your commit is not signed-off (`git commit --amend -s` to sign off your existing commit).

lucbv:
@japlews is this an issue that you have when building with Trilinos or using Kokkos Kernels on its own?
I see that Trilinos does the following in it own `FindTPLROCBLAS.cmake`:
```
find_package(ROCBLAS)

if(ROCBLAS_FOUND)
    tribits_extpkg_create_imported_all_libs_target_and_config_file( ROCBLAS
        INNER_FIND_PACKAGE_NAME  ROCBLAS
        IMPORTED_TARGETS_FOR_ALL_LIBS  roc::rocblas )
endif()
```
We use to have very similar logic up until issue #2238 (and Trilinos issue [#13086](https://github.com/trilinos/Trilinos/issues/13086)) which led to PR #2250.
I remain unconvinced by the work that was done at that point and would actually to try to go back to the simpler version of the TPL module that just calls `find_package(ROCBLAS)`.
Beyond making things simpler this would likely allow us to make things work again with spack

cwpearson:
I agree about ditching as much of our custom stuff as possible in favor of FIND_PACKAGE

japlews:
@lucbv and @cwpearson I also support abandoning this in favor of `find_package`! :) ROCm packages seem to be all CMake based, and so finding them should be possible in nearly any installation scenario. Also @lucbv yes, this is in the context of building Trilinos and other products against a standalone KokkosKernels.

lucbv:
@japlews I propose an alternative approach in #2554 it prioritizes using find_package, let me know if that looks reasonable to you. It currently only addresses rocBLAS but will replicate changes to rocSPARSE and rocSOLVER. The approach works for building as an internal or external package of Trilinos.

lucbv:
Never mind, I see that something is failing in the PR tester so clearly the logic is not foolproof yet...
We could merge your changes for the time being if that helps, eventually I want to move to something like #2554

japlews:
@lucbv This is a priority for supporting external kokkos-kernels, probably in the next Sierra release in the summer (June timeframe). It's not very urgent, but we will probably want to start testing a little later this spring (April).

japlews:
@lucbv no problem! I actually switched over to unsetting the `ROC*_ROOT` in imported ROCm packages and just relying on `find_package()`, which works. Thanks a lot for the more robust fix, though!

PR #2553, opened 2025-03-18
Fix #2336: in StaticCrsGraph use default_size_type

### LINKED ISSUE #2336: inconsistent default size type between `CrsMatrix` and `StaticCrsGraph`
@brian-kelley 
@romintomasetti

In the following PR:
- https://github.com/kokkos/kokkos-kernels/pull/2149

the default size types of `KokkosSparse::CrsMatrix` and several other matrix types were changed to `default_size_type`. This is typically `int` or `size_t` depending on the configuration of Kokkos-kernels.

However, in `Kokkos::StaticCrsGraph`, the default size type is taken from a `ViewTraits` instance. Typically, that size type is in turn a size type of a space, and it need not be `int` or `size_t`.

- https://github.com/kokkos/kokkos/blob/a14a5d9bee2f84e1b160df203ed2dd6c8f2060c3/containers/src/Kokkos_StaticCrsGraph.hpp#L256-L260

The issue is thus that these default size types are not necessarily the same. And because `KokkosSparse::CrsMatrix` and `Kokkos::StaticCrsGraph` are so interrelated, this discrepancy can lead to problems when using eg. the `KokkosSparse::CrsMatrix` constructor that takes a `Kokkos::StaticCrsGraph` among its arguments.

This issue is thus a suggestion to consider whether further consistency across the size types may be worth pursuing.

brian-kelley:
to match the default size type for ``CrsMatrix``, ``BsrMatrix``, etc.
Fixes example provided by @MalachiTimothyPhillips that relies on ``StaticCrsGraph`` and ``CrsMatrix`` having the same default ``size_type``. (except for that we also brought ``StaticCrsGraph`` into ``KokkosSparse::``, but that's not unrelated)

### DIFF
--- a/sparse/src/KokkosSparse_StaticCrsGraph.hpp
+++ b/sparse/src/KokkosSparse_StaticCrsGraph.hpp
@@ -18,6 +18,7 @@
 #define KOKKOSSPARSE_STATICCRSGRAPH_HPP_
 
 #include <Kokkos_Core.hpp>
+#include "KokkosKernels_default_types.hpp"
 
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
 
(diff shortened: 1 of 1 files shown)


lucbv (approved):
The change by itself is fine with me, the only question is whether this might create some issues for customers as it is part of the public API?
I do not think that we can easily deprecated this so the only option would be a guard that switches between old and new behavior?
I am happy to just merge and ask on slack if anyone has an issue with this?

brian-kelley:
I don't really like the idea of adding a switch just to control this. We could just wait until 5.0 to actually release this change, right?

lucbv:
I think we can ask on slack and our internal customers if the change creates any issues. If we get no complaints we can just have in 4.7.0 no problem for me.
Merging

PR #2554, opened 2025-03-19
rocTPLs: modifying cmake logic to prioritize find_package

lucbv:
We know first try to use find_package and then pull out the properties from it. Warn users that we will only support that mode of detection in the future.


japlews:
@lucbv I like this approach better.

cwpearson (approved):

lucbv:
Tested in PR, on Caraway, with Trilinos source override and with Trilinos as a TPL it all builds fine

PR #2558, opened 2025-03-21
Assume independent thread scheduling for Blackwell

brian-kelley:
Like all CUDA architectures since Volta, Blackwell has independent thread scheduling. I don't have access to any hardware to test this but it probably fixes SpGEMM on Blackwell (https://github.com/trilinos/Trilinos/issues/13907).


lucbv (approved):

maartenarnst:
We compiled Trilinos with this patch on a system with a `blackwell120`. In our build, we used the current `Kokkos` `develop` branch to have the architecture flags. Trilinos compiled fine and the tests that we activated in our build all passed with this patch (Tpetra, MueLu, Stokhos, Panzer, Anasazi).

maartenarnst (approved):

cwpearson:
I'd propose a unit test that checks this macro against the compute capability reported by the CUDA device properties (presumably, that test will immediately fail on the next Nvidia architecture and we can extend this macro right away).

cwpearson (approved):

PR #2559, opened 2025-03-21
Documentation changes enforcement

lucbv:
Adding an action and python script to check if public headers have been modified.
If so check that associated documentation files are also modified to reflect the API change.
If not, return an exit code to enforce documentation of API change.
The check can be avoided by using the `CI: skip-docs` label in the PR


.github/workflows/docs.yml
```diff
@@ -10,6 +10,22 @@ permissions:
   contents: none
 
 jobs:
+  check_updates:
+    runs-on: [macos-latest]
```
yasahi-hpc, line 14: Any reason to prefer macOS over ubuntu?

scripts/check_api_updates.py
```diff
@@ -0,0 +1,61 @@
+src_directories = ["batched/dense/src",
+                   "batched/sparse/src",
+                   "blas/src",
+                   "common/src",
+                   "graph/src",
+                   "lapack/src",
+                   "ode/src",
+                   "sparse/src"]
```
yasahi-hpc, line 8:
```suggestion
#@HEADER
# ************************************************************************
#
#                        Kokkos v. 4.0
#       Copyright (2022) National Technology & Engineering
#               Solutions of Sandia, LLC (NTESS).
#
# Under the terms of Contract DE-NA0003525 with NTESS,
# the U.S. Government retains certain rights in this software.
#
# Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
# See https://kokkos.org/LICENSE for license information.
# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
#
#@HEADER

"""
This script checks for API updates by comparing modified source files with 
their corresponding documentation files, and reports missing or incomplete updates.

It reads a list of modified files from 'modified_files.txt', filters those that are in 
the source directories, and then checks whether any modifications require new or updated 
documentation.
"""

import sys

src_directories = [
    "batched/dense/src",
    "batched/sparse/src",
    "blas/src",
    "common/src",
    "graph/src",
    "lapack/src",
    "ode/src",
    "sparse/src"
]
```

yasahi-hpc: Docstrings can be improved. If you can add an example, it is better

scripts/check_api_updates.py
```diff
@@ -0,0 +1,61 @@
...
+
+src_doc_mapping = dict([('lapack/src/KokkosLapack_gesv.hpp', ['docs/source/API/lapack/gesv.rst']),
+                        ('lapack/src/KokkosLapack_svd.hpp', ['docs/source/API/lapack/gesvd.rst']),
+                        ('lapack/src/KokkosLapack_trtri.hpp', ['docs/source/API/lapack/trtri.rst'])])
+
+diffs_list = open("./modified_files.txt", 'r')
+modified_files = [line.strip() for line in diffs_list.readlines()]
+diffs_list.close()
```
yasahi-hpc, line 16:
```suggestion
with open("./modified_files.txt", 'r', encoding='utf-8') as diffs_list:
    modified_files = [line.strip() for line in diffs_list.readlines()]
```

scripts/check_api_updates.py
```diff
@@ -0,0 +1,61 @@
...
+undocumented_changes  = []
+
+# Loop over changed files
+for modified_file in modified_files:
+    # Check if file belongs to one of our source directories
+    if any(src_dir in modified_file for src_dir in src_directories):
+        modified_public_files.append(modified_file)
+        # Look for documentation associated with modified file
+        doc_files = src_doc_mapping.get(modified_file)
+        if doc_files == None:
```
yasahi-hpc, line 29:
```suggestion
        if doc_files is None:
```

scripts/check_api_updates.py
```diff
@@ -0,0 +1,61 @@
...
+    print("")
+    return_value = 1
+
+if undocumented_changes:
+    print("Likely undocumented public files:")
+    for undoc_change in undocumented_changes:
+        print("   "+str(undoc_change))
+    return_value = 1
+
+exit(return_value)
```
yasahi-hpc, line 61:
```suggestion
sys.exit(return_value)
```

yasahi-hpc (changes requested):
I have applied pylint to get the following message.
Probably better to fix these. 
Also, it is nicer if you can detail the expected behavior more in the docstrings

```
************* Module check_api_updates
check_api_updates.py:12:0: C0301: Line too long (101/100) (line-too-long)
check_api_updates.py:1:0: C0114: Missing module docstring (missing-module-docstring)
check_api_updates.py:14:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
check_api_updates.py:29:11: C0121: Comparison 'doc_files == None' should be 'doc_files is None' (singleton-comparison)
check_api_updates.py:61:0: R1722: Consider using 'sys.exit' instead (consider-using-sys-exit)
check_api_updates.py:14:13: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)

-----------------------------------
Your code has been rated at 8.29/10
```

lucbv:
@yasahi-hpc there is no requirement for pylint currently so I do not feel too worried about. I did fix the syntax and logic warning. It still shows some warning for line length, I am aware but do not think we should worry about this unless we are going to enforce pylint on python files.

.github/workflows/docs.yml
yasahi-hpc: Sure

yasahi-hpc:
> @yasahi-hpc there is no requirement for pylint currently so I do not feel too worried about. I did fix the syntax and logic warning. It still shows some warning for line length, I am aware but do not think we should worry about this unless we are going to enforce pylint on python files.

Apart from the license and docstrings, I also do not care too much. 

If I understand correctly, the current CI fails since you only modified `lapack/src/KokkosLapack_trtri.hpp` without modifying the corresponding rst file. The current changes do not seem to impact the APIs, so I am not sure what kind of modifications should we make on the rst file. Comments or suggestions are appreciated

lucbv:
Yes, there is no good mechanism to know if the changes made are "cosmetic" or actual logic that changed. At least the current version will not be blocking but should catch the eye of reviewers that can ask why some changes are not reflected in the documentation.

yasahi-hpc:
> Yes, there is no good mechanism to know if the changes made are "cosmetic" or actual logic that changed. At least the current version will not be blocking but should catch the eye of reviewers that can ask why some changes are not reflected in the documentation.

Do you mean that we can ignore the CI failure due to the undocumented changes if the changes are "cosmetic"?
Probably, we can add a special approval for that case to skip this CI.

lucbv:
@yasahi-hpc a new flag will be used to skip the documentation checks, we could discuss using it to avoid building the documentation as well but that's a different feature and I would want that in another PR.

yasahi-hpc (approved):
LGTM 
Please update the description of this PR to explain the behavior of `CI: skip-docs` label

PR #2568, opened 2025-03-25
sptrsv_cusparse: must call cusparseSpSV_analysis when matrix values have changed

jczhang07:
This is a bug fix.

In theory, we need a new `sptrsv_numeric` to handle cases that a triangular matrix's sparsity pattern is the same but values have changed.

But because there is no  `sptrsv_numeric` now, I refactored `sptrsvcuSPARSE_symbolic` so that it could be called multiple times.


lucbv (approved):
Okay, looks good. If you could do a rebase on develop that would help the CI as it has changed a bit and will run 2x faster now

lucbv (approved):

PR #2569, opened 2025-03-25
unit test for KOKKOSKERNELS_CUDA_INDEPENDENT_THREADS

cwpearson:
Make sure we update KOKKOSKERNELS_CUDA_INDEPENDENT_THREADS for new GPU architectures.

Follow up after https://github.com/kokkos/kokkos-kernels/pull/2558

### DIFF
--- a/common/unit_test/Test_Common.hpp
+++ b/common/unit_test/Test_Common.hpp
@@ -21,6 +21,7 @@
 // #include<Test_Common_float128.hpp>
 #include <Test_Common_set_bit_count.hpp>
 #include <Test_Common_Sorting.hpp>
+#include <Test_Common_CudaIndependentThreads.hpp>
 #include <Test_Common_IOUtils.hpp>
 #include <Test_Common_Error.hpp>
 #include <Test_Common_Version.hpp>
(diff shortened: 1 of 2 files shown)


cwpearson:
@brian-kelley @lucbv conceptually does this seem okay?

lucbv (approved):
Yeah, that's reasonable, although as mentioned before it would be great to have Kokkos Core let us know about this...

cwpearson:
I've returned to thinking this PR is the right approach: it ensures the properties of the code we compiled (`KOKKOSKERNELS_CUDA_INDEPENDENT_THREADS`) matches the properties of the device we're running on (here retrieved via the CUDA runtime API) and ensure they match. Core would have to provide member functions on the execution space instance to give us the device properties...and they do, but they're deprecated. So I think it's a small enough lift for us to do directly ourselves.

Go ahead an merge if you find my reasoning compelling, or please convince me otherwise 😄

brian-kelley:
This is fine with me. This would be a rare situation, but if someone compiled for an older architecture and ran on a newer architecture this would catch it, while a compile-time check couldn't.

PR #2572, opened 2025-03-25
replace uses of get_suggested_vector__size with kk_get_suggested_vector_size

cwpearson:
* `get_suggested_vector__size` is a reserved ID
* it seems like an unneeded wrapper


cwpearson:
Skipping docs as this is in `Impl`

lucbv (approved):
I think it's okay since it is in the `Impl` namespace, otherwise a deprecation would have been required.

PR #2574, opened 2025-03-25
KokkosKernels_HashmapAccumulator.hpp: rename reserved ids

cwpearson:

No public API changes

### DIFF
--- a/common/src/KokkosKernels_HashmapAccumulator.hpp
+++ b/common/src/KokkosKernels_HashmapAccumulator.hpp
@@ -113,13 +113,13 @@ struct HashmapAccumulator {
    * Assumption: hash_begins_ are all initialized to -1.
    */
   KOKKOS_INLINE_FUNCTION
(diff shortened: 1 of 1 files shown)


jgfouca (approved):

common/src/KokkosKernels_HashmapAccumulator.hpp
```diff
@@ -542,30 +542,30 @@ struct HashmapAccumulator {
       const team_member_t & /* teamMember */, const int /* vector_size */, size_type hash, const key_type key,
       const value_type value, volatile size_type *used_size_, const size_type max_value_size_) {
```
lucbv, line 543:
```suggestion
      const value_type value, volatile size_type *used_size_, const size_type max_value_size) {
```

lucbv: We could also just rename this something else to avoid confusion?

lucbv (commented):
Just a name conflict that needs to be resolved

lucbv (approved):

PR #2579, opened 2025-03-27
docs: adding API documentation template

lucbv:
This should make it easier for anyone to start a new documentation file for new or undocumented functions.

### DIFF
--- a/docs/api_template.rst
+++ b/docs/api_template.rst
@@ -1,7 +1,7 @@
 namespace::function
 ###################
 
-Defined in header: :code:`Namespace_function.hpp`
+Defined in header: :code:`<namespace>_<function>.hpp`
 
 .. code:: c++
 
--- a/docs/api_template.rst
+++ b/docs/api_template.rst
@@ -1,0 +1,60 @@
+namespace::function
+###################
+
+Defined in header: :code:`Namespace_function.hpp`
+
+.. code:: c++
+
+  template <class execution_space, class input_type, class output_type>
+  void function(const execution_space& space, const input_type& in, const output_type& out);
+
+  template <class input_type, class output_type>
+  void function(const input_type& in, const output_type& out);
+
+Explain what the function does and what each overall specificities are.
+
+Parameters
+==========
+
+:space: execution space instance
+
+:in: input
+
+:out: output
+
+Type Requirements
+-----------------
+
+- `execution_space` must be a Kokkos `execution space <https://kokkos.org/kokkos-core-wiki/API/core/execution_spaces.html>`_
+
+- `input_type` must be a Kokkos `View <https://kokkos.org/kokkos-core-wiki/API/core/view/view.html>`_
+
+  - ``Kokkos::SpaceAccessibility<execution_space, typename input_type::memory_space>::accessible``
+  - ...
+
+- `output_type` must be a Kokkos `View <https://kokkos.org/kokkos-core-wiki/API/core/view/view.html>`_ that satisfies
+
+  - ``Kokkos::SpaceAccessibility<execution_space, typename output_type::memory_space>::accessible``
+  - ...
+
+Example
+=======
+
+.. code:: cppkokkos
+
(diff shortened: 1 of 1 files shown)


docs/api_template.rst
```diff
@@ -0,0 +1,60 @@
...
+  - ``Kokkos::SpaceAccessibility<execution_space, typename input_type::memory_space>::accessible``
+  - ...
+
+- `output_type` must be a Kokkos `View <https://kokkos.org/kokkos-core-wiki/API/core/view/view.html>`_ that satisfies
+
+  - ``Kokkos::SpaceAccessibility<execution_space, typename output_type::memory_space>::accessible``
+  - ...
+
+Example
+=======
```
yasahi-hpc, line 41: Is example section optional? 
`KokkosBlas::abs` does not have this.

docs/api_template.rst
```diff
@@ -0,0 +1,60 @@
...
+
+.. code:: cppkokkos
```
yasahi-hpc, line 43: When should we use `cppkokkos` over `c++`?

docs/api_template.rst
```diff
@@ -0,0 +1,60 @@
...
+
+  #include "Kokkos_Core.hpp"
+  #include "Namespace_function.hpp"
```
yasahi-hpc, line 46:
```suggestion
  #include <Kokkos_Core.hpp>
  #include <namespace_function.hpp>
```

yasahi-hpc (changes requested):
Overall structure looks good, but I have two comments.

1. I think it is also necessary to include the license at the top
2. Probably, it is better to add a general guidance. It may make sense to define a template as a nested rst file using the code block, whose items are explained in this file.

```rst
   .. code-block:: rst
      
      something
```

docs/api_template.rst
```diff
@@ -0,0 +1,60 @@
+namespace::function
```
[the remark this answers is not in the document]
lucbv: Actually on line 1 we do put something like: `KokkosBlas::abs`, so `namespace::function` not the title you suggest above...

docs/api_template.rst
lucbv: We did not write an example for it yet, but when we have time we should. These examples are also found in `kokkos-kernels/example/wiki/...` so we know they compile and run fine for users

docs/api_template.rst
lucbv: I am not quite sure, we should ask the Kokkos Core folks

docs/api_template.rst
lucbv: I disagree since despite all our serious we are not a standard library: https://stackoverflow.com/questions/21593/what-is-the-difference-between-include-filename-and-include-filename

docs/api_template.rst
yasahi-hpc: OK

docs/api_template.rst
yasahi-hpc: Sure

docs/api_template.rst
yasahi-hpc: Hmm, that is discussing. 
Many Kokkos and Kokkos-kernels examples are written in my way.
Can we get an opinion from someone else?

docs/api_template.rst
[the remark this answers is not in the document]
lucbv: Yes we should have more description in our `contributing` section which applies to both code and documentation.

docs/api_template.rst
lucbv: It looks like kokkos core is ditching cppkokkos, so we can switch to cpp or c++

lucbv:
Should have addressed all the issues now

yasahi-hpc (approved):
LGTM. 
If you would like to ensure that the example is valid, you may embed the tested example in cpp file rather than hard-coding. 

```rst
.. literalinclude:: <path-to-the-example.cpp>
   :language: C++
```

PR #2580, opened 2025-03-30
ConjTrans support for batched team gemm

yasahi-hpc:
This PR aims at improving the implementation and testing of team Gemm (see also #2469).

 - [x] Moving implementation details into `Impl` namespace
 - [x] Check function has been moved to  `KokkosBatched_Gemm_Common_Impl.hpp`
 - [x] Add all the specializations of the combinations of `Non-Trans/Trans/ConjTrans`
 - [x] Add a check function of input to force A, B, C are rank 0, 1, 2 `Kokkos::View`
 - [x] Remove `using namespace KokkosBatched` from the test code
 - [x] Rename variables starting from underscores in the test code 
 - [x] Deprecated warnings for the older implementation details under `KokkosBatched` namespace
 - [x] As a preliminary attempt, an error bound analysis for Gemm operation is introduced

Updated 1/May
- [x] Introduce a helper to avoid illegal usage of stride method
- [x] Merge blocked and unblocked versions

### DIFF
--- a/batched/dense/impl/KokkosBatched_Gemm_Common_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Gemm_Common_Impl.hpp
@@ -29,9 +29,12 @@ KOKKOS_INLINE_FUNCTION static int checkGemmInput([[maybe_unused]] const AViewTyp
   static_assert(Kokkos::is_view_v<BViewType>, "KokkosBatched::gemm: BViewType is not a Kokkos::View.");
   static_assert(Kokkos::is_view_v<CViewType>, "KokkosBatched::gemm: CViewType is not a Kokkos::View.");
 
-  static_assert(AViewType::rank <= 2, "KokkosBatched::gemm: AViewType must have rank 0, 1 or 2.");
-  static_assert(BViewType::rank <= 2, "KokkosBatched::gemm: BViewType must have rank 0, 1 or 2.");
-  static_assert(CViewType::rank <= 2, "KokkosBatched::gemm: CViewType must have rank 0, 1 or 2.");
+  static_assert(AViewType::rank() <= 2, "KokkosBatched::gemm: AViewType must have rank 0, 1 or 2.");
+  static_assert(BViewType::rank() <= 2, "KokkosBatched::gemm: BViewType must have rank 0, 1 or 2.");
+  static_assert(CViewType::rank() <= 2, "KokkosBatched::gemm: CViewType must have rank 0, 1 or 2.");
+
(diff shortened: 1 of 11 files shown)


batched/dense/impl/KokkosBatched_Gemm_Common_Impl.hpp
```diff
@@ -0,0 +1,79 @@
...
+namespace Impl {
+template <typename ArgTransA, typename ArgTransB, typename AViewType, typename BViewType, typename CViewType>
+KOKKOS_INLINE_FUNCTION static int checkGemmInput([[maybe_unused]] const AViewType &A,
+                                                 [[maybe_unused]] const BViewType &B,
+                                                 [[maybe_unused]] const CViewType &C) {
+  static_assert(Kokkos::is_view_v<AViewType>, "KokkosBatched::gemm: AViewType is not a Kokkos::View.");
+  static_assert(Kokkos::is_view_v<BViewType>, "KokkosBatched::gemm: BViewType is not a Kokkos::View.");
+  static_assert(Kokkos::is_view_v<CViewType>, "KokkosBatched::gemm: CViewType is not a Kokkos::View.");
+
+  static_assert(AViewType::rank <= 2, "KokkosBatched::gemm: AViewType must have rank 0, 1 or 2.");
```
lucbv, line 32: Please use `rank()` instead, see View documentation in Kokkos Core API.

lucbv (approved):
Okay just a small comment, we should also check the performance impact, I don't think there should be any but still good to make sure.

batched/dense/impl/KokkosBatched_Gemm_Common_Impl.hpp
yasahi-hpc: Yes, I have updated accordingly

lucbv:
The documentation does not exist yet for batched GEMM so it is up to you if you want to write it from scratch? Otherwise you can apply the label to skip the documentation update. Documentation updates are only required for new APIs and existing APIs that are already documented to avoid burdening bug fixes when an existing API is not yet documented.
Of course I highly appreciate documentation contributions!

yasahi-hpc:
> The documentation does not exist yet for batched GEMM so it is up to you if you want to write it from scratch? Otherwise you can apply the label to skip the documentation update. Documentation updates are only required for new APIs and existing APIs that are already documented to avoid burdening bug fixes when an existing API is not yet documented. Of course I highly appreciate documentation contributions!

Thanks! I have added a doc-skip label. I am willing to add docs in another PR.

lucbv (approved):
Okay, looks reasonable

Issue #2592, opened 2025-04-11
Nightly test failures: several tests fail with `Assertion 'r < extents_type::rank()' failed.`


ndellingwood:
Nightly builds are failing tests due to some apparent incompatibilities with https://github.com/kokkos/kokkos/pull/7427

List of failures (taken from cuda/11.8.0 job on hopper)

```
00:42:23 The following tests FAILED:
00:42:23 	  4 - batched_dla_cuda (Subprocess aborted)
00:42:23 	  5 - batched_gemm_cuda (Subprocess aborted)
00:42:23 	  7 - batched_gemm_serial (Subprocess aborted)
00:42:23 	 16 - sparse_cuda (Subprocess aborted)
00:42:23 	 18 - sparse_serial (Subprocess aborted)
00:42:23 	 20 - ode_cuda (Subprocess aborted)
00:42:23 	 21 - ode_serial (Subprocess aborted)
```

The failures seem to have in common a failed assertion check `Assertion 'r < extents_type::rank()' failed.` , some sample snips:

```
00:42:06 KokkosKernels_ode_serial: /home/jenkins/blake-new/workspace/KokkosKernels_Nightly_Blake_Cuda_11_8_0_Gcc_11_3_0_Hopper90-cusparse-cublas/Build/kokkos-install/include/mdspan/../experimental/__p2642_bits/layout_padded.hpp:800: constexpr Kokkos::Experimental::layout_right_padded<padding_value>::mapping<_Extents>::index_type Kokkos::Experimental::layout_right_padded<padding_value>::mapping<_Extents>::stride(Kokkos::Experimental::layout_right_padded<padding_value>::mapping<_Extents>::rank_type) const [with Extents = Kokkos::extents<long unsigned int, 18446744073709551615>; long unsigned int PaddingValue = 18446744073709551615; Kokkos::Experimental::layout_right_padded<padding_value>::mapping<_Extents>::index_type = long unsigned int; Kokkos::Experimental::layout_right_padded<padding_value>::mapping<_Extents>::rank_type = long unsigned int]: Assertion `r < extents_type::rank()' failed.
```

```
[...]

Issue #2591, opened 2025-04-11
Error: array subscript 1 is above array bounds of 'const long unsigned int [1]'


tpadioleau:
Hi there,

I got the following compilation error in the nightly of DDC:
```
In member function 'constexpr const T& Kokkos::detail::possibly_empty_array<T, N>::operator[](size_t) const [with T = long unsigned int; long unsigned int N = 1]',
    inlined from 'constexpr Kokkos::layout_stride::mapping<Extents>::index_type Kokkos::layout_stride::mapping<Extents>::stride(rank_type) const [with Extents = Kokkos::extents<long unsigned int, 18446744073709551615>]' at /data/opt/kokkos/include/experimental/__p0009_bits/layout_stride.hpp:555:33,
    inlined from 'constexpr Kokkos::Impl::BV::BasicView<ElementType, Extents, LayoutPolicy, AccessorPolicy>::index_type Kokkos::Impl::BV::BasicView<ElementType, Extents, LayoutPolicy, AccessorPolicy>::stride(rank_type) const [with ElementType = double; Extents = Kokkos::extents<long unsigned int, 18446744073709551615>; LayoutPolicy = Kokkos::layout_stride; AccessorPolicy = Kokkos::Impl::SpaceAwareAccessor<Kokkos::HostSpace, Kokkos::Impl::ReferenceCountedAccessor<double, Kokkos::HostSpace, Kokkos::default_accessor<double> > >]' at /data/opt/kokkos/include/View/Kokkos_BasicView.hpp:605:24,
    inlined from 'constexpr std::enable_if_t<is_integral_v<iType>, long unsigned int> Kokkos::View<DataType, Properties>::stride(iType) const [with iType = int; DataType = double*; Properties = {Kokkos::LayoutStride, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<0>}]' at /data/opt/kokkos/include/Kokkos_View.hpp:267:26,
    inlined from 'constexpr size_t Kokkos::View<DataType, Properties>::stride_1() const [with DataType = double*; Properties = {Kokkos::LayoutStride, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<0>}]' at /data/opt/kokkos/include/Kokkos_View.hpp:252:68,
[...]

PR #2593, opened 2025-04-12
Fix illegal use of stride

crtrott:
This is the initial fix for the illegal use of View::stride (and actually also View::extent going forward) reported in #2592 and #2591. I only worked on Serial on my Mac laptop. Let's see which other backends trigger stuff which fail in the testing.

### DIFF
--- a/batched/dense/impl/KokkosBatched_HostLevel_Gemm_Serial_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_HostLevel_Gemm_Serial_Impl.hpp
@@ -153,8 +153,22 @@ class BatchedSerialGemm {
     // matrix transpositions, here we must perform the GEMM on:
     // row_vec x col_vec, which is svA_row' x svB_col to compute the element
     // of C.
-    KokkosBatched::SerialGemm<Trans::Transpose, Trans::NoTranspose, ArgMode>::invoke(alpha, svA_row, svB_col, beta,
-                                                                                     svC_ele);
+    // KokkosBatched::SerialGemm<Trans::Transpose, Trans::NoTranspose, ArgMode>::invoke(alpha, svA_row, svB_col, beta,
+    //                                                                                    svC_ele);
+    using ValueType             = typename CViewType::value_type;
+    ValueType svA_row_x_svB_col = 0;
+    // KokkosBatched::SerialDotInternal::invoke(svA_row.extent(0), svA_row.data(), svA_row.stride(0),
+    //   svB_col.data(), svB_col.stride(0), &svA_row_x_svB_col);
+
+    using ats = Kokkos::ArithTraits<ValueType>;
+    // iC[0]      = ValueType(0);
+#if defined(KOKKOS_ENABLE_PRAGMA_UNROLL)
+#pragma unroll
+#endif
+    for (int j = 0; j < int(svA_row.extent(0)); ++j) {
+      svA_row_x_svB_col += ats::conj(svA_row(j)) * svB_col(j);
+    }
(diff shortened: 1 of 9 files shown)


ndellingwood:
I'm running some spot checks with Trilinos to see if anything shows up

cwpearson:
Adding blas 1 docs in https://github.com/kokkos/kokkos-kernels/pull/2599

batched/dense/impl/KokkosBatched_HostLevel_Gemm_Serial_Impl.hpp
```diff
@@ -153,8 +153,22 @@ class BatchedSerialGemm {
     // matrix transpositions, here we must perform the GEMM on:
     // row_vec x col_vec, which is svA_row' x svB_col to compute the element
     // of C.
-    KokkosBatched::SerialGemm<Trans::Transpose, Trans::NoTranspose, ArgMode>::invoke(alpha, svA_row, svB_col, beta,
-                                                                                     svC_ele);
+    // KokkosBatched::SerialGemm<Trans::Transpose, Trans::NoTranspose, ArgMode>::invoke(alpha, svA_row, svB_col, beta,
+    //                                                                                    svC_ele);
+    using ValueType             = typename CViewType::value_type;
+    ValueType svA_row_x_svB_col = 0;
+    // KokkosBatched::SerialDotInternal::invoke(svA_row.extent(0), svA_row.data(), svA_row.stride(0),
```
cwpearson, line 160: This didn't work?

cwpearson (changes requested):

batched/dense/impl/KokkosBatched_HostLevel_Gemm_Serial_Impl.hpp
crtrott: No I couldn't make it work.

ndellingwood:
I tested Trilinos a Serial, Cuda and Hip builds with kokkos@4.6.00 and this PR (the changes needed here were exposed by the View refactor but do not require those changes for testing), and a few packages enabled Tpetra, Sacado, and Ifpack2. No regressions with the changes in this PR

crtrott:
As far as i can tell the old view never returned zero. So if they expected that i am not sure what they do?

tpadioleau:
Could it be of any help to implement a function that extends a 0D or 1D View to a 2D View by adding extra dimensions of size 1 ? It seems well-posed for LayoutRight and LayoutLeft. In the case of the LayoutStride the strides of the new dimensions may not be chosen uniquely. I guess we could take 0 or 1, whatever fits better for Kokkos Kernels ?

yasahi-hpc:
> Could it be of any help to implement a function that extends a 0D or 1D View to a 2D View by adding extra dimensions of size 1 ? It seems well-posed for LayoutRight and LayoutLeft. In the case of the LayoutStride the strides of the new dimensions may not be chosen uniquely. I guess we could take 0 or 1, whatever fits better for Kokkos Kernels ?

For the batched solvers, the internal views can have `LayoutStride` because they can be subviews of original views.

batched/dense/impl/KokkosBatched_HostLevel_Gemm_Serial_Impl.hpp
cwpearson: We'll track this in https://github.com/kokkos/kokkos-kernels/issues/2611

lucbv (approved):

ndellingwood:
Lots of passing nightly tests again, thanks all :)

PR #2599, opened 2025-04-15
Blas1::set docs

cwpearson:


ndellingwood (approved):

lucbv:
I am still not quite sure how/where to document this type of hierarchical BLAS thing, or whatever we want to call it. Also I believe we might need to run this through IR first.

cwpearson:
Do you want me to run this through IR or do you have a different proposal about documenting this, or just table it for now?

lucbv (approved):

PR #2605, opened 2025-04-23
Par_ilut enhancements and fixes

jgfouca:
1) Improve documentation of parameters and confusing code segments
2) Do not let leftover junk in L_entries and U_entries to cause the
   sorted check to fail. Those are output-only parameters.
3) Allow the user to skip the residual computation by setting
   residual_norm_delta_stop to zero (or < 0). This can save some
   memory if memory is tight.

### DIFF
--- a/sparse/src/KokkosSparse_par_ilut.hpp
+++ b/sparse/src/KokkosSparse_par_ilut.hpp
@@ -325,13 +325,24 @@ void par_ilut_numeric(KernelHandle* handle, ARowMapType& A_rowmap, AEntriesType&
   if (handle->get_par_ilut_handle()->is_symbolic_complete() == false) {
     std::ostringstream os;
     os << "KokkosSparse::Experimental::par_ilut_numeric: par_ilut_symbolic "
-          "must be "
-          "called before par_ilut_numeric.";
+          "must be called before par_ilut_numeric.";
     KokkosKernels::Impl::throw_runtime_exception(os.str());
   }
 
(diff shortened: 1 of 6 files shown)


sparse/src/KokkosSparse_par_ilut.hpp
```diff
@@ -330,8 +330,13 @@ void par_ilut_numeric(KernelHandle* handle, ARowMapType& A_rowmap, AEntriesType&
...
   }
 
-  KK_REQUIRE_MSG(KokkosSparse::Impl::isCrsGraphSorted(L_rowmap, L_entries), "L is not sorted");
-  KK_REQUIRE_MSG(KokkosSparse::Impl::isCrsGraphSorted(U_rowmap, U_entries), "U is not sorted");
+  // These are outputs, so we don't want whatever junk that may be in these to
+  // cause the isSorted check to fail.
+  Kokkos::deep_copy(L_entries, 0);
+  Kokkos::deep_copy(U_entries, 0);
+
+  KK_REQUIRE_MSG(KokkosSparse::Impl::isCrsGraphSorted(L_rowmap, L_entries), "L rowmap is not sorted");
```
brian-kelley, line 338: If L_entries and U_entries are filled with 0, and their rowmaps are still valid, then isCrsGraphSorted will always return true. But if the rowmap contains garbage values then it could segfault.

To test if L_rowmap itself is sorted, there's ``Kokkos::Experimental::is_sorted(exec_space(), L_rowmap)``.

brian-kelley (approved):

Issue #2612, opened 2025-04-29
Nightly compilation errors with Trilinos: KokkosSparse_par_ilut.hpp:356:18: error: no member named 'is_sorted' in namespace 'Kokkos::Experimental'


ndellingwood:
Nightly builds of Trilinos are failing compilation with output:

```
17:43:31 /home/jenkins/blake-new/workspace/KokkosEco_Trilinos_Blake_OneAPI2023_2_0_ICPX_Serial/Trilinos/kokkos-kernels/sparse/src/KokkosSparse_par_ilut.hpp:356:18: error: no member named 'is_sorted' in namespace 'Kokkos::Experimental'; did you mean 'std::is_sorted'?
```

This follows merge of PR #2605 , @jgfouca can you help advise?

The Trilinos nightly builds have the options `-D Kokkos_ENABLE_DEPRECATED_CODE_4:BOOL=ON` set (default in Trilinos), in case this is relevant

ndellingwood:
Blake reproducer notes (llvm, Serial backend):

```bash
# Repos
git clone -b develop https://github.com/trilinos/Trilinos.git
TRILINOS_DIR=$PWD/Trilinos
git clone -b develop https://github.com/kokkos/kokkos.git
KOKKOS_DIR=$PWD/kokkos
git clone -b develop https://github.com/kokkos/kokkos-kernels.git
KOKKOSKERNELS_DIR=$PWD/kokkos-kernels
cd $TRILINOS_DIR
ln -s ${PWD}/../kokkos kokkos
ln -s ${PWD}/../kokkos-kernels kokkos-kernels
cd ..
mkdir -p Build
cd Build

# Setup environment
salloc -N 1 -p all

module load cmake llvm openblas/0.3.23
module list

# Configure Trilinos
cmake \
  -D CMAKE_CXX_COMPILER="`which clang++`" \
  -D CMAKE_C_COMPILER="`which clang`" \
  -D CMAKE_Fortran_COMPILER="`which gfortran`" \
  -D CMAKE_CXX_FLAGS="-g -fPIC" \
  -D CMAKE_C_FLAGS="-g -fPIC" \
  -D CMAKE_Fortran_FLAGS="-g -fPIC" \
  -D CMAKE_BUILD_TYPE:STRING=DEBUG \
  -D BUILD_SHARED_LIBS:BOOL=OFF \
  -DTPL_ENABLE_MPI=OFF \
[...]

PR #2613, opened 2025-04-29
Add missing header to KokkosSparse_par_ilut.hpp

ndellingwood:
Including the `std_algorithms/Kokkos_IsSorted.hpp` header is needed for Trilinos builds to compile

Addresses issue #2612

### DIFF
--- a/sparse/src/KokkosSparse_par_ilut.hpp
+++ b/sparse/src/KokkosSparse_par_ilut.hpp
@@ -40,6 +40,7 @@
 #include "KokkosKernels_Error.hpp"
 #include "KokkosSparse_par_ilut_symbolic_spec.hpp"
 #include "KokkosSparse_par_ilut_numeric_spec.hpp"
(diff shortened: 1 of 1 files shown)


lucbv (approved):

jgfouca (approved):
Thanks, Nate!

brian-kelley (approved):
@lucbv  @ndellingwood To answer the question from today's meeting (of how this compiled in a standalone build), it seems that the code using ``is_sorted`` was only instantiated in the sparse unit tests, not libkokkoskernels. The sparse tests also include ``KokkosSparse_coo2crs_impl.hpp`` which pulls in ``Kokkos_StdAlgorithms.hpp``.

I think this explains why Ifpack2 hit the error but standalone KokkosKernels did not.

ndellingwood:
@brian-kelley thanks for tracking that down, very helpful to know!

PR #2615, opened 2025-04-30
Avoid warning for explicit instantiations with storage class

masterleinad:
Latest oneAPI release warn about the storage specifier (`static`) in the explicit specialization. `gcc` doesn't allow explicit specializations in classes at all while MSVC requires the duplication of the storage specifier.
We can just use `if constexpr` to avoid the need for a explicit specialization altogether.


lucbv (approved):
Approving but you'll have to format this before it can be merged.

lucbv (approved):

PR #2616, opened 2025-04-30
Avoid using pointers with Kokkos::single

masterleinad:
oneAPI 2025.0.5 became the default compiler for `Aurora`. 

Unfortunately, it can't deal with pointers in `sycl::select_from_group` which ends up being called for `Kokkos::single` with a scalar input. `Kokkos Kernels` only checks that the value returned after calling `Kokkos::single` is distributed after `allocate_chunk` is called. Since, we never really care about the dereferenced value, we might as well just `reinterpret_cast` to `uintptr_t` instead. This is also the reason why using a pointer to `volatile` should never be necessary.

### DIFF
--- a/sparse/impl/KokkosSparse_bspgemm_impl_kkmem.hpp
+++ b/sparse/impl/KokkosSparse_bspgemm_impl_kkmem.hpp
@@ -376,17 +376,18 @@ struct KokkosBSPGEMM<HandleType, a_row_view_t_, a_lno_nnz_view_t_, a_scalar_nnz_
           nnz_lno_t *globally_used_hash_indices = NULL;
 
           if (global_memory_hash_size > thread_shmem_key_size) {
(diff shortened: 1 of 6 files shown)


brian-kelley (approved):

masterleinad:
> @masterleinad After some thought, could the `std::is_pointer<T>` case be specialized by `single` on the SYCL backend? By that I mean call `sycl::select_from_group` with `uintptr_t` to do the broadcast, but then cast it back to `T` and assign that to the reference the user passed in.

Let me look into it.

cwpearson (approved):

masterleinad:
See https://github.com/kokkos/kokkos/pull/8051.

masterleinad:
Fixed in https://github.com/kokkos/kokkos/pull/8051.

brian-kelley:
@masterleinad Thanks for fixing this in core!

PR #2617, opened 2025-04-30
Workflows: removing rocm 5.6.1 from at2

lucbv:
This older rocm toolchain is very slow compared to the new rocm 6.2.1 so it is disabled in favor of that newer version.


cwpearson (approved):
🥲 it's beautiful

PR #2618, opened 2025-04-30
osx workflow: removing explicit parallelism levels

lucbv:
Switching to implicit parallelism levels for build and installation as it may change over time based on github hardware and we don't want to muck with it everytime...

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -69,49 +69,49 @@ jobs:
           mkdir -p kokkos/{build,install}
           cd kokkos/build
           cmake \
-          -DKokkos_ENABLE_${{ matrix.backend }}=ON \
-          -DCMAKE_CXX_FLAGS="-Werror" \
-          -DCMAKE_CXX_STANDARD=17 \
-          -DKokkos_ENABLE_COMPILER_WARNINGS=ON \
-          -DKokkos_ENABLE_DEBUG_BOUNDS_CHECK:BOOL=${{ matrix.debug_bounds_check }} \
-          -DKokkos_ENABLE_DEPRECATED_CODE_3=OFF \
-          -DKokkos_ENABLE_TESTS=OFF \
-          -DKokkos_ENABLE_DEPRECATED_CODE_4=OFF \
-          -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-          -DCMAKE_INSTALL_PREFIX=$PWD/../install \
-          ..
+	  -S "$PWD/.." \
+	  -B "$PWD" \
+          -D CMAKE_CXX_FLAGS="-Werror" \
+          -D CMAKE_CXX_STANDARD=17 \
+          -D CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+          -D CMAKE_INSTALL_PREFIX=$PWD/../install \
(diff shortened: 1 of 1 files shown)


cwpearson:
My macOS doesn't have `nproc` sadly

lucbv:
Will probably want to clean that then, I think `cmake --parallel ` will still try maximum concurrency automatically though?

cwpearson:
If I'm reading the manual right, `--parallel` needs a number after it.

Could try `sysctl -n hw.physicalcpu` as a replacement for `nproc`, not sure if that works in a GitHub action.

lucbv:
Okay, switching back to setting the parallelism level explicitly but using `--parallel 3` based on the three cores allocated to the latest osx runners.

.github/workflows/osx.yml
```diff
@@ -69,49 +69,49 @@ jobs:
...
+          -S "$PWD/.." \
+          -B "$PWD" \
+          -D CMAKE_CXX_FLAGS="-Werror" \
+          -D CMAKE_CXX_STANDARD=17 \
+          -D CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+          -D CMAKE_INSTALL_PREFIX=$PWD/../install \
+          -D Kokkos_ENABLE_${{ matrix.backend }}=ON \
+          -D Kokkos_ENABLE_COMPILER_WARNINGS=ON \
+          -D Kokkos_ENABLE_DEBUG_BOUNDS_CHECK:BOOL=${{ matrix.debug_bounds_check }} \
+          -D Kokkos_ENABLE_DEPRECATED_CODE_3=OFF \
```
ndellingwood, line 81:
```suggestion
```
This can be removed as part of cleanup, but not requiring a rerun of CI if you'd rather we defer that to a later and more broad cleanup

.github/workflows/osx.yml
```diff
@@ -69,49 +69,49 @@ jobs:
...
-          -DKokkos_ENABLE_DEPRECATED_CODE_3=OFF \
-          -DKokkos_ENABLE_TESTS=OFF \
-          -DKokkos_ENABLE_DEPRECATED_CODE_4=OFF \
-          -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-          -DCMAKE_INSTALL_PREFIX=$PWD/../install \
-          ..
+          -S "$PWD/.." \
+          -B "$PWD" \
+          -D CMAKE_CXX_FLAGS="-Werror" \
+          -D CMAKE_CXX_STANDARD=17 \
```
ndellingwood, line 75: Do we know if the compiler in this workflow supports c++20? We might want to consider changing this soon if so

ndellingwood (approved):
Thanks @lucbv !

.github/workflows/osx.yml
lucbv: I'll add it in the PR and if does not pass we will know hahaha

.github/workflows/osx.yml
lucbv: Okay, will add that with the c++ standard change

lucbv:
Note that the increased parallelism level does not speed-up the build, it might be memory bound? Hard to say but since the PR now includes other clean-ups let us still continue testing and merge it when ready.

PR #2629, opened 2025-05-06
Fix unused variable warning

brian-kelley:
Suprised this was never caught but this variable was unused. It's not really necessary since AMatrix can implicitly convert to AMatrix_Internal but it's not a bad thing to be explicit.

@lucbv This changes a source header but there are no doc changes


lucbv (approved):

brian-kelley:
@lucbv Can this be merged by the red button? It looks like a couple of the AT2 builds are stuck in 'queued' since yesterday.

PR #2630, opened 2025-05-06
Improve TPL singletons

brian-kelley:
Add a simple smart-pointer class ``KokkosKernels::Impl::Singleton<T>`` which is kind of like ``std::unique_ptr``, except its object is freed by ``Kokkos::finalize`` instead of its destructor.

Compared to unique_ptr, this fixes a double-free bug in programs which called ``Kokkos::finalize`` as an atexit hook. In this situation the unique_ptr could be destroyed before Kokkos::finalize, so our Kokkos::finalize hook would try to free it again. This PR adds a test in common for this case.

This also reduces some duplicate code in the TPL singletons. The logic to check initialized, construct object and push finalize hook was the same in all of them. Now this is put in ``Singleton<T>::get()``.

### DIFF
--- a/common/src/KokkosKernels_Singleton.hpp
+++ b/common/src/KokkosKernels_Singleton.hpp
@@ -1,0 +1,59 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOSKERNELS_SINGLETON_HPP
+#define KOKKOSKERNELS_SINGLETON_HPP
+
+#include <Kokkos_Core.hpp>
+
+namespace KokkosKernels {
+namespace Impl { 
+
+// Singleton structure for a default-constructible type.
+// If initialized, the object will be destructed only during Kokkos::finalize.
+// This is safe to use as a global/static variable type.
+//   This is unlike std::unique_ptr, whose destructor frees the object but doesn't
+//   set its internal pointer to null. This can cause a double-free error when Kokkos::finalize
+//   tries to free the same object later.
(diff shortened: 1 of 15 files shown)


brian-kelley:
@lucbv Yes, the new test replicated the double-free or crash (depending on what TPLs are enabled) like reported on slack. It passes with the singleton changes.

common/src/KokkosKernels_Singleton.hpp
```diff
@@ -0,0 +1,56 @@
...
+struct Singleton {
+  void init() {
+    if (!ptr) {
+      ptr = new T();
+      Kokkos::push_finalize_hook([this]() {
+        delete this->ptr;
+        this->ptr = nullptr;
+      });
+    }
+  }
```
masterleinad, line 41: Any reason you need `init` to be visible outside of this class?

common/src/KokkosKernels_Singleton.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+  operator bool() const { return ptr != nullptr; }
```
masterleinad, line 43: Would it make sense to rather call it something like `is_allocated`?

common/src/KokkosKernels_Singleton.hpp
brian-kelley: No, and actually since it's used only in one place (``get()``) it might as well be inlined there.

brian-kelley:
@lucbv Does this singleton thing deserve a docs page? It does only live in Impl::.

brian-kelley:
@masterleinad I just pushed new changes based on your feedback.

sparse/src/KokkosSparse_spmv.hpp
```diff
@@ -752,7 +752,7 @@ void spmv_struct(const ExecutionSpace& space, const char mode[], const int stenc
                                   typename AMatrix::const_size_type>
       AMatrix_Internal;
 
-  AMatrix_Internal A_i = A;
```
masterleinad, line 755: What happened here?

brian-kelley: Oops, I didn't mean to commit that here. That is #2629 and it was needed to let my local build succeed.

brian-kelley: I just reverted that line.

masterleinad (approved):
Looks good to me!

lucbv:
No need for the documentation to be updated here but maybe this points out a need to have two files, one in src and one in impl directory?

lucbv (approved):

Issue #2600, opened 2025-04-15
Can't build docs locally: no hash for `certifi`


cwpearson:
I suspect this problem was introduced in https://github.com/kokkos/kokkos-kernels/pull/2527

```bash
python -m venv .venv
. .venv/bin/activate

cd docs
pip install -r build_requirements.txt
```

```
...
Collecting tomli>=2
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    tomli>=2 from https://nexus.web.sandia.gov/repository/pypi-proxy/packages/tomli/2.2.1/tomli-2.2.1-py3-none-any.whl#sha256=cb55c73c5f4408779d0cf3eef9f762b9c9f147a77de7b258bef0a5628adc85cc (from Sphinx==7.4.7->-r build_requirements.txt (line 1))
```

lucbv:
Looks to me that the sandia pypi is just not up to date so I am tempted to close this and encourage you to open a ticket with the local admins?
Otherwise you can use the public pipy or build with a more relaxed requirements than those we have for the CI.
That's actually something I have been wondering about, should we have a different "build_requirements.txt" for CI and for users?
The CI is fairly stringent as it does require you to follow a `--require-hashes` recipe...

lucbv:
Note that `pip install -r build_requirements.txt` works just fine in our CI and we want to try to keep that up to date if only to get bug and vulnerability fixes from the python packages...

lucbv:
No response in a couple of weeks, marking as closed.

cwpearson:
Looking further....

It's probably because GA uses python 3.13.2 (at time of writing), and I was using python 3.10. Native TOML support was added in Python 3.11.

I can look into a way to require python 3.11+ so this is a more explicit error.

cwpearson:
Using Python 3.13.2 worked, but I ran into another problem:

```
Collecting certifi>=2017.4.17 (from requests==2.32.3->-r build_requirements.txt (line 49))
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    certifi>=2017.4.17 from https://files.pythonhosted.org/packages/4a/7e/3db2bd1b1f9e95f7cddca6d6e75e2f2bd9f51b1246e546d88addca0106bd/certifi-2025.4.26-py3-none-any.whl (from requests==2.32.3->-r build_requirements.txt (line 49))
```

On GitHub, `certifi` seems to be pre-installed somehow
```
[...]

PR #2632, opened 2025-05-07
Add certifi==2025.4.26 to docs `build_requirements.txt`

cwpearson:
certifi is required by requests==2.32.2, but requests does not provide a hash. This causes our --require-hashes to fail, but we don't catch it on github because certifi is pre-installed there.

Fix for https://github.com/kokkos/kokkos-kernels/issues/2600

### DIFF
--- a/docs/build_requirements.txt
+++ b/docs/build_requirements.txt
@@ -190,3 +190,8 @@ typing_extensions==4.13.2 \
 urllib3==2.4.0 \
   --hash=sha256:414bc6535b787febd7567804cc015fee39daab8ad86268f1310a9250697de466 \
   --hash=sha256:4e16665048960a0900c702d4a66415956a584919c03361cac9f1df5c5dd7e813
+
+# transitive requirement, automatically installed in GA but no hash provided
+# by requests package
+certifi==2025.4.26 \
+  --hash=sha256:30350364dfe371162649852c63336a15c70c6510c2ad5015b21c2345311805f3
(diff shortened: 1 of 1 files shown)


docs/build_requirements.txt
```diff
@@ -190,3 +190,8 @@ typing_extensions==4.13.2 \
 urllib3==2.4.0 \
   --hash=sha256:414bc6535b787febd7567804cc015fee39daab8ad86268f1310a9250697de466 \
   --hash=sha256:4e16665048960a0900c702d4a66415956a584919c03361cac9f1df5c5dd7e813
+
+# transitive requirement, automatically installed in GA but no hash provided
+# by requests package
+certifi==2025.4.26 \
+  --hash=sha256:30350364dfe371162649852c63336a15c70c6510c2ad5015b21c2345311805f3
```
lucbv, line 197:
```suggestion
  --hash=sha256:0a816057ea3cdefcef70270d2c515e4506bbc954f417fa5ade2021213bb8f0c6 \
  --hash=sha256:30350364dfe371162649852c63336a15c70c6510c2ad5015b21c2345311805f3
```

lucbv (changes requested):
Let us add the sha for the source in case a system gets into trouble with binary packages...

docs/build_requirements.txt
cwpearson: Thanks, I didn't know how to do this!

cwpearson:
@lucbv I think you may also have to sign-off on the jointly-authored commit.

lucbv (approved):

PR #2639, opened 2025-05-14
Perftests benchmark cleanup

lucbv:
There are quite a few things happening at the same time in this PR, here is the gist of it:

1. The benchmarks currently under `kokkos-kernels/perf_tests` are migrating to `kokkos-kernels/benchmarks`
2. `KokkosKernels_ENABLE_BENCHMARK` becomes `KokkosKernels_ENABLE_BENCHMARKS` to be more consistent
3. The new benchmarks do not require the perf_tests to be enabled, they only require `KokkosKernels_ENABLE_BENCHMARKS`
4. unless `KokkosKernels_ENABLE_BENCHMARKS=ON` the benchmarks won't run but will show up as Disabled in our tests
5. There is some duplication and separation of perf_tests and benchmark so both can build independently.

Assuming this all looks/sounds good I will enable building and running the benchmarks at OLCF and ALCF and start collecting some data on these systems.


brian-kelley (approved):
LGTM!

jgfouca (approved):

cwpearson (approved):

PR #2647, opened 2025-05-19
CMake: Remove unused export set

dalg24:
Mostly for consistency with recent changes in Kokkos Core. See https://github.com/kokkos/kokkos/pull/8095#pullrequestreview-2850453820


lucbv:
How about waiting until tomorrow morning so we can see if any of the Trilinos/Kokkos builds are failing after merging #8095.

PR #2648, opened 2025-05-20
drop ROCm 5.6.1 PR test configs

cwpearson:
* rename `mi210_rocm621.yaml` -> `mi210.yaml`


lucbv (approved):

lucbv:
Do you know why we see the two mi210 builds in the summary above?

cwpearson:
I am not certain, possibly because some are defined in `develop` and some are defined in this branch? It's also a bit strange that the ones-to-be-revmoed labeled `mi210_rocm621` in this PR are not actually running.

PR #2654, opened 2025-05-26
sparse(static-crs-graph): remove useless initializers

romintomasetti:
Cleaning useless initializers.


sparse/src/KokkosSparse_StaticCrsGraph.hpp
lucbv, line 1: I might be misunderstanding this change but it looks to me that you are doing a few things at once without clearly stating it here:
1. **most important problem** these are public API you are casually deleting, how about you mark them as deprecated instead so when this merges nightly build with Trilinos emit warning instead of taking them all down...
2. All these constructors, operators= are marked as `KOKKOS_INLINE_FUNCTION` so they can be called on device, is that still the case with implicit constructor/destructor.
3. More of the above, if someone currently creates a graph on device they will get a warning when a destructor is called because it won't be marked as `__device__`
4. Why is calling `from` better than using `operator=` or a copy constructor?

lucbv (changes requested):
Please argue why this change is important and why it will have zero impact on users downstream of Kokkos Kernels? Otherwise you need at the very least to introduces deprecations and a rational for asking all customers to change their code.

lucbv (approved):
Okay, let's get this tested.

lucbv:
@romintomasetti would you consider contributing a page of documentation for StaticCrsGraph? We have documentation for both CrsMatrix and BsrMatrix which can be used as example and we provide a template here: [docs/api_template.rst](https://github.com/kokkos/kokkos-kernels/blob/develop/docs/api_template.rst)

romintomasetti:
> @romintomasetti would you consider contributing a page of documentation for StaticCrsGraph? We have documentation for both CrsMatrix and BsrMatrix which can be used as example and we provide a template here: [docs/api_template.rst](https://github.com/kokkos/kokkos-kernels/blob/develop/docs/api_template.rst)

Sorry, not sure I'll find some time for this soon enough...

cwpearson:
I think this only works if Kokkos is compiled with `-DKokkos_ENABLE_DEPRECATED_CODE_4=OFF`. Otherwise we pick up the unmodified implementation from Kokkos.

PR #2658, opened 2025-05-30
Rename remaining reserved identifiers in library

cwpearson:

Haven't checked the tests and examples yet.

Each file can be taken as an atomic change, so each file can be reviewed/considered separately.

### DIFF
--- a/sparse/impl/KokkosSparse_spmv_impl_merge.hpp
+++ b/sparse/impl/KokkosSparse_spmv_impl_merge.hpp
@@ -69,28 +69,28 @@ struct SpmvMergeHierarchical {
 
   template <bool NONZEROS_USE_SCRATCH, bool ROWENDS_USE_SCRATCH, bool Y_USE_SCRATCH, bool CONJ>
   struct SpmvMergeImplFunctor {
-    SpmvMergeImplFunctor(const y_value_type& _alpha, const AMatrix& _A, const XVector& _x, const YVector& _y,
+    SpmvMergeImplFunctor(const y_value_type& alpha, const AMatrix& A, const XVector& x, const YVector& y,
                          const A_size_type pathLengthThreadChunk)
-        : alpha(_alpha), A(_A), x(_x), y(_y), pathLengthThreadChunk_(pathLengthThreadChunk) {}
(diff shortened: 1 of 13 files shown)


cwpearson:
This doesn't change any of the documented CrsMatrix APIs

sparse/impl/KokkosSparse_spmv_impl_merge.hpp
```diff
@@ -69,28 +69,28 @@ struct SpmvMergeHierarchical {
 
   template <bool NONZEROS_USE_SCRATCH, bool ROWENDS_USE_SCRATCH, bool Y_USE_SCRATCH, bool CONJ>
   struct SpmvMergeImplFunctor {
-    SpmvMergeImplFunctor(const y_value_type& _alpha, const AMatrix& _A, const XVector& _x, const YVector& _y,
+    SpmvMergeImplFunctor(const y_value_type& _alpha, const AMatrix& A, const XVector& _x, const YVector& _y,
```
lucbv, line 72: Would have gone with `A_` here if possible

sparse/impl/KokkosSparse_spmv_impl_merge.hpp
```diff
@@ -69,28 +69,28 @@ struct SpmvMergeHierarchical {
...
                          const A_size_type pathLengthThreadChunk)
-        : alpha(_alpha), A(_A), x(_x), y(_y), pathLengthThreadChunk_(pathLengthThreadChunk) {}
+        : alpha(_alpha), A_(A), x(_x), y(_y), pathLengthThreadChunk_(pathLengthThreadChunk) {}
 
     y_value_type alpha;
-    AMatrix A;
+    AMatrix A_;
```
lucbv, line 77: Then this can be kept as `A` which might be easier to work with in the implementation below?

lucbv (commented):
Looked at 8/13 files, will finish this later...

sparse/impl/KokkosSparse_spmv_impl_merge.hpp
cwpearson: Well `A_` is the member name, but I will fix this whole thing to be more consistent in any case.

cwpearson:
We do, for some reason my clang-tidy build is not catching them so I was only finding them in the PR tests. Maybe just a configuration difference.

lucbv (approved):

PR #2659, opened 2025-05-31
Add pbtrf/pbtrs example

yasahi-hpc:
This PR aims at adding a serial `pbtrf/pbtrs` example.
I would like to embed this in the batched documentation. 
See also #2655


lucbv (approved):
Looks good, thanks for adding the example

PR #2671, opened 2025-06-09
Remove stray ;

masterleinad:
I found a bunch of redundant lines of code only containing a semicolon. This pull request deals with all these lines to improve readability.


masterleinad:
> Btw, how did you find them?

Something like
```bash
grep "^[ ]*;\n"
```
There were a bunch of matches for functions that only provide definitions when ETI enabled that I kept.

yasahi-hpc:
> > Btw, how did you find them?
> 
> Something like
> 
> ```shell
> grep "^[ ]*;\n"
> ```
> 
> There were a bunch of matches for functions that only provide definitions when ETI enabled that I kept.

Makes sense. Thanks for the info.

lucbv (approved):

lucbv:
Looking at when the first one was introduced, it seems to be related to clang-format but can't be too sure about it and clearly it is not complaining about removing it in this PR... thanks for the clean-up!

PR #2672, opened 2025-06-10
Add getrf-getrs example

yasahi-hpc:
This PR aims at adding a serial `getrf/getrs` example.
I would like to embed this in the batched documentation. 
See also #2655, #2659, and #2664.


lucbv (approved):
Looks good

PR #2689, opened 2025-06-20
Batched: rename reserved identifiers

cwpearson:
partial fix for https://github.com/kokkos/kokkos-kernels/issues/2364

### DIFF
--- a/batched/sparse/src/KokkosBatched_Krylov_Handle.hpp
+++ b/batched/sparse/src/KokkosBatched_Krylov_Handle.hpp
@@ -80,10 +80,10 @@ class KrylovHandle {
   bool host_synchronised;
 
  public:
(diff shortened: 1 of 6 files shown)


lucbv (approved):
Okay, just a comment for future reference.

batched/sparse/src/KokkosBatched_Krylov_Handle.hpp
```diff
@@ -80,10 +80,10 @@ class KrylovHandle {
   bool host_synchronised;
 
  public:
-  KrylovHandle(int _batched_size, int _N_team, int _max_iteration = 200, bool _monitor_residual = false)
```
lucbv, line 83: Here you could just rename the constructor parameter instead of the struct's attribute, it would require only once change: `_N_team` to `N_team_`

PR #2707, opened 2025-07-22
Build-system: adding support for Apple Accelerate library

lucbv:
Adding option to request the library and modifying the BLAS logic to skip the Fortran check and correctly define the mangling macro for Accelerate. This last part might be an undue overload of the macro purpose?


brian-kelley:
@lucbv Could you rebase on develop?

lucbv:
Rebased

cmake/Modules/FindTPLACCELERATE.cmake
```diff
@@ -0,0 +1,4 @@
+set(BLA_VENDOR Apple)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DACCELERATE_NEW_LAPACK")
+find_package(BLAS REQUIRED)
+kokkoskernels_create_imported_tpl(ACCELERATE INTERFACE LINK_LIBRARIES ${ACCELERATE_LIBRARIES})
```
JBludau: just a drive by, but this file has the wrong ending

brian-kelley (approved):
Aside from @JBludau's comment everything looks good to me!

lucbv:
Okay, I cleaned the CMake logic a bit and as a side effect, the ci build with accelerate is also a bit cleaner now. This should be good to go at this point : )

blas/tpls/KokkosBlas_Host_tpl.cpp
```diff
@@ -317,6 +362,7 @@ void F77_BLAS_MANGLE(cscal, CSCAL)(const KK_INT* N, const std::complex<float>* a
                                    /* */ std::complex<float>* x, const KK_INT* x_inc);
 void F77_BLAS_MANGLE(zscal, ZSCAL)(const KK_INT* N, const std::complex<double>* alpha,
                                    /* */ std::complex<double>* x, const KK_INT* x_inc);
+}  // extern "C"
```
JBludau, line 365: hmm that looks weird since it is touching the #else branch.

lucbv: It is fully contained in the "else" and the branch is based on which library is being used since they are written in different languages...

PR #2713, opened 2025-07-31
Rename type HostMirror -> host_mirror_type

ndellingwood:
Compatibility update following https://github.com/kokkos/kokkos/pull/8232

PR #2721, opened 2025-08-05
Clean up old workaround, update RCB for host_mirror_type renaming

brian-kelley:
- In SortCrs, clean up the workaround that disabled bulk sorting in some cases. This was a workaround for a Kokkos issue that has been fixed since 4.4.
- In RCB impl, rename ``HostMirror`` -> ``host_mirror_type`` to allow building with deprecated code off. This change is just like #2713. The RCB PR was auto-tested just a couple days before HostMirror was deprecated, which is why it was mergeable without this change.


lucbv (approved):

brian-kelley:
I missed a couple more HostMirrors in the unit test, but now I got them all (everything builds/runs cleanly with deprecated off).

vqd8a (approved):

lucbv (approved):

PR #2727, opened 2025-08-11
Rename more reserved identifiers

cwpearson:


example/batched_solve/team_GMRES.cpp
```diff
@@ -176,7 +176,7 @@ int main(int /*argc*/, char ** /*argv*/) {
     using XYType           = Kokkos::View<double **, layout, exec_space>;
 
     std::string name_A = "mat.mm";
-    std::string name_B = "rhs.mm";
```
lucbv, line 179: This one looks a bit odd, especially since `name_A` was not changed in a similar fashion?

cwpearson: actually those are unused. I'll delete them.

lucbv (approved):

PR #2730, opened 2025-08-12
 Minor fix for coefficient type in SPMV_Functor

vqd8a:
The code currently declares```alpha``` and ```beta``` as ```value_type``` of ```AMatrix```, which causes build errors when ```AMatrix``` is real and ```YVector``` is complex. This PR is to fix this issue by switching to use ```value_type``` of ```YVector``` for ```alpha``` and ```beta```


lucbv (approved):
Okay that sounds reasonable, eventually everything will be case into the type of YVector to update it...

PR #2738, opened 2025-08-18
ci: Add V100 CUDA 12.6.2 build with TPLs

cwpearson:
also remove V100 compute sanitizer test (present on H100)


cwpearson:
Looks like with BLAS enabled, the test expects us to throw when we're `(complex && hermitian && !LL)` (this logic has a bit of redundancy in it)

https://github.com/kokkos/kokkos-kernels/blob/838633e723dbd5e5e3255951d5f11b383dd527a9/blas/unit_test/Test_Blas2_ger.hpp#L210-L219

and instead we're observing this

```
[ RUN      ] Cuda.ger_complex_double
/home/runner/_work/kokkos-kernels/kokkos-kernels/kokkos-kernels/blas/unit_test/Test_Blas2_ger.hpp:1167: Failure
Expected equality of these values:
  gotStdException
    Which is: false
  _kkGerShouldThrowException
    Which is: true
Failed test, 'non const {x,y}': kk ger() should have thrown a std::exception
/home/runner/_work/kokkos-kernels/kokkos-kernels/kokkos-kernels/blas/unit_test/Test_Blas2_ger.hpp:1167: Failure
Expected equality of these values:
  gotStdException
    Which is: false
  _kkGerShouldThrowException
    Which is: true
Failed test, 'non const {x,y}': kk ger() should have thrown a std::exception
```

cwpearson:
It looks to me like we do intentionally handle the `'H'` case: both the team and thread parallel GER implementations check the transpose mode and do something or other accordingly

lucbv (approved):

PR #2746, opened 2025-08-26
cusparse spmv_mv: use native fallback if y not 16B aligned

brian-kelley:
This is the same workaround as #1889, but now for the spmv mv (rank-2) case.

Also add a new spmv test that replicates this kind of issue for both rank-1 and rank-2 cases.


brian-kelley:
@lucbv This changes KokkosSparse_spmv.hpp, but it doesn't make any interface changes that would need the documentation to be updated.

lucbv:
Okay the documentation can be left alone, although this begs a bit the question about our documentation of what TPLs we support and when we go down the TPL paths. That would probably be good documentation information to include down the line...

lucbv (approved):

PR #2747, opened 2025-08-27
CI workflows: updating a few workflows to CXX=20

lucbv:
Getting this in our CI ahead of release 5.0 of the ecosystem which will make it a requirement. The compilers for github OSX and Linux actions are recent enough (gcc 13 and clang 15)


cwpearson (approved):

cwpearson (changes requested):
Remove the `spr.yml` workflow file too? I presume that's where the new job in `host.yml` came from?

lucbv:
Correct, I will also unmark "Required" the two jobs that have moved...

.github/workflows/host.yml
```diff
@@ -176,8 +176,8 @@ jobs:
 #        working-directory: kokkos-kernels/build
 #        run: ctest --output-on-failure -V --timeout 3600
 
-  PR_BDW_GNU1020_OPENMP_SERIAL_LEFT_OPENBLAS_REL:
-    name: PR_BDW_GNU1020_OPENMP_SERIAL_LEFT_OPENBLAS_REL
+  PR_GNU1020_OPENMP_SERIAL_LEFT_OPENBLAS_REL:
```
ndellingwood, line 179: Since gcc/10.2.0 isn't supported for c++20 testing should this job just be removed here?

ndellingwood: Or maybe the c++20 option is fine for gcc/10.2, but gcc/10.2 isn't supported for kokkos@5.0

cwpearson (approved):

.github/workflows/host.yml
lucbv: I'm thinking we can keep it until 5.0 comes out then it's easy to turn it off : )

PR #2750, opened 2025-09-03
[Docs] List available batched dense functions in a table

yasahi-hpc:
This PR aims at introducing a table to list available batched dense functions.
If a doc is already available, I added a link to that page. Otherwise, I just put a name of the API.

### DIFF
--- a/docs/source/API/batched/dense-index.rst
+++ b/docs/source/API/batched/dense-index.rst
@@ -1,6 +1,15 @@
 API: Batched Dense (DLA)
 ########################
 
+.. toctree::
+   :maxdepth: 2
+   :hidden:
+
+   dense/batched_pbtrf
+   dense/batched_pbtrs
+   dense/batched_pttrf
+   dense/batched_pttrs
+
 Our implementation of batched dense linear algebra (DLA) allows user to compose various batched DLA operations.
 For example, consider a case where small element matrices are created via `gemm` and those matrices are triangular solved by `lu` and `trsv`.
 An approach using conventional batched BLAS interfaces would be like the following pseudo code:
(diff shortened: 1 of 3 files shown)


docs/source/API/batched/dense-index.rst
```diff
@@ -60,8 +69,297 @@ Kokkos batched APIs provide generic header-only functor-level implementations th
 The APIs can be mapped to Kokkos hierarchical parallelism and also provide various implementations of algorithms with a template argument,
 which allows users to choose (or customize) the batched routines for their application context.
 
-.. toctree::
-   :maxdepth: 1
-   :glob:
+BLAS support
+============
+
+Below are tables summarizing the currently supported function calls in Kokkos Kernels. The tables are updated with each release of the library to reflect recently added support.
```
lucbv:
```suggestion
Below are tables summarizing the currently supported function calls in Kokkos Kernels.
```

docs/source/API/batched/dense-index.rst
```diff
@@ -60,8 +69,297 @@ Kokkos batched APIs provide generic header-only functor-level implementations th
...
+     - `TeamVectorTrmm`
+   * - TRSM
+     - `SerialTrsm`
+     - `TeamTrsm`
+     - `TeamVectorTrsm`
+
+LAPACK support
+==============
+
+Below are tables summarizing the currently supported function calls in Kokkos Kernels. The tables are updated with each release of the library to reflect recently added support.
```
lucbv:
```suggestion
Below are tables summarizing the currently supported function calls in Kokkos Kernels.
```

lucbv (changes requested):
Just a small change, I do not feel that we need to state that we update the documentation here. I believe we have a policy mentioning it already and we have the GitHub action so it is pretty clear that we update the documentation. We should not have to repeat in many places.

yasahi-hpc:
> Just a small change, I do not feel that we need to state that we update the documentation here. I believe we have a policy mentioning it already and we have the GitHub action so it is pretty clear that we update the documentation. We should not have to repeat in many places.

Makes sense. I will remove them. 
These are claims in [Blas](https://github.com/kokkos/kokkos-kernels/blob/7f26b41f2d5a580e6701f57c5f49101adf570fb0/docs/source/API/blas-index.rst?plain=1#L133) and [Lapack](https://github.com/kokkos/kokkos-kernels/blob/7f26b41f2d5a580e6701f57c5f49101adf570fb0/docs/source/API/lapack-index.rst?plain=1#L16) index pages as well.
Should we also remove the statement from these pages?

lucbv:
Yes, we should remove the other mentions as well for BLAS and LAPACK eventually.
I might well be the author of these sentences but I think it is superfluous now so if you want to remove them in this PR go ahead otherwise I can make a PR to take care of that

yasahi-hpc:
I made it in this PR.
Can I have another review

lucbv (approved):

PR #2751, opened 2025-09-03
Avoid using `Kokkos::reduction_identity`

dalg24:
It is intended as a customization point for enabling builtin reducers with user-defined types, and we generally don't want it to appear in user code.

### DIFF
--- a/sparse/impl/KokkosSparse_mdf_impl.hpp
+++ b/sparse/impl/KokkosSparse_mdf_impl.hpp
@@ -144,8 +144,8 @@ struct MDF_discarded_fill_norm {
 
     KOKKOS_INLINE_FUNCTION
     static void init(value_type& val) {
-      val.discarded_norm = Kokkos::reduction_identity<scalar_mag_type>::sum();
-      val.numFillEntries = Kokkos::reduction_identity<ordinal_type>::sum();
+      val.discarded_norm = 0;
+      val.numFillEntries = 0;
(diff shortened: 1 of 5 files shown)


sparse/impl/KokkosSparse_mdf_impl.hpp
```diff
@@ -144,8 +144,8 @@ struct MDF_discarded_fill_norm {
 
     KOKKOS_INLINE_FUNCTION
     static void init(value_type& val) {
-      val.discarded_norm = Kokkos::reduction_identity<scalar_mag_type>::sum();
-      val.numFillEntries = Kokkos::reduction_identity<ordinal_type>::sum();
+      val.discarded_norm = 0;
+      val.numFillEntries = 0;
       val.diag_val       = KAS::zero();
```
dalg24, line 149: Same here I would just assign `0`.

lucbv (approved):

sparse/impl/KokkosSparse_mdf_impl.hpp
dalg24:
```suggestion
      val.diag_val       = 0;
```
is what I meant.  Up to you.

lucbv (approved):
This looks fine to me and I think we can keep the KAS::zero() for now, it usually helps when we get Sacado or Stokhos types passed in

PR #2766, opened 2025-09-11
Update the bisect break condition in RCB

vqd8a:
This PR is to make a minor change for the bisect break condition in RCB to make RCB work with Sierra T/F coordinates.


graph/impl/KokkosGraph_RCB_impl.hpp
```diff
@@ -132,9 +132,12 @@ inline void bisect(const coors_view_type &coors_1d, const value_type &init_min_v
...
   value_type max_val   = init_max_val;
   value_type p1_weight, p2_weight;
-  value_type prev_weight_ratio = 0;
-  value_type curr_weight_ratio;
-  while (1) {
+  value_type weight_ratio;
+  int cnt = 0;
+  // For now, limit the number of times finding mid point to ten to make RCB work with Sierra T/F coordinates
+  // TODO: switch to use the median is probably a better solution
+  while (cnt <= 10) {
```
lucbv: Then let us make it a proper for loop that will avoid some confusion
```suggestion
  for (int bisection_step = 0; bisection_step < max_bisection_steps; ++bisection_step) {
```

graph/impl/KokkosGraph_RCB_impl.hpp
```diff
@@ -132,9 +132,12 @@ inline void bisect(const coors_view_type &coors_1d, const value_type &init_min_v
   value_type min_val   = init_min_val;
   value_type max_val   = init_max_val;
   value_type p1_weight, p2_weight;
-  value_type prev_weight_ratio = 0;
-  value_type curr_weight_ratio;
-  while (1) {
+  value_type weight_ratio;
+  int cnt = 0;
```
lucbv:
```suggestion
const int max_bisection_steps = 11;
```

graph/impl/KokkosGraph_RCB_impl.hpp
```diff
@@ -132,9 +132,12 @@ inline void bisect(const coors_view_type &coors_1d, const value_type &init_min_v
...
+  // For now, limit the number of times finding mid point to ten to make RCB work with Sierra T/F coordinates
+  // TODO: switch to use the median is probably a better solution
+  while (cnt <= 10) {
+    cnt++;
```
lucbv:
```suggestion
```

lucbv (approved):
We can change this later so users can pass `max_bisection_steps` as an input parameter

cwpearson:
We don't require the h100 test as of https://github.com/kokkos/kokkos-kernels/pull/2755, I'm not sure why it's showing up in the list of checks.

graph/impl/KokkosGraph_RCB_impl.hpp
vqd8a: @lucbv Thanks for your suggestion. I have changed it to for loop.

graph/impl/KokkosGraph_RCB_impl.hpp
vqd8a: @lucbv Changed.

graph/impl/KokkosGraph_RCB_impl.hpp
vqd8a: @lucbv Deleted.

lucbv (approved):
Thanks for the updates, do you need this PR to be part of 4.7.01?

lucbv:
> We don't require the h100 test as of #2755, I'm not sure why it's showing up in the list of checks.

I probably need to remove it from the set of required action, which also makes me feel I should update the required actions to include the new Lychee and Kumquat builds.

vqd8a:
> Thanks for the updates, do you need this PR to be part of 4.7.01?

Yes. It would be great to have this in 4.7.01. Is it too late now?

lucbv:
@cwpearson after updating the branch protection settings the old github action does not appear anymore : )

vqd8a:
@ndellingwood Can this PR included in 4.7.01? Thanks.

PR #2771, opened 2025-09-15
Common - ArithTraits: moving from Kokkos to KokkosKernels

lucbv:
This simply moves the class to a new namespace to avoid having it in the Kokkos namespace.


common/src/Kokkos_ArithTraits.hpp
```diff
@@ -1644,9 +1644,15 @@ struct [[deprecated]] ArithTraits<qd_real> {
...
+}  // namespace KokkosKernels
+
+namespace Kokkos {
+template <typename T>
+using ArithTraits [[deprecated("Use KokkosKernels::ArithTraits instead")]] = ::KokkosKernels::ArithTraits<T>;
+
 namespace Details {
 template <typename T>
-using ArithTraits [[deprecated("Use Kokkos::ArithTraits instead")]] = ::Kokkos::ArithTraits<T>;
+using ArithTraits [[deprecated("Use KokkosKernels::ArithTraits instead")]] = ::KokkosKernels::ArithTraits<T>;
```
dalg24: How long ago did you deprecate this alias?
Is removing it an option?

dalg24 (commented):
Any chance to deprecate the header as well and redirect to a new header with appropriate prefix?

common/src/Kokkos_ArithTraits.hpp
lucbv: We are pretty close to 5.0.0, why would we do a backward breaking change against our policies at this point?

dalg24: I am talking about the alias in Details:: that was already marked as deprecated

lucbv: I understand and what I said above still stands, do we really need to be removed before 5.0.0? That's only a few months away...

dalg24: What do you mean "before 5.0"?

lucbv: I don't want to remove this until after the last release in the 4.7 series

dalg24: 4.7 was the last minor release in the 4.x series

lucbv: Okay so not 4.7.2 right? straight to 5.0.0?

dalg24: 4.7.2 would come from 4.7.1 not develop

common/src/Kokkos_ArithTraits.hpp
```diff
@@ -1093,14 +1093,12 @@ class ArithTraits<std::complex<RealFloatType>> {
...
-      return isinf(real(x)) || isinf(imag(x));
+      return isinf(std::real(x)) || isinf(std::imag(x));
     }
   }
 #else
   static bool isInf(const std::complex<RealFloatType>& x) {
-    KOKKOS_IF_ON_HOST((using std::isinf;))
-    KOKKOS_IF_ON_DEVICE((using Kokkos::isinf;))
-    return isinf(real(x)) || isinf(imag(x));
+    return std::isinf(std::real(x)) || std::isinf(std::imag(x));
```
dalg24: Why std::?

dalg24:
> Any chance to deprecate the header as well and redirect to a new header with appropriate prefix?

Would you be willing to move the traits definition to a new `<KokkosKernels_ArithTraits.hpp>`header  and have `< Kokkos_ArithTraits.hpp >` just include it along with a using directive in the Kokkos:: namespace.

common/src/Kokkos_ArithTraits.hpp
lucbv: Honestly I am not sure I understand how the hole thing works on device but if you look further down, `real` and `imag` function are implemented by call `std::real` and `std::imag`.
So for now it's not worse than it was and it allows to complete the scope I defined for this PR.
Of course these functions could easily cause problem down the line and opening an issue to work on it is appropriate.

dalg24 (approved):

ndellingwood:
We'll need to fix Trilinos for these changes, e.g.
```
In file included from /root/Trilinos/packages/stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_MP_Vector.hpp:29:
/root/Trilinos/packages/stokhos/src/sacado/kokkos/vector/linalg/Kokkos_ArithTraits_MP_Vector.hpp:24:7: error: cannot specialize a dependent template
   24 | class ArithTraits< Sacado::MP::Vector<S> > {
```

ndellingwood:
@lucbv just to check, these changes only apply to versions >= 5.0 ? (So I know how to guard in Trilinos)

dalg24:
I suppose you could do
```C++
#if __has_include(KokkosKernels_ArithTraits.hpp)  // since 5.0

lucbv:
Yes this would only apply to versions of Kokkos >= 5.0 so for a 4.7.2 release we would not want this change.

PR #2779, opened 2025-09-18
added check for git submodule checkout

### LINKED ISSUE #2777: Build fails without manual installation of rajaperf
On building from source, the user needs to ensure that the submodule at `tpls/rajaperf` is populated before building, else the option `-DKokkosKernels_ENABLE_TESTS_AND_PERFSUITE=ON` causes build to fail.

The top-level `CMakeLists.txt` does not check if the submodule has been initialized and/or checked out and simply includes the files in the folders `tpls/rajaperf` and `tpls/rajaperf/src`.

prateekchawla168:
Fixes #2777

Added code in [CMakeLists.txt](CMakeLists.txt) to check if the rajaperf submodule has been checked out.

If the compile-time flag -DKokkosKernels_ENABLE_TESTS_AND_PERFSUITE=ON is set and the submodule does not exist, the flag is switched off, and the user is informed to run `git submodule update --init --recursive` to download the submodule first. The compilation is allowed to proceed so the build does not fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,19 @@ IF (NOT KOKKOSKERNELS_HAS_TRILINOS)
     set(BLT_CODE_CHECK_TARGET_NAME "fix-for-blt" CACHE STRING "Docstring")
     set(INFRASTRUCTURE_ONLY ON CACHE BOOL "Only build the RAJAPerf infrastructure, no builtin kernels")
     add_definitions("-DRAJAPERF_INFRASTRUCTURE_ONLY")
+    # Check if this directory is empty,
+    # Populate submodule if possible - Fail gracefully otherwise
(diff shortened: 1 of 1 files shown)


lucbv:
@prateekchawla168 the raja perf tests support was experimental and never got to a stable stage, we should really just remove it actually. Now we are using google perf test and CMake should pull that in automatically for you, we have two associated flags to build and run the tests.

prateekchawla168:
I see, that makes sense. In that case it may be a good idea to remove this flag entirely, else the build still fails in case a user decides to think of using a single flag to activate both tests and the perf suite at the same time.

lucbv:
Sorry for the long delay on this but with PR #2902 I am finally removing the raja perf test dependency so I will close this PR now. Feel free to comment on the other PR and ask any questions.

Issue #2785, opened 2025-09-24
Windows Intel OneAPI compiler build issue with pop_count()/least_set_bit()


hkthorn:
Xyce is encountering a compile-time error trying to build the Trilinos development branch on Windows using Intel oneAPI (Version 2022.1.0) compiler with VS2022 (version 17.4.2).

- In `KokkosKernels_BitUtils.hpp`, the functions `pop_count()` and `least_set_bit()` need to include a definition that will compile with the Windows Intel OneAPI compiler. Since the MSVC definitions are compatible with the OneAPI compiler, we recommend always using the MSVC implementation on Windows. The attached patch file includes the recommended change.

[kokkos_fix_intrinsics_on_windows.patch](https://github.com/user-attachments/files/22471345/kokkos_fix_intrinsics_on_windows.patch)

Side Note:

Selecting the right preprocessor definitions to check for including the appropriate intrinsic function definitions was a bit confusing. There are a mix of references to native compiler flags (i.e. __GNUC__ and __IBMC__) and KOKKOS_COMPILER flags (i.e. KOKKOS_COMPILER_MSVC and KOKKOS_COMPILER_IBM).

Furthermore, there are no conditional definitions for the Intel OneAPI compiler in packages\kokkos-kernels\common\src\KokkosKernels_BitUtils.hpp (no references to either KOKKOS_COMPILER_INTEL_LLVM or __INTEL_LLVM_COMPILER). The pop_count() function includes a definition if __INTEL_COMPILER is defined, but my understanding is that flag is for the Intel Classic Compiler. The OneAPI compiler uses __INTEL__LLVM_COMPILER, and is not guaranteed to set __INTEL_COMPILER.

lucbv:
@hkthorn sorry for not checking with you earlier, we made changes in PR #2786 that hopefully resolved your issue. Let us know if that's not the case?

lucbv:
Since there has not been much feedback in a while I will close this for now, feel free to re-open if the issue is persisting.

PR #2786, opened 2025-09-24
Prefer bit manipulation functions from Kokkos Core

dalg24:
Kokkos provides the same function templates as the C++20 standard library header <bit>.  We propose to defer to these as they are well specified and well tested.

Currently we use the `*_builtin version` that is defined in the `Kokkos::Experimental::` namespace but eventually we should be able to just call the regular unsuffixed function, once the Core library refactors them to defer to the builtins when not constant evaluated (something along the lines of `if consteval { /* constexpr friendly version */ } else { /* builtin fuction */ }`.

One thing to note is that the standard functions only operate on unsigned integers.  I opted for specifying the corresponding unsigned integer type as the argument.


dalg24:
I forgot to mention that we looked into it as a resolution for #2785

lucbv (approved):
Thanks for the clean-up, I will raise the issue to the Kokkos Kernels team and see who wants to have a look at switching to unsigned types for our bit manipulation.

PR #2798, opened 2025-09-30
Require `KOKKOS_ENABLE_DEPRECATED_CODE_5` to be defined for including `<Kokkos_ArithTraits.hpp>`

dalg24:
Error out at compile-time otherwise

### DIFF
--- a/common/src/Kokkos_ArithTraits.hpp
+++ b/common/src/Kokkos_ArithTraits.hpp
@@ -22,6 +22,10 @@
 
 #include <KokkosKernels_ArithTraits.hpp>
 
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_5
+static_assert(false, "this header is deprecated, include <KokkosKernels_ArithTraits.hpp> instead");
+#endif
+
 namespace Kokkos {
 template <typename T>
 using ArithTraits [[deprecated("Use KokkosKernels::ArithTraits from KokkosKernels_ArithTraits.hpp header instead")]] =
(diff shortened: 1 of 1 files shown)


common/src/Kokkos_ArithTraits.hpp
```diff
@@ -22,6 +22,10 @@
 
 #include <KokkosKernels_ArithTraits.hpp>
 
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_5
+static_assert(false, "this header is deprecated, include <KokkosKernels_ArithTraits.hpp> instead");
```
yasahi-hpc:
```suggestion
static_assert(false, "Header `Kokkos_ArithTraits.hpp` is deprecated, include `KokkosKernels_ArithTraits.hpp` instead");
```

I think it is better to explicitly print the name of this header

lucbv (approved):

lucbv:
@ndellingwood assuming we remove the old headers everywhere in Trilinos this should not have an impact on it but just want you to have an eye on this.

yasahi-hpc (changes requested):

yasahi-hpc (approved):
LGTM

ndellingwood:
> @ndellingwood assuming we remove the old headers everywhere in Trilinos this should not have an impact on it but just want you to have an eye on this.

@lucbv thanks I'll keep an eye on it, hopefully I addressed all the header and namespace updates in Trilinos in https://github.com/trilinos/Trilinos/pull/14515 and it shouldn't be an issue (my testing with kokkos-kernels@develop passed with that PR in any case)

ndellingwood (approved):

PR #2800, opened 2025-09-30
Avoid opening the Kokkos namespace

dalg24:
This is a nitpick.  Qualify the reduction_identity specialization instead of opening the Kokkos namespace.

With that change, one can grep for "namespace Kokkos {" to find all naughty code.


lucbv (approved):
That's fine with me but the change makes it harder to detect where naughty things happen? No more `namespace Kokkos {` for you to spot?

PR #2802, opened 2025-10-02
Define Fortran interface macros for LAPACK

brian-kelley:
If LAPACK TPL is enabled but BLAS and MKL are not, we would get build errors like:
```
/home/bmkelle/KK_AllowLapackOnly/kokkos-kernels/lapack/tpls/KokkosLapack_Host_tpl.cpp:19:27: error: expected ‘)’ before ‘,’ token
   19 | void F77_BLAS_MANGLE(sgesv, SGESV)(int*, int*, float*, int*, int*, float*, int*, int*);
```
This PR will ensure the F77_BLAS_MANGLE macro is defined if LAPACK is enabled. Built+tested successfully for serial backend and ``-DKokkosKernels_ENABLE_TPL_LAPACK=ON`` only (no BLAS).

LAPACK without BLAS is probably not a common configuration, but a real user that was installing via Spack ran into this.

Side note: for spack, I opened PR https://github.com/spack/spack-packages/pull/1838 that automatically adds ``+blas`` if ``+lapack`` is enabled. This fix is nicer to apply for Spack because it works for all KokkosKernels versions, but it doesn't conflict with the changes here.


lucbv (approved):
Yeah, I guess so

PR #2809, opened 2025-10-07
Add a member function to BsrMatrix to convert to Crs

jgfouca:
Also, add a test.

Also, fixes some problems with the existing Bsr test. Some checks were being lost due to not `&=` into `d_results` inside a row loop.

### DIFF
--- a/sparse/src/KokkosSparse_BsrMatrix.hpp
+++ b/sparse/src/KokkosSparse_BsrMatrix.hpp
@@ -956,7 +956,9 @@ class BsrMatrix {
     crs_values_t crsValues("crsValues", numCrsEntries);
 
     // Create the policy, we have 3 levels of parallelism available in the algorithm
-    policy_t policy(numBlockRows, Kokkos::AUTO(), blockDim);
+    const crs_size_t maxvec = policy_t::vector_length_max();
+    const crs_size_t veclen = (blockDim <= maxvec) ? blockDim : maxvec;
+    policy_t policy(numBlockRows, Kokkos::AUTO(), veclen);
 
     // Fill CrsMatrix row map, entries, and values
     Kokkos::parallel_for(
--- a/sparse/src/KokkosSparse_BsrMatrix.hpp
+++ b/sparse/src/KokkosSparse_BsrMatrix.hpp
@@ -912,6 +912,94 @@ class BsrMatrix {
     return const_block_type(&values(i * blockDim_ * blockDim_), blockDim_, blockDim_);
   }
 
+  /// \brief Convert the Bsr into a CrsMatrix
+  ///
+  /// The default return type will be a CrsMatrix with all the same template arguments
+  /// as this Bsr, but you can provide your own type if needed. The only requirement
+  /// is that the execution spaces match.
+  ///
+  /// This is a host function.
+  ///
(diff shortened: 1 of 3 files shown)


lucbv (approved):

sparse/src/KokkosSparse_BsrMatrix.hpp
```diff
@@ -912,6 +912,94 @@ class BsrMatrix {
...
+    const crs_size_t numCrsCols    = numBlockCols * blockDim;
+    const crs_size_t numCrsEntries = numBlockEntries * blockSize;
+
+    // Allocate CrsMatrix views
+    crs_rowmap_t crsRowMap("crsRowMap", numCrsRows + 1);
+    crs_entries_t crsEntries("crsEntries", numCrsEntries);
+    crs_values_t crsValues("crsValues", numCrsEntries);
+
+    // Create the policy, we have 3 levels of parallelism available in the algorithm
+    policy_t policy(numBlockRows, Kokkos::AUTO(), blockDim);
```
brian-kelley: The vector length should be limited to ``policy_t::vector_length_max()``

brian-kelley (changes requested):
All looks good except there is one case that the team policy construction might fail

jgfouca:
@lucbv , @ndellingwood , the failure on lychee:

```
/home/runner/_work/kokkos-kernels/kokkos-kernels/kokkos-kernels/test_common/KokkosKernels_TestUtils.hpp:127: Failure
Expected: ((double)AT1::abs(val1 - val2)) <= ((double)AT3::abs(tol)), actual: 2.53593 vs 2.11288
row: 275, crs_col_ids_ref(25461) = 13 mismatched values!
Begin arguments for above failure...
RandCooMat<f, N6Kokkos10LayoutLeftE, N6Kokkos4CudaE421...): rand seed: 3769690815
scalar: f
layout: N6Kokkos10LayoutLeftE
m: 421, n: 421
...end arguments for above failure.

[  FAILED  ] Cuda.sparse_coo2crs (40174 ms)
```

Does not look like it's related to my PR. Is this a known issue?

sparse/src/KokkosSparse_BsrMatrix.hpp
jgfouca: @brian-kelley , so should it be:
`blockDim <= vector_length_max ? blockDim : vector_length_max` ?

brian-kelley: Yep

jgfouca: @brian-kelley , done

brian-kelley (approved):
Thanks!

brian-kelley:
@jgfouca I agree that ``sparse_coo2crs`` failing is unrelated to what you're changing here. Don't know of any issues (open or closed) about it on Cuda though. There used to be one about Sycl.

jgfouca:
Well, it looks like the veclen change fixed the test. That, or the fail is random.

brian-kelley:
@jgfouca I just found out that Kokkos treats the vector length as a hint only. So this didn't actually change the behavior in your kernel (the ThreadVectorRange will still iterate over the same range).

I think the sparse_coo2crs failing must be random.

PR #2825, opened 2025-10-15
[Docs] batched serial lacgv

yasahi-hpc:
This PR aims at introducing batched docs for lacgv.

- [x] docs for lacgv

### DIFF
--- a/docs/source/API/batched/dense/batched_lacgv.rst
+++ b/docs/source/API/batched/dense/batched_lacgv.rst
@@ -1,0 +1,35 @@
+KokkosBatched::Lacgv
+####################
+
+Defined in header: :code:`KokkosBatched_Lacgv.hpp`
+
+.. code:: c++
+
+    struct SerialLacgv {
+      template <typename XViewType>
+      KOKKOS_INLINE_FUNCTION static int invoke(const XViewType &x);
+    };
+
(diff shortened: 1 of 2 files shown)


docs/source/API/batched/dense/batched_lacgv.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Parameters
+==========
+
+:x: :math:`x` is a length n vector.
+
+Type Requirements
+-----------------
+
+- ``XViewType`` must be a Kokkos `View <https://kokkos.org/kokkos-core-wiki/API/core/view/view.html>`_ of rank 1 containing a vector :math:`x`
```
lucbv:
```suggestion
- ``XViewType`` must be a Kokkos `View <https://kokkos.org/kokkos-core-wiki/API/core/view/view.html>`_ of rank 1 representing the vector :math:`x`
```

lucbv (changes requested):
It looks fine but I would rephrase a little bit, the Kokkos View contains the values associated with the components of the vector. It technically does not contain a `std::vector` or something like that.

docs/source/API/batched/dense/batched_lacgv.rst
yasahi-hpc: Makes sense. I have updated accordingly

lucbv (approved):

PR #2830, opened 2025-10-20
Add batched serial trsv documentation

yasahi-hpc:
This PR aims at introducing batched docs for trsv.

- [x] docs for trsv


lucbv (approved):
Good, thanks for the documentation

PR #2844, opened 2025-10-30
CI - gcc/1050: adding new action with c++20

lucbv:
This updates the image we are running on to a new one with gcc/10.5 which allows us to turn on c++20 with Kokkos Core : ) Also cleaning up a few things as well and will likely add some compiler warnings here as we are only testing with them on in our osx CI...

### DIFF
--- a/.github/workflows/bdw.yml
+++ b/.github/workflows/bdw.yml
@@ -252,3 +252,81 @@ jobs:
       - name: test
         working-directory: kokkos-kernels/build
         run: ctest --output-on-failure -V --timeout 3600
+
+  PR_BDW_GNU1050_OPENMP_SERIAL_LEFT_OPENBLAS_REL:
+    name: PR_BDW_GNU1050_OPENMP_SERIAL_LEFT_OPENBLAS_REL
+    runs-on: [kk-env-openblas-0.3.21-gcc-10.2.0-latest]
+    
+    steps:
+      - name: checkout_kokkos_kernels
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          path: kokkos-kernels
+
+      - name: checkout_kokkos
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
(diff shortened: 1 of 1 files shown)


cwpearson (approved):

.github/workflows/bdw.yml
```diff
@@ -252,3 +252,84 @@ jobs:
...
+            -DBUILD_SHARED_LIBS=OFF \
+            -DCMAKE_CXX_COMPILER=$(spack location -i gcc@10.5.0)/bin/g++ \
+            -DCMAKE_CXX_FLAGS=-O3 \
+            -DCMAKE_INSTALL_PREFIX=$PWD/../install \
+            -DCMAKE_VERBOSE_MAKEFILE=ON \
+            -DCMAKE_CXX_EXTENSIONS=OFF \
+            -DCMAKE_CXX_STANDARD=20 \
+            -DKokkos_ENABLE_SERIAL=ON \
+            -DKokkos_ENABLE_OPENMP=ON \
+            -DKokkos_ARCH_NATIVE=ON \
```
ndellingwood, line 290: We might want to rename the build, looking at the kokkos configure using Kokkos_ARCH_NATIVE enabled AVX512 which I think maps to SKX rather than BDW?

```
-- SIMD: AVX512 detected
-- Built-in Execution Spaces:
--     Device Parallel: NoTypeDefined
--     Host Parallel: Kokkos::OpenMP
--       Host Serial: SERIAL
-- 
-- Architectures:
--  NATIVE
```

ndellingwood: Re: naming update, I'm referring to "bdw" prefix in the GA link/label, 
[bdw / PR_GNU1050_OPENMP_SERIAL_LEFT_OPENBLAS_REL](https://github.com/kokkos/kokkos-kernels/actions/runs/18950015736/job/54111879599?pr=2844#logs)

cwpearson: `bdw` comes from this job being defined in `bdw.yml`. Could move it to a different / new yaml file accordingly.

lucbv: That's true, I'll do that in the follow up PR #2747 which also adds more c++20...

PR #2846, opened 2025-11-04
Improve performance of par_ilut and improve benchmark

jgfouca:
Main change: threshold_select was doing a small piece of par_ilut on host. I assumed based on the code and ginkgo's OMP implementation that the computational cost of this piece would be trivial, but it was actually very expensive. Instead of relying on std::nth_element, just make a copy of values and sort based on the absolute value. This requires a bit more device memory, but is 20 times faster, so I think this is well worth it.

Secondary change: various improvements to the par_ilut benchmark. The main improvements are the ability to validate the ILU via gmres+luprec and efforts to reduce memory usage so you can benchmark larger matrices.

### DIFF
--- a/benchmarks/sparse/KokkosSparse_par_ilut.cpp
+++ b/benchmarks/sparse/KokkosSparse_par_ilut.cpp
@@ -2,7 +2,7 @@
 // SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
 
 #include <cstdio>
-
+#include <chrono>
 #include <ctime>
 #include <cstring>
 #include <cstdlib>
@@ -15,6 +15,8 @@
 
 #include "KokkosSparse_Utils.hpp"
 #include "KokkosSparse_spiluk.hpp"
+#include "KokkosSparse_gmres.hpp"
+#include "KokkosSparse_LUPrec.hpp"
 #include "KokkosSparse_par_ilut.hpp"
 #include "KokkosSparse_spmv.hpp"
 #include "KokkosBlas1_nrm2.hpp"
@@ -33,6 +35,7 @@
 
 namespace {
 
+using KokkosSparse::Experimental::gmres;
 using KokkosSparse::Experimental::par_ilut_numeric;
 using KokkosSparse::Experimental::par_ilut_symbolic;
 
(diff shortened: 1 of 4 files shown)


jgfouca:
Some perf data from lychee:
```
General info:

num iters no prec: 765

=================== KK::PAR_ILUT ===================

LUPrec results: num iters 345 (took 123.082 seconds)

KokkosSparse_par_ilut_par_ilut/rows:4173281/real_time       24.5 s          24.1 s             1

=================== GINKGO ========================

LUPrec results: num iters 429 (took 167.163 seconds)

KokkosSparse_par_ilut_gingko/rows:4173281/real_time       7.47 s          7.37 s             1

=================== KK::SPILUK ========================

LUPrec results: num iters 96 (took 86.1897 seconds)

KokkosSparse_par_ilut_spiluk_symbolic/rows:4173281/real_time       69.3 s          69.1 s             1
KokkosSparse_par_ilut_spiluk_numeric/rows:4173281/real_time        36.2 s          36.1 s             1

=================== KK::PAR_ILUT W/ THRESHOLD OPT ===================

LUPrec results: same as without the optimization

KokkosSparse_par_ilut_par_ilut/rows:4173281/real_time       5.67 s          5.66 s             1
```

sparse/impl/KokkosSparse_par_ilut_numeric_impl.hpp
```diff
@@ -16,12 +16,26 @@
...
 #include <limits>
 
 namespace KokkosSparse {
 namespace Impl {
 namespace Experimental {
 
+// Resize a view without allocating or initializing if possible.
+template <typename ViewT>
+inline void resize_no_preserve(ViewT& view, const size_t new_size) {
+  // const auto curr_size = view.size();
```
lucbv: Looks like you are only calling realloc at this point, should this function be removed?

lucbv (commented):
Mostly fine but I think the resize function is really not required at this point unless you want to get back to it in a follow up PR?

sparse/impl/KokkosSparse_par_ilut_numeric_impl.hpp
jgfouca: Ah, thanks for reminding me; I was going to make a note of this here. The functionality I'm looking for here is what std::vector offers when you resize to a smaller size, it doesn't do an allocation. The vector memory and capacity stays the same, it's just the size that changes. I had a hard time figuring out of Kokkos::realloc supports this.

lucbv: That's something you can ask in the general channel in slack, there is a chance this is just not support.

jgfouca: @lucbv , I removed the function. The timers show that the realloc cost is trivial.

benchmarks/sparse/KokkosSparse_par_ilut.cpp
```diff
@@ -41,24 +44,147 @@ using KokkosSparse::spiluk_symbolic;
...
-using EntriesType = Kokkos::View<lno_t*, device>;
-using ValuesType  = Kokkos::View<scalar_t*, device>;
-
-using sp_matrix_type = KokkosSparse::CrsMatrix<scalar_t, lno_t, device, void, size_type>;
-using KernelHandle =
-    KokkosKernels::Experimental::KokkosKernelsHandle<size_type, lno_t, scalar_t, exe_space, mem_space, mem_space>;
-using float_t = typename KokkosKernels::ArithTraits<scalar_t>::mag_type;
+using scalar_t    = KokkosKernels::default_scalar;
+using lno_t       = KokkosKernels::default_lno_t;
+using size_type_t = lno_t;  // ginkgo does not have the concept of size_type separate from ordinal type
```
vqd8a, line 49: I am fine with ```size_type_t```. Is adding ```_t``` to agree with other type names? I just feel it is a bit redundant as ```size_type``` already means a type.

sparse/impl/KokkosSparse_par_ilut_numeric_impl.hpp
```diff
@@ -71,8 +73,8 @@ struct IlutWrap {
                                   LU_row_map);
 
     const size_type lu_nnz_size = kh.get_spgemm_handle()->get_c_nnz();
-    Kokkos::resize(LU_entries, lu_nnz_size);
-    Kokkos::resize(LU_values, lu_nnz_size);
+    Kokkos::realloc(Kokkos::WithoutInitializing, LU_entries, lu_nnz_size);
```
vqd8a, line 76: Just for my understanding. Why do you need to use ```realloc```? ```Kokkos::WithoutInitializing``` can also be used with ```resize```, can't it? Or was ```resize``` deprecated?

vqd8a (approved):
@jgfouca This looks good to me. I just have two minor comments.

benchmarks/sparse/KokkosSparse_par_ilut.cpp
jgfouca: Just for consistency, I like all typenames to have a trailing `_t`.

sparse/impl/KokkosSparse_par_ilut_numeric_impl.hpp
jgfouca: Resize preserves the original content of the view. We are going to overwrite all the contents, so there's no need to preserve which saves a bit of copying.

vqd8a: I missed that. Thanks for explaining!

PR #2851, opened 2025-11-10
Avoid calls to potentially non-static member function `::rank` in the case of `Kokkos::DynRankView` types

### LINKED ISSUE #2850: `Kokkos::DynRankView` type can no longer be used with `KokkosBatched::Algo::LU::Blocked::Functor_BatchedTeamLU`
If one attempts to use a `Kokkos::DynRankView` in `KokkosBatched::Algo::LU::Blocked::Functor_BatchedTeamLU`, one encounters compilation errors like:

```
include/KokkosBatched_Trsm_Team_Impl.hpp:62:30: error: call to non-static member function without an object argument
   62 |     static_assert(AViewType::rank() == 2);
```

MalachiTimothyPhillips:
Closes https://github.com/kokkos/kokkos-kernels/issues/2850

@lucbv


lucbv (changes requested):
Bad approach in my opinion, rather than removing checks we want to add more compile time verification to catch potential mistakes. See my comment on the issue

MalachiTimothyPhillips:
> Bad approach in my opinion, rather than removing checks we want to add more compile time verification to catch potential mistakes. See my comment on the issue

Noted -- I'll opt to close this and defer the issue to you. Thanks!

PR #2853, opened 2025-11-11
Fix batched assertions

lucbv:
@MalachiTimothyPhillips 
Fixing the logic in Trsm to allow both View and DynRankView as input parameters while still trying to catch obvious issues at compile time.


ndellingwood (approved):
Thanks @lucbv

ndellingwood:
Cherry-pick to 5.0 https://github.com/kokkos/kokkos-kernels/pull/2854

PR #2855, opened 2025-11-11
[5.0] cherry-pick #2846: Improve performance of par_ilut and improve benchmark 

ndellingwood:
* Improve performance of par_ilut and improve benchmark

Main change: threshold_select was doing a small piece of par_ilut on host. I assumed based on the code and ginkgo's OMP implementation that the computational cost of this piece would be trivial, but it was actually very expensive. Instead of relying on std::nth_element, just make a copy of values and sort based on the absolute value. This requires a bit more device memory, but is 20 times faster, so I think this is well worth it.

Secondary change: various improvements to the par_ilut benchmark. The main improvements are the ability to validate the ILU via gmres+luprec and efforts to reduce memory usage so you can benchmark larger matrices.

* Fix complex

* Remove resize_no_preserve, realloc works fine

---------

### DIFF
--- a/benchmarks/sparse/KokkosSparse_par_ilut.cpp
+++ b/benchmarks/sparse/KokkosSparse_par_ilut.cpp
@@ -2,7 +2,7 @@
 // SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
 
 #include <cstdio>
-
+#include <chrono>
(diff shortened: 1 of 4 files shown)


jgfouca (approved):
Thanks!

PR #2859, opened 2025-11-13
More config options for par_ilut benchmark

jgfouca:
Adds ability to select:
* max_iters
* fill_in_limit
* residual_norm_delta_stop
* gmres_max_subspace

Also, fixes a huge performance regression on CPU. The new threshold_select implementation worked well on GPU but was terrible on CPU. I have added some code so that the old implementation is used on CPU.

@ndellingwood , can you cherry-pick this? I don't think we want the slow code being released.


lucbv (approved):
Looks fine to me

ndellingwood:
5.0 cherry-pick #2860

PR #2861, opened 2025-11-17
Cleanup KK warning system

jgfouca:
Change list:
1) Bring over similar warning system from main kokkos, except encapsulate it into a macro `kokkoskernels_enable_warnings` which was added to new file `kokkoskernels_util.cmake`
2) Formatting improvements, especially in regards to remove all-caps, in a couple cmake files
3) Fix warnings all over the place
4) Add new KokkosKernels options `KokkosKernels_ENABLE_COMPILER_WARNINGS` and  `KokkosKernels_ENABLE_WARNINGS_AS_ERRORS`.
5) Change CI to use these options instead of hardcoding cxx flags.


CMakeLists.txt
```diff
@@ -1,170 +1,190 @@
...
+  BOOL
+  "Whether to build docs. Default: OFF"
+)
 
-KOKKOSKERNELS_ADD_OPTION(
-        "ENABLE_DOCS"
-        OFF
-        BOOL
-        "Whether to build docs. Default: OFF"
+kokkoskernels_add_option(
```
jgfouca, line 113: Real new code is here.

batched/dense/src/KokkosBatched_Vector_SIMD.hpp
```diff
@@ -76,7 +76,20 @@ class Vector<SIMD<T>, l> {
 
   KOKKOS_INLINE_FUNCTION
   value_type &operator[](const int &i) const { return _data[i]; }
+
+  KOKKOS_INLINE_FUNCTION
```
jgfouca: These changes are worth a close review as they potentially change code behavior.

lucbv:
I have some general comments:

- If you want this do be a reasonably fast PR I would only add the current set of warnings listed in our osx build: `-Wall -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wignored-qualifiers -Wempty-body -Wuninitialized`. Other subsequent PRs can propose to add more warnings later on
- The switch from upper to lower case in CMakeLists.txt is arguably an improvement but it also makes it fairly hard to know what changed in this PR, I would suggest to make that change separately, maybe after consulting everyone in the weekly meeting?
- There are still some issues in the OSX build with the new warning and there are new warnings emitted in the CUDA/HIP builds but we are not using `-Werror` there?

jgfouca:
@lucbv ,

> If you want this do be a reasonably fast PR I would only add the current set of warnings listed in our osx build: -Wall -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wignored-qualifiers -Wempty-body -Wuninitialized. Other subsequent PRs can propose to add more warnings later on

I'm in no hurry, so we can take the time to do this the way we want it long term. I went through and fixed the new warnings that were discovered in our code. Most of them were extra semicolons which were easy to fix. The only tricky one was "deprecated copy", which required some changes to our Vector class. I think there's probably some value in having the same warnings on as Kokkos core since we are in that ecosystem.

> The switch from upper to lower case in CMakeLists.txt is arguably an improvement but it also makes it fairly hard to know what changed in this PR, I would suggest to make that change separately, maybe after consulting everyone in the weekly meeting?

I agree. I tried to flag the "real" change via gh comment. The only real change to the main CMakeLists.txt file is this:
```
kokkoskernels_add_option(
  "ENABLE_COMPILER_WARNINGS"
  OFF
  BOOL
  "Whether to turn on more warnings. Default: OFF"
)

kokkoskernels_add_option(
  "ENABLE_WARNINGS_AS_ERRORS"
  OFF
  BOOL
  "Whether warnings should be treated as errors. Default: OFF"
)

include(cmake/kokkoskernels_util.cmake)

... Later, in the KK standalone build section:

    if (KokkosKernels_ENABLE_COMPILER_WARNINGS)
      kokkoskernels_enable_warnings()
    endif()
```

I also added a macro for turning on warnings (code that was copied with minor mods from Kokkos) in the new file `kokkoskernels_util.cmake`.

> There are still some issues in the OSX build with the new warning and there are new warnings emitted in the CUDA/HIP builds but we are not using -Werror there?

I will clean these issues up.

blas/impl/KokkosBlas1_reciprocal_impl.hpp
```diff
@@ -14,6 +14,12 @@ namespace Impl {
 // reciprocal
 //
 
+#if defined(__GNUC__) && ((__GNUC__ == 12) || (__GNUC__ == 13)) && !defined(__NVCC__)
```
jgfouca, line 17: This is a change worth reviewing as well.

blas/impl/KokkosBlas1_reciprocal_impl.hpp
```diff
@@ -14,6 +14,12 @@ namespace Impl {
...
+#define NEED_TO_DEOPT 1
```
ndellingwood: might need to update the macro naming here, maybe something like `KOKKOSKERNELS_GCC_BUG_DISABLE_TREE_VECTORIZE` to add more detail?

jgfouca: Done.

batched/dense/src/KokkosBatched_Vector_SIMD.hpp
```diff
@@ -76,7 +76,20 @@ class Vector<SIMD<T>, l> {
...
+  type &operator=(const type &rhs) {
```
lucbv: Why not allow the same type of input as the copy constructor?

```
template <typename ArgValueType>
  KOKKOS_INLINE_FUNCTION
    type &operator=(const Vector<SIMD<ArgValueType>, vector_length> &rhs) {
    KOKKOSKERNELS_FORCE_SIMD
    for (int i = 0; i < vector_length; ++i) _data[i] = rhs._data[i];
  }
```

batched/dense/src/KokkosBatched_Vector_SIMD.hpp
```diff
@@ -76,7 +76,20 @@ class Vector<SIMD<T>, l> {
...
+    for (int i = 0; i < vector_length; ++i) _data[i] = rhs._data[i];
+    return *this;
+  }
+
+  KOKKOS_INLINE_FUNCTION Vector(const type &b) {
+    KOKKOSKERNELS_FORCE_SIMD
+    for (int i = 0; i < vector_length; ++i) _data[i] = b[i];
+  }
 };
+
```
lucbv, line 103: Don't you need a destructor for `Vector`?

cmake/kokkoskernels_warnings.cmake
lucbv, line 1: I do not like the name of this file, `util` is the least descriptive name and will lead to others lazily adding anything to it until we have no clue what this file is intended to provide.

CMakeLists.txt
```diff
@@ -1,170 +1,190 @@
...
+    if (${Kokkos_VERSION} VERSION_GREATER_EQUAL "4.6.02")
+      message(STATUS "Found Kokkos version ${Kokkos_VERSION} at ${Kokkos_DIR}")
+      if (${Kokkos_VERSION} VERSION_GREATER "4.7.99")
+        message(WARNING "Configuring with Kokkos ${Kokkos_VERSION} which is newer than the expected develop branch - version check may need update")
+      endif()
+    else()
+      message(FATAL_ERROR "Kokkos Kernels ${KokkosKernels_VERSION} requires Kokkos 4.6.02 or greater (found ${Kokkos_VERSION})")
+    endif()
+
+    if (KokkosKernels_ENABLE_COMPILER_WARNINGS)
```
lucbv, line 175: I might be wrong but aren't you supposed to write cmake logic using the cache variable instead of the user passed parameter?

```suggestion
    if (KOKKOSKERNELS_ENABLE_COMPILER_WARNINGS)
```

Note that I could be wrong here, so please do not just make the change, it would be good to understand what is the proper way to do this

lucbv (changes requested):
Some questions and some changes required but nothing major

CMakeLists.txt
jgfouca: `KokkosKernels_XXX` is the CACHE variable:

```
  SET(CAMEL_NAME KokkosKernels_${SUFFIX})
  STRING(TOUPPER ${CAMEL_NAME} UC_NAME)

  # Make sure this appears in the cache with the appropriate DOCSTRING                                                                                                                                                          
  SET(${CAMEL_NAME} ${DEFAULT} CACHE ${TYPE} ${DOCSTRING})
```

There's a lot of stuff in `kokkoskernels_add_option` that doesn't make sense to me and is not standard CMake. I assume this has something to do with tribits. It looks like we make sure that the cache is always camel case but then set non-cache variables in all caps. I assume this is to allow case-insensitive cmake code. In other words, both

`if (KokkosKernels_XXX)` and `if (KOKKOSKERNELS_XXX)` will work. I think it's better to just use the CACHE variable (which is also what the user would pass).

lucbv:
@jgfouca the last PR I merged created a small conflict, can you have a look, I'm happy to review again after

jgfouca:
@lucbv , done

lucbv (approved):

PR #2874, opened 2025-12-01
Format cmake to lowercase

jgfouca:
Continuing the little bit of CMake cleanup I did in a previous PR, this PR expands that cleanup to all our cmake files. I used cmake-format to help.

The main goals were to change commands, functions, and macros to lowercase and to fix indentation. I also aimed for consistency when splitting longer argument lists across multiple lines.

Testing:
1) Do a configure for the common ancestor commit
2) Do a configure for this branch
3) Do a recursive case-insensitive `diff` between both directories. This takes a bit of work as there are a lot of false positives (differences that are inconsequential).
 3.a) To reduce false positives, all lines are normalized before being comapared. The normalization strips whitespace, lowercases everything, and changing all integers to 0.
 3.b) Only a couple files will have post-normalized diffs. I used tkdiff to examine these and determined that the differences would not impact anything.

I did the above for a pedantic build on CPU and GPU.


batched/CMakeLists.txt
```diff
@@ -1,79 +1,45 @@
...
-  SOURCE_LIST SOURCES
-  TYPE_LISTS  FLOATS RIGHT_LAYOUTS DEVICES
-)
-KOKKOSKERNELS_GENERATE_ETI(Batched_Gemm_t_t_bll Gemm
-  COMPONENTS  batched
-  HEADER_LIST ETI_HEADERS
-  SOURCE_LIST SOURCES
-  TYPE_LISTS  FLOATS RIGHT_LAYOUTS DEVICES
-)
+macro(gen_batched_eti FUNC_NAME)
```
jgfouca, line 30: There are a couple places, like here, where I used macros to reduce a lot of code duplication.

blas/CMakeLists.txt
```diff
@@ -59,268 +52,50 @@ ENDIF()
...
-  TYPE_LISTS  FLOATS LAYOUTS DEVICES
-)
 
-KOKKOSKERNELS_GENERATE_ETI(Blas3_trmm trmm
-  COMPONENTS  blas
-  HEADER_LIST ETI_HEADERS
-  SOURCE_LIST SOURCES
-  TYPE_LISTS  FLOATS LAYOUTS DEVICES
-)
+macro(gen_blas_eti FUNC_NAME SUBDIR)
```
jgfouca, line 56: And here.

batched/CMakeLists.txt
jgfouca: I found it odd that `bll` was for RIGHT_LAYOUTS and `blr` was for LEFT_LAYOUTS, but that's the original behavior, so I kept it.

jgfouca: I should note, fixing this actually causes the build to fail, which is a bit concerning. Something is depending on `bll` being layout right or vice versa.

sparse/CMakeLists.txt
```diff
@@ -1,189 +1,57 @@
...
 
+macro(gen_sparse_eti FUNC_NAME SUBDIR)
+  kokkoskernels_generate_eti(${FUNC_NAME} ${SUBDIR}
+    COMPONENTS  sparse
+    HEADER_LIST ETI_HEADERS
+    SOURCE_LIST SOURCES
+    TYPE_LISTS  FLOATS ORDINALS OFFSETS LAYOUTS DEVICES)
+endmacro()
+
 #Build up a list of DECL, AVAIL, and INST macros
```
jgfouca, line 24: More ETI cleanup here as well.

lucbv:
@jgfouca did I forget to mention that Kokkos Core already has a cmake-format and uses an action to enforce it?
I thought I did and looking at all these changes, I wish we could have used their approach since it it likely faster but also prevents regression. Please have a look [here](https://github.com/kokkos/kokkos/blob/develop/.github/workflows/cmake-format-check.yml) for the action and [here](https://github.com/kokkos/kokkos/blob/develop/.cmake-format.py) for the format definition. More info on the cmake-format and cmake-lint are available [here](https://cmake-format.readthedocs.io/en/latest/index.html)

jgfouca:
@lucbv , yes, you did mention it and I used `cmake-format` to do a lot of these changes. The problem is that the only thing I am pretty sure the whole team agrees on is that we shouldn't have CAPS for everything and we should have consistent indentation. When I applied `cmake-format` using the kokkos core settings, I saw of lot changes that did not look like improvements and I struggled to come up with a format configuration that didn't do a bunch of line splits that I didn't want, so I did a manual post-processing step of cleaning-up line splits.

It's up to the team to decide what to do next... In my opinion, our cmake now looks pretty nice and is consistent, so I'm fine moving forward with just an informal commitment to keeping new cmake stylistically similar to what we have now. I think most developers use copy/paste for cmake, so it could be fine? If folks want something more formal like a cmake-format.py and a CI job to enforce it, that's fine with me too, but then we'll need to agree on the contents of cmake-format.py and there appear to be some limitations to the tool when it comes to telling it not to do stuff.

lucbv (approved):
Fine with me although I might revisit the automated cmake-format idea using pre-commit for the github action

jgfouca:
@lucbv , thanks. I think the CMake in it's current state (in this PR) should be a good starting point for whatever automated stuff we want to add.

PR #2902, opened 2026-01-07
PerfTests: removing raja perf tests

lucbv:
This effort was never fully completed and it is now left un-maintained. Let us remove it and focus on the solution we have adopted with google benchmarks instead.


jgfouca (approved):
lgtm

ndellingwood (approved):

PR #2908, opened 2026-01-12
Batched - Vector: marking defaulted method properly

lucbv:
We should use the proper Kokkos macro to decorate our defaulted constructors, destructors and operators to avoid Cuda/HIP warnings.


ndellingwood (approved):
Thanks @lucbv ! Should this be a part of a patch release if there is a 5.0.2?

PR #2910, opened 2026-01-13
Refactor batched copy

yasahi-hpc:
This PR aims at improving the implementation and adding unit-tests of serial `copy`.

 - [x] Adding unit-tests for `SerialCopy`
 - [x] Remove the `rank` argument from the API. `rank` is derived from `View` rank.
 - [x] Moving implementation details inside `Impl` namespace. 
 - [x] Supporting `ConjTrans` operator
 - [x] Add docstring and assertion for `Arg` template parameters
 
Current API
```C++
template <typename ArgTrans = Trans::NoTranspose, int rank = 2>
struct SerialCopy {
  template <typename AViewType, typename BViewType>
  KOKKOS_INLINE_FUNCTION static int invoke(const AViewType &A, const BViewType &B);
};
```

New API

```C++
/// \brief Serial Batched Copy:
/// Performs B = Op(A)
/// where Op is one of NoTranspose, Transpose, ConjTranspose
/// A and B are 1D or 2D views
///
/// \tparam ArgTrans: one of NoTranspose, Transpose, ConjTranspose
/// \tparam Args: (deprecated) rank information
template <typename ArgTrans = Trans::NoTranspose, class... Args>
struct SerialCopy {
  static constexpr size_t size = sizeof...(Args);
  static_assert(size == 0,
                "KokkosBatched::SerialCopy<ArgTrans, rank> is deprecated. Please use "
                "KokkosBatched::SerialCopy<ArgTrans> instead.");
  static_assert(KokkosBlas::is_trans_v<ArgTrans>, "KokkosBatched::SerialCopy: ArgTrans must be a KokkosBlas::Trans.");

  /// \brief invoke the SerialCopy
  /// \tparam AViewType: Kokkos::View type for A
  /// \tparam BViewType: Kokkos::View type for B
  /// \param[in] A Input view A
  /// \param[out] B Output view B
  template <typename AViewType, typename BViewType>
  KOKKOS_INLINE_FUNCTION static int invoke(const AViewType &A, const BViewType &B);
};
```

### DIFF
--- a/batched/dense/src/KokkosBatched_Copy_Decl.hpp
+++ b/batched/dense/src/KokkosBatched_Copy_Decl.hpp
@@ -10,6 +10,14 @@
 
 namespace KokkosBatched {
 
+template <bool>
+struct SerialCopy_Deprecated_Warning {};
+
+template <>
+struct [[deprecated(
+    "KokkosBatched::SerialCopy<ArgTrans, rank> is deprecated. Please use "
+    "KokkosBatched::SerialCopy<ArgTrans> instead.")]] SerialCopy_Deprecated_Warning<true> {};
+
 /// \brief Serial Batched Copy:
 /// Performs B = Op(A)
 /// where Op is one of NoTranspose, Transpose, ConjTranspose
@@ -18,11 +26,7 @@ namespace KokkosBatched {
 /// \tparam ArgTrans: one of NoTranspose, Transpose, ConjTranspose
 /// \tparam Args: (deprecated) rank information
 template <typename ArgTrans = Trans::NoTranspose, class... Args>
-struct SerialCopy {
-  static constexpr size_t size = sizeof...(Args);
-  static_assert(size == 0,
-                "KokkosBatched::SerialCopy<ArgTrans, rank> is deprecated. Please use "
-                "KokkosBatched::SerialCopy<ArgTrans> instead.");
+struct SerialCopy : SerialCopy_Deprecated_Warning<sizeof...(Args) != 0> {
   static_assert(KokkosBlas::is_trans_v<ArgTrans>, "KokkosBatched::SerialCopy: ArgTrans must be a KokkosBlas::Trans.");
 
   /// \brief invoke the SerialCopy
@@ -34,6 +38,14 @@ struct SerialCopy {
   KOKKOS_INLINE_FUNCTION static int invoke(const AViewType &A, const BViewType &B);
 };
 
+template <bool>
+struct TeamCopy_Deprecated_Warning {};
+
+template <>
+struct [[deprecated(
+    "KokkosBatched::TeamCopy<ArgTrans, rank> is deprecated. Please use "
+    "KokkosBatched::TeamCopy<ArgTrans> instead.")]] TeamCopy_Deprecated_Warning<true> {};
+
 /// \brief Team Batched Copy:
 /// Performs B = Op(A)
 /// where Op is one of NoTranspose, Transpose, ConjTranspose
@@ -43,11 +55,7 @@ struct SerialCopy {
 /// \tparam ArgTrans: one of NoTranspose, Transpose, ConjTranspose
 /// \tparam Args: (deprecated) rank information
 template <typename MemberType, typename ArgTrans = Trans::NoTranspose, class... Args>
-struct TeamCopy {
-  static constexpr size_t size = sizeof...(Args);
-  static_assert(size == 0,
-                "KokkosBatched::TeamCopy<MemberType, ArgTrans, rank> is deprecated. Please use "
-                "KokkosBatched::TeamCopy<MemberType, ArgTrans> instead.");
+struct TeamCopy : TeamCopy_Deprecated_Warning<sizeof...(Args) != 0> {
   static_assert(KokkosBlas::is_trans_v<ArgTrans>, "KokkosBatched::TeamCopy: ArgTrans must be a KokkosBlas::Trans.");
 
   /// \brief invoke the TeamCopy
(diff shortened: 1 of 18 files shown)


lucbv (commented):
I only have one question, should we have a quick return when the views have `extent(0) == 0` ?

yasahi-hpc:
> I only have one question, should we have a quick return when the views have `extent(0) == 0` ?

Yeah, I added a quick return. Since this supports 2D case as well, I think we should quick return when `size() == 0`

batched/dense/src/KokkosBatched_Copy_Decl.hpp
```diff
@@ -4,55 +4,120 @@
...
-/// Serial Copy
-///
-
-template <typename ArgTrans = Trans::NoTranspose, int rank = 2>
+/// \tparam ArgTrans: one of NoTranspose, Transpose, ConjTranspose
+/// \tparam Args: (deprecated) rank information
+template <typename ArgTrans = Trans::NoTranspose, class... Args>
 struct SerialCopy {
+  static constexpr size_t size = sizeof...(Args);
+  static_assert(size == 0,
```
lucbv: Is there anyway to have a deprecation warning emitted instead of a build error? 
It would be nice to avoid a breaking change if possible although I do see that with the defaulted template argument we might not be able to easily achieve this?

lucbv (commented):
Mostly okay but I would prefer a softer approach than having build error for users.

batched/dense/src/KokkosBatched_Copy_Decl.hpp
yasahi-hpc: I understand your point. Actually, I have already tried to do that, but my conclusion is that it is probably impossible.

For example, if we add the following it does not compile since class template `"KokkosBatched::SerialCopy"` has already been defined

```C++
template <typename ArgTrans, int rank>
struct [[deprecated("KokkosBatched::SerialCopy<ArgTrans, rank> is deprecated. Please use KokkosBatched::SerialCopy<ArgTrans> instead.")]] SerialCopy {
  template <typename AViewType, typename BViewType>
  KOKKOS_INLINE_FUNCTION static int invoke(const AViewType &A, const BViewType &B) {
    return SerialCopy<ArgTrans>::invoke(A, B);
  }
};
```

yasahi-hpc: I found a way. May I have another review please

lucbv (approved):
Thanks for taking the time to add a deprecation warning

lucbv:
It looks like some there are some build issues to look at.
It would be good to also update the documentation to reflect the deprecation and new API assuming there is documentation for this function.

yasahi-hpc:
> It looks like some there are some build issues to look at. It would be good to also update the documentation to reflect the deprecation and new API assuming there is documentation for this function.

Fixed accordingly.
I could not find a documentation file for this. 
I will do that after finishing related PRs: Team, TeamVector and examples

yasahi-hpc:
This works at least with clang. No warnings with gcc
https://godbolt.org/z/E59rj1G3f

lucbv (approved):
Thanks for getting all of this done, it looks good to me!

ndellingwood:
I'm seeing failures in Ifpack2 in Hip and Cuda builds after these changes, here are some output snippets

Ifpack2_BlockTriDiContainer_Jacobi_UnitTests_MPI_4
```
...
11: SUCCESS: test_BR_BTDC (A = I) test_BR_BTDC: bs 1 contig jacobi point_wise uniform_lines nvec 1 rd 0
11: FAIL: test_BR_BTDC (A = I + R) test_BR_BTDC: bs 1 contig jacobi point_wise uniform_lines nvec 3 rd 0.853659
11: FAIL: test_BR_BTDC (A = D) test_BR_BTDC: bs 1 contig jacobi point_wise uniform_lines nvec 3 rd 1
11: FAIL: test_BR_BTDC (A = D + R) test_BR_BTDC: bs 1 contig jacobi point_wise uniform_lines nvec 3 rd 1
...
11: FAIL: test_BR_BTDC (A = I + R) test_BR_BTDC: bs 1 contig jacobi overlap_comm point_wise uniform_lines nvec 3 rd 1
11: FAIL: test_BR_BTDC (A = D) test_BR_BTDC: bs 1 contig jacobi overlap_comm point_wise uniform_lines nvec 3 rd 1
11: FAIL: test_BR_BTDC (A = D + R) test_BR_BTDC: bs 1 contig jacobi overlap_comm point_wise uniform_lines nvec 3 rd 1
11: FAIL: test_BR_BTDC (A = D + R, damping factor) test_BR_BTDC: bs 1 contig jacobi overlap_comm point_wise uniform_lines nvec 3 rd 1 eps 2.22045e-15
11: FAIL: test_BR_BTDC (A = D + R, damping factor) test_BR_BTDC: bs 1 contig jacobi overlap_comm point_wise uniform_lines nvec 3 rd 1 eps 2.22045e-15
...
```

yasahi-hpc:
@ndellingwood 
Can you try this [PR](https://github.com/kokkos/kokkos-kernels/pull/2935)
While implementing tests for `Team/TeamVector`, I found that I introduced a copy-paste error
This PR introduces tests for these functions and fixes the issues (including preexisting ones)

ndellingwood:
@yasahi-hpc thanks for pointing that out, testing in progress

ndellingwood:
@yasahi-hpc testing with PR #2935 passed in the Hip build, thanks!

yasahi-hpc:
@ndellingwood Great! Thank you for the quick response

PR #2911, opened 2026-01-14
Sell spmv

lucbv:


sparse/impl/KokkosSparse_spmv_sellmatrix_impl.hpp
```diff
@@ -0,0 +1,40 @@
...
+namespace Impl {
+
+template <class ExecutionSpace, class AMatrix, class XVector, class YVector>
+void spmv_sellmatrix_impl(const ExecutionSpace& space, [[maybe_unused]] const char mode[],
+                          const typename YVector::non_const_value_type& alpha, const AMatrix& A, const XVector& x,
+                          const typename YVector::non_const_value_type& beta, const YVector& y) {
+  using ordinal_type = typename AMatrix::non_const_ordinal_type;
+  using y_value_type = typename YVector::value_type;
+
+  const ordinal_type row_length = A.sell_nnz / A.num_rows;
```
brian-kelley, line 20: I'm probably confused but don't you have to look at ``A.slice_offsets`` to find the row length per slice? I think it's possible for the loop
```
        for (ordinal_type idx = 0; idx < row_length; ++idx) {
```
to read entries/values belonging to the next slice

sparse/unit_test/Test_Sparse_spmv_sell.hpp
```diff
@@ -0,0 +1,253 @@
...
+  //   }
+  //   std::cout << std::setw(2) << matA.entries((row_length - 1) * num_rows + rowIdx) << "]     ";
+
+  //   // Output values as well
+  //   std::cout << "   [";
+  //   for(int colIdx = 0; colIdx < row_length - 1; ++colIdx) {
+  //     std::cout << std::setw(14) << matA.values(colIdx * num_rows + rowIdx) << ", ";
+  //   }
+  //   std::cout << std::setw(14) << matA.values((row_length - 1) * num_rows + rowIdx) << "]" << std::endl;
+  // }
```
brian-kelley, line 195:
```suggestion
```

sparse/unit_test/Test_Sparse_spmv_sell.hpp
```diff
@@ -0,0 +1,253 @@
...
+  //       0  1  2  3  4  5  6
+  //     [ 2 -1               ]             [ 0  1 -1]            [ 2.0 -1.0  0.0]
+  //     [-1  2 -1            ]             [ 0  1  2]            [-1.0  2.0 -1.0]
+  //     [   -1  2 -1         ]             [ 1  2  3]            [-1.0  2.0 -1.0]
+  // A = [      -1  2 -1      ]   colinds = [ 2  3  4]   values = [-1.0  2.0 -1.0]
+  //     [         -1  2 -1   ]             [ 3  4  5]            [-1.0  2.0 -1.0]
+  //     [            -1  2 -1]             [ 4  5  6]            [-1.0  2.0 -1.0]
+  //     [               -1  2]             [ 5  6 -1]            [-1.0  2.0  0.0]
+
+  constexpr ordinal_type nrows = 7, ncols = 7, rows_per_slice = 7, nslices = 1;
```
brian-kelley, line 71: Is there a test case with nslices > 1?

sparse/impl/KokkosSparse_spmv_sellmatrix_impl.hpp
lucbv: Yes, that's correct, there should be a loop over the slices and then loops for each slice.
Here I am allowing to only do one slice because I check for it in the matrix constructor at runtime.

sparse/unit_test/Test_Sparse_spmv_sell.hpp
lucbv: No again starting with only one slice and making sure that this special use case works well.
It is checked in the constructor and it is the only case supported in rocsparse... although we can easily do multiple slices by calling the rocsparse kernels on each of the slices but that's why I am doing only one slice for now.

brian-kelley (approved):

sparse/unit_test/Test_Sparse_spmv_sell.hpp
brian-kelley: Ah, forgot about that. This is fine then

brian-kelley (approved):

brian-kelley (approved):

PR #2934, opened 2026-02-10
docs: adding wiki example for axpy and including it in docs

lucbv:
Moving the code for the axpy example into a properly tested example and including it back to documentation with `literalinclude` directive.
Also fixing the math mistake in the verification piece at the end of the code.

@cmakrides-hpe thanks for reporting this issue!

### DIFF
--- a/docs/source/API/blas/blas1_axpy.rst
+++ b/docs/source/API/blas/blas1_axpy.rst
@@ -52,33 +52,11 @@ Type Requirements
 Example
 =======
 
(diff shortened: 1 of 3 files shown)


yasahi-hpc (changes requested):
Probably, you need to format `example/wiki/blas/KokkosBlas1_wiki_axpy.cpp`
Other than that looks good

yasahi-hpc (approved):
LGTM

PR #2936, opened 2026-02-12
Add shell sort option for matrix and graph sorting.

cgcgcg:
See 
https://github.com/trilinos/Trilinos/pull/14935
https://github.com/trilinos/Trilinos/pull/14936


lucbv (changes requested):
Can you please update the documentation with the parameter for SHELL sort?
See here: https://github.com/kokkos/kokkos-kernels/blob/develop/docs/source/API/sparse/sort_crs.rst

lucbv (approved):
Thx! : )

PR #2937, opened 2026-02-13
[Docs] batched Serial/Team/TeamVector copy

yasahi-hpc:
This PR aims at introducing batched docs for copy with an example.

- [x] docs for copy
- [x] An example for copy

### DIFF
--- a/example/batched_solve/serial_copy.cpp
+++ b/example/batched_solve/serial_copy.cpp
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
+
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Random.hpp>
+#include <KokkosBatched_Copy_Decl.hpp>
+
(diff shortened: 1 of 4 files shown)


example/batched_solve/serial_copy.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+/// \brief Example of batched copy
+/// Performs B = Op(A) where Op is one of NoTranspose, Transpose, ConjTranspose
+/// A and B are 1D or 2D views
+///
+/// Usage example: copy A to B
+/// A: [1,2,3]
+/// B: [1,2,3]
+///
+int main(int /*argc*/, char** /*argv*/) {
```
lucbv, line 18: Note that in general you might want to pass these arguments to `Kokkos::initialize()` in case the user wants to test something like a different number of threads or selecting the GPU ID to run on etc...

lucbv (approved):
Okay, no problem overall, thanks!

example/batched_solve/serial_copy.cpp
yasahi-hpc: Should I update?
I am open to changes

PR #2938, opened 2026-02-14
Implement batched serial syr2

yasahi-hpc:
This PR implements [syr2](https://www.netlib.org/lapack/explore-html/dd/de5/group__her2_ga8e576e9319c25b883b11dc1f39366bcc.html#ga8e576e9319c25b883b11dc1f39366bcc) function.

`KokkosBatched_Syr.hpp` has also been slightly modified.

Following files are added:
1. `KokkosBatched_Syr2_Serial_Impl.hpp`: Internal interfaces
2. `KokkosBatched_Syr2_Serial_Internal.hpp`: Implementation details
3. `KokkosBatched_Syr2.hpp`: APIs
4. `Test_Batched_SerialSyr2.hpp`: Unit tests for that

## Detailed description

It performs the rank 2 operation `A:= alpha*x*y**T + alpha*y*x**T + A` ({s,d,c,z}syr2) or `alpha*x*y**H + conjg(alpha)*y*x**H + A` ({c,z}her2)
- `x`: `(batch_count, n)`  
On entry, it contains the n elements of x.
On entry, the leading n by n part of the array A must contain the matrix of coefficients. 
On exit, A is overwritten by the updated matrix. 
It should be noted that only the Upper or Lower triangular part is modified. 
Imaginary parts of the diagonal elements are suppressed for {c, z}her2.

```C++
Kokkos::parallel_for('syr2', 
    [=](const int k) {
      auto sub_y = Kokkos::subview(m_y, k, Kokkos::ALL());
      auto sub_A = Kokkos::subview(m_A, k, Kokkos::ALL(), Kokkos::ALL());

      KokkosBatched::SerialSyr2<typename ParamTagType::uplo, typename ParamTagType::trans>::invoke(m_alpha, sub_x,
                                                                                                               sub_y, sub_A);
    });
```

## Tests
1.  Make a random `x`, `y`, and `A`, while copying `A` into `A_ref`. The reference `A_ref` is computed by `A_ref:= alpha*x*y**T + alpha*y*x**T + A_ref` or `A_ref:= alpha*x*y**H + conjg(alpha)*y*x**H + A_ref` at host.  Finally, we confirm `A` computed by serial ger and `A_ref` are the same. A == 0 case is tested as well.
1.  Simple and small analytical test, i.e. choose `x`,  and `A` as follows to confirm `A` is updated as expected. Both upper and lower cases are tested.
```bash
A = [[1, -3, -2,  0],
     [0,  1, -3, -2],
     [0,  0,  1, -3],
     [0,  0,  0,  1]]
x = [1, 2, 3, 4]
y: [4, 3, 2, 1]
Ref: [[13.,  13.5,  19.,  25.5],
         [ 0.,  19.,  16.5,  19., ],
         [ 0.,   0.,  19.,  13.5, ],
         [ 0.,   0.,   0.,  13.,  ]]
```

### DIFF
--- a/batched/dense/impl/KokkosBatched_Syr2_Serial_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Syr2_Serial_Impl.hpp
@@ -1,0 +1,152 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
+
+#ifndef KOKKOSBATCHED_SYR2_SERIAL_IMPL_HPP_
+#define KOKKOSBATCHED_SYR2_SERIAL_IMPL_HPP_
+
+#include <KokkosBlas_util.hpp>
+#include <KokkosBatched_Util.hpp>
+#include "KokkosBatched_Syr2_Serial_Internal.hpp"
+
+namespace KokkosBatched {
+namespace Impl {
+template <typename XViewType, typename YViewType, typename AViewType>
+KOKKOS_INLINE_FUNCTION static int checkSyr2Input([[maybe_unused]] const XViewType &x,
+                                                 [[maybe_unused]] const YViewType &y,
+                                                 [[maybe_unused]] const AViewType &A) {
+  static_assert(Kokkos::is_view_v<XViewType>, "KokkosBatched::syr2: XViewType is not a Kokkos::View.");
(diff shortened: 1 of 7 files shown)


lucbv (approved):
Not important but we could add a utility function: is_uplo to perform checks instead of the current "or" statement

yasahi-hpc:
> Not important but we could add a utility function: is_uplo to perform checks instead of the current "or" statement

Good point
I can make concepts for these checks
Maybe in another PR?

lucbv:
Sure, another PR trying to do this with concepts would be good. I'll merge this now

Issue #2942, opened 2026-02-17
`constexpr` `mb()` BLAS utility functions could affect type layout


cwpearson:
Kokkos forbids using these macros in a `constexpr`-y way

https://github.com/kokkos/kokkos-kernels/blob/dde1d20ba380415724c68fe017011e6ab075dc3a/blas/impl/KokkosBlas_util.hpp#L98-L102

https://github.com/kokkos/kokkos-kernels/blob/dde1d20ba380415724c68fe017011e6ab075dc3a/blas/impl/KokkosBlas_util.hpp#L145-L149

These functions are only safe if they are used in a way that never affects type layout. 

Imagine the following on a unified-memory system:

```
struct Functor {
  int data[mb()];
  double x;
};

...

Functor f;
f.x = 3.14;
kernel<<<1,1>>>(f);
```

Member `x` of `Functor` is at a different offset on host and device, so the implicit copy of `f` from host to device at the kernel launch won't work.

We already use them dangerously, e.g. anyone who modifies `InnerGemmFixC` needs to magically know the struct layout can't depend on these template parameters, and that's only because of how those template parameters are produced, nothing intrinsic about `InnterGemmFixC` itself.

https://github.com/kokkos/kokkos-kernels/blob/dde1d20ba380415724c68fe017011e6ab075dc3a/batched/dense/impl/KokkosBatched_Gemm_Serial_Internal.hpp#L75-L77

https://github.com/kokkos/kokkos-kernels/blob/dde1d20ba380415724c68fe017011e6ab075dc3a/batched/dense/impl/KokkosBatched_Gemm_Serial_Internal.hpp#L89

There's no way to enforce this, so these could be rewritten in a way where the return value is the same no matter whether the host or device compilation pass is asking.

lucbv:
@cwpearson is this now resolved with #2943 or do we need additional work to finish it up?

PR #2943, opened 2026-02-17
Don't use KOKKOS_IF_ON_HOST/DEVICE in a constexpr way in `mb`

cwpearson:
Fix for #2942

The basic transformation is from this

```c++
static constexpr KOKKOS_FUNCTION int mb() { 
  KOKKOS_IF_ON_HOST((return 4;)) 
  KOKKOS_IF_ON_DEVICE((return 2;)) 
} 

// called inside a parallel region
KOKKOS_INLINE_FUNCTION int foo() {
  constexpr int B = mb();
  GEMM<B> g();
  g.invoke();
}
```

to this

```c++
// now, each call to `mb` is the same from host and device side
struct Host{};
struct Device{};

template <typename Where>
constexpr static KOKKOS_INLINE_FUNCTION int mb() {
  if constexpr (std::is_same_v<Where, Host>) {
    return 4;
  } else {
    return 2;
  }
}

// dispatch specialized for host or device is always self-consistent
KOKKOS_INLINE_FUNCTION int foo() {
  auto host_or_device = [&](auto tag) {
    constexpr int B = mb<decltype(tag)>();
    GEMM<B> g();
    g.invoke();
  };

  KOKKOS_IF_ON_HOST((host_or_device(Host{});))
  KOKKOS_IF_ON_DEVICE((host_or_device(Device{});))
}
```

### DIFF
--- a/batched/dense/impl/KokkosBatched_Gemm_Team_Internal.hpp
+++ b/batched/dense/impl/KokkosBatched_Gemm_Team_Internal.hpp
@@ -136,8 +136,8 @@ KOKKOS_INLINE_FUNCTION int TeamGemmInternal<Algo::Gemm::Blocked>::invoke(
         // } // for jj
       }
     };
-    KOKKOS_IF_ON_HOST((host_or_device(Algo::Host{});))
-    KOKKOS_IF_ON_DEVICE((host_or_device(Algo::Device{});))
+    KOKKOS_IF_ON_HOST((host_or_device(Algo::Gemm::Blocked::Impl::Host{});))
+    KOKKOS_IF_ON_DEVICE((host_or_device(Algo::Gemm::Blocked::Impl::Device{});))
   }
   return 0;
 }
--- a/batched/dense/impl/KokkosBatched_Gemm_Team_Internal.hpp
+++ b/batched/dense/impl/KokkosBatched_Gemm_Team_Internal.hpp
@@ -73,9 +73,6 @@ KOKKOS_INLINE_FUNCTION int TeamGemmInternal<Algo::Gemm::Blocked>::invoke(
   // C = beta C + alpha A B
   // C (m x n), A(m x k), B(k x n)
 
-  constexpr int mbAlgo = Algo::Gemm::Blocked::Impl::mb();
-  constexpr int nbAlgo = Algo::Gemm::Blocked::Impl::mb();
-
   const ScalarType one(1.0), zero(0.0);
 
   if (beta == zero)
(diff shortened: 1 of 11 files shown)


cwpearson:
Mostly curious about whether any of our compilers complain about the generic lambda. If not, I'll apply that pattern everywhere instead of duplicating so much code.

blas/impl/KokkosBlas_util.hpp
```diff
@@ -95,10 +98,8 @@ struct Algo {
       // - team policy (smaller) or range policy (bigger)
       // - space (gpu vs host)
       // - blocksize input (blk <= 4 mb = 2, otherwise mb = 4), etc.
-      static constexpr KOKKOS_FUNCTION int mb() {
-        KOKKOS_IF_ON_HOST((return 4;))
-        KOKKOS_IF_ON_DEVICE((return 2;))
-      }
+      static constexpr KOKKOS_INLINE_FUNCTION int mb(Host) { return 4; }
+      static constexpr KOKKOS_INLINE_FUNCTION int mb(Device) { return 2; }
```
yasahi-hpc:
```suggestion
template <typename ExecSpace>
static constexpr KOKKOS_INLINE_FUNCTION int mb();

template <>
static constexpr KOKKOS_INLINE_FUNCTION int mb<Host>() {
    return 4;
}

template <>
static constexpr KOKKOS_INLINE_FUNCTION int mb<Device>() {
    return 2;
}
```

Can we use a template argument instead?

batched/dense/impl/KokkosBatched_Gemm_Team_Internal.hpp
```diff
@@ -88,57 +85,61 @@ KOKKOS_INLINE_FUNCTION int TeamGemmInternal<Algo::Gemm::Blocked>::invoke(
...
-      });
+    auto host_or_device = [&](auto tag) {
+      constexpr int mbAlgo = Algo::Gemm::Blocked::mb(decltype(tag){});
+      constexpr int nbAlgo = Algo::Gemm::Blocked::mb(decltype(tag){});
+      InnerGemmFixC<mbAlgo, nbAlgo> inner(as0, as1, bs0, bs1, cs0, cs1);
+      auto gemm = [&](const int ib, const int jb, const int pb, const ValueType *KOKKOS_RESTRICT AA,
+                      const ValueType *KOKKOS_RESTRICT BB,
+                      /**/ ValueType *KOKKOS_RESTRICT CC) {
+        // Made this non-const in order to WORKAROUND issue #349
+        int mb = mbAlgo, mp = (ib % mb), mq = (ib / mb) + (mp > 0), nb = nbAlgo, np = (jb % nb),
```
yasahi-hpc: It may be out of the scope, but can we make this `const` now?

blas/impl/KokkosBlas_util.hpp
```diff
@@ -83,6 +83,9 @@ template <class T>
 static constexpr bool is_trans_v = is_trans<T>::value;
 
 struct Algo {
+  struct Host {};
+  struct Device {};
```
yasahi-hpc: Should we expose these types in API level?
Unless people are already using `Level3::Blocked::mb` value in their downstream apps, it may be safer to put these inside `Impl` namespace

yasahi-hpc (commented):
I think it is safer to put `Host` and `Device` types under `Impl` namespace.
Other than that looks good to me

blas/impl/KokkosBlas_util.hpp
cwpearson: Good point, probably we should deprecate `mb` and make an `Impl` version. I'll do that as a separate PR to merge first, and then also put these in Impl too.

yasahi-hpc: Good
Then, fine with me

yasahi-hpc (approved):

blas/impl/KokkosBlas_util.hpp
cwpearson: https://github.com/kokkos/kokkos-kernels/pull/2950

blas/impl/KokkosBlas_util.hpp
```diff
@@ -166,7 +185,10 @@ struct Algo {
       // - team policy (smaller) or range policy (bigger)
       // - space (cuda vs host)
       // - blocksize input (blk <= 4 mb = 2, otherwise mb = 4), etc.
-      [[deprecated]] static constexpr KOKKOS_FUNCTION int mb() { return Impl::mb(); }
+      [[deprecated]] static constexpr KOKKOS_FUNCTION int mb() {
```
yasahi-hpc:
```suggestion
      [[deprecated("Use mb<Where>() instead")]] static constexpr KOKKOS_FUNCTION int mb() {
```

Or may be better to claim that 
```C++
[[deprecated("Do not use implementation detail mb()")]]
```

yasahi-hpc (approved):
LGTM
I would prefer to add a deprecated message to `mb()`

lucbv (approved):

lucbv:
Doing a quick update of this branch before merging it to develop since a lot of other PRs have gone in first.

lucbv (approved):

PR #2944, opened 2026-02-17
Add Blas Concepts

yasahi-hpc:
This PR introduces concepts from `KokkosBlas` traits which are planed to be used in `KokkosBatched`.
Currently, we only have the following concepts.
Similar concepts should be defined for `KokkosBatched`

```C++
template <typename T>
concept TransposeOperation = is_trans_v<T>;

template <typename T>
concept BlasLevel2 = is_level2_v<T>;

template <typename T>
concept BlasLevel3 = is_level3_v<T>;
```

### DIFF
--- a/blas/unit_test/Test_Blas_Concepts.hpp
+++ b/blas/unit_test/Test_Blas_Concepts.hpp
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
+#ifndef TEST_BLAS_CONCEPTS_HPP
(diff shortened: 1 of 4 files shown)


blas/unit_test/Test_Blas_Concepts.hpp
lucbv, line 1: Can we add a test where we actually use the concept, i.e. create a dummy class/struct with a requirement close and observe that it goes well?

lucbv (changes requested):
It looks good overall but I think at least one more test to exemplify how this will work in practice would be nice : )

blas/unit_test/Test_Blas_Concepts.hpp
yasahi-hpc: I added a dummy functor to explain the typical use case in mind

PR #2945, opened 2026-02-17
cm_generate_makefile: add flags for benchmarks

brian-kelley:
In cm_generate_makefile, add flags ``--enable-benchmarks`` and ``--enable-kokkos-benchmarks`` to enable KokkosKernels and Kokkos (core) benchmarks respectively to both build and run with ctest. These are still off by default.


ndellingwood (approved):
thanks @brian-kelley

lucbv (approved):

PR #2948, opened 2026-02-18
[Docs] batched Serial syr2

yasahi-hpc:
This PR aims at introducing batched docs for syr2 with an example.

- [x] docs for `syr2`
- [x] An example for `syr2`


lucbv (approved):
Looks good, thanks for documenting!

PR #2949, opened 2026-02-18
SortCRS: pick between shell/radix automatically; speed up radix

brian-kelley:
- Add STDSORT as a sorting algorithm option for graphs on CPU
  - ``std::sort`` has no sort-by-key version, which is why this is only available for graphs
- Change behavior of DEFAULT SortCrs algorithm on CPUs
  - If it's a matrix with < 20 entries per row on average, use shell sort
  - Otherwise use radix
  - If it's a graph with < 30 entries per row, use std::sort
  - Otherwise use radix
- Improve radix implementation (take half as many passes through data) and testing

For matrices, there is a more conservative cutoff of 20 entries/row, even though radix is often now faster for shuffled inputs of fewer entries/row, for 2 reasons:
- shell is faster when the matrix is already sorted or almost sorted, while radix takes the same amount of time for any input permutation. The numbers below are for randomly shuffled input.
- radix gets slower with the log of maximum key (aka column, when sorting CRS).

These two properties can't be measured in constant time like nnz/row, so they're not used in the heuristic.

Here are some timings (serial backend, 50k rows, N approx entries per row, CrsMatrix):

Eclipse (Broadwell):

| N  | Radix (old) | Shell      | Radix (new) | Default (new) |
|----|-------------|------------|-------------|---------------|
| 10 | 0.0115071   | 0.00811405 | 0.0055027   |  0.00834978   |
| 20 | 0.0168709   | 0.0234198  | 0.00954263  |  0.0229293   |
| 30 | 0.0196942   | 0.0416072  | 0.0136842   |   0.0150321   |
| 40 | 0.0239652   | 0.0616507  | 0.0188068   |   0.0196313   |

Blake (Granite rapids):

| N  | Radix (old) | Shell      | Radix (new) | Default (new) |
|----|-------------|------------|-------------|---------------|
| 10 |  0.0182581  | 0.00507935  |   0.0110481 |   0.00506758  |
| 20 |  0.020158  |  0.0156176 | 0.0104301  |    0.015632  |
| 30 | 0.037822   |  0.0280926 |  0.0119157  |   0.0123646   |
| 40 |  0.0433125  |  0.0409841 |  0.0152853  |   0.0151538   |

### DIFF
--- a/common/src/KokkosKernels_Sorting.hpp
+++ b/common/src/KokkosKernels_Sorting.hpp
@@ -216,9 +216,16 @@ KOKKOS_INLINE_FUNCTION void SerialRadixSort(ValueType* values, ValueType* values
   static_assert(std::is_integral<ValueType>::value && std::is_unsigned<ValueType>::value,
                 "radixSort can only be run on unsigned integers.");
   if (n <= 1) return;
-  ValueType maxVal = 0;
+  // sort 4 bits at a time, into 16 buckets
+  constexpr ValueType mask = 0xF;
+  // Find the maximum value so we know how many passes to run,
+  // and at the same time do the first round of counting buckets
+  ValueType maxVal  = 0;
+  Ordinal count[16] = {0};
   for (Ordinal i = 0; i < n; i++) {
-    if (maxVal < values[i]) maxVal = values[i];
+    auto val = values[i];
+    count[val & mask]++;
+    if (maxVal < val) maxVal = val;
   }
   // determine how many significant bits the data has
   int passes = 0;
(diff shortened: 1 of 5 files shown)


brian-kelley:
This PR failed the docs check, but I don't think any updates are needed there. The sort_crs page already talks about the default/radix/shell algorithms and I didn't change any interfaces.

update: I added STDSORT to the list of algorithms in the docs

lucbv:
Yes, the documentation failure is an artifact of having the implementation details in the public header. It's something we are not too strict about in the "common/" directory since we do not have so much code there but we could look at cleaning it up in the future.

lucbv (approved):
This looks reasonable, I read the bit manipulation part as best I could

PR #2950, opened 2026-02-18
Deprecate `Level2::Blocked::mb` and `Level3::Blocked::mb`

cwpearson:
No replacement. Move into an `Impl` namespace.

Chances are no one was using these.


blas/impl/KokkosBlas_util.hpp
cwpearson, line 1: Somehow the format check was happy before this and also after this change.

yasahi-hpc (approved):
LGTM

lucbv (approved):

PR #2952, opened 2026-02-19
Add some missing SPMV labels

### LINKED ISSUE #2951: Missing SPMV label
https://github.com/kokkos/kokkos-kernels/blob/62be7921c5387239be0f67b4eb62848363c813cf/sparse/src/KokkosSparse_spmv.hpp#L248-L261

The `else` branch here is not necessarily happening via TPL, so we would really benefit from a label here.

For example, in a Serial build of the Tpetra CGsolve performance test, there's no KokkosSparse::spmv profiling region anywhere.

cwpearson:
Fixes #2951

Also fixes a missing `popRegion` that is passed over by an earlier `return`

### DIFF
--- a/sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
+++ b/sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
@@ -106,6 +106,8 @@ struct SPMV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, false, 
 
   static void spmv_bsrmatrix(const ExecutionSpace &space, Handle *handle, const char mode[], const YScalar &alpha,
                              const AMatrix &A, const XVector &X, const YScalar &beta, const YVector &Y) {
+    std::string label = "KokkosSparse::spmv[NATIVE,BSRMATRIX," + KokkosKernels::ArithTraits<YScalar>::name() + "]";
+    Kokkos::Profiling::pushRegion(label);
     const bool modeIsNoTrans        = (mode[0] == NoTranspose[0]);
     const bool modeIsConjugate      = (mode[0] == Conjugate[0]);
     const bool modeIsConjugateTrans = (mode[0] == ConjugateTranspose[0]);
@@ -114,8 +116,10 @@ struct SPMV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, false, 
     // use V41 if requested
     if (handle->algo == SPMV_BSR_V41) {
       if (modeIsNoTrans || modeIsConjugate) {
+        Kokkos::Profiling::popRegion();
         return Bsr::spMatVec_no_transpose(space, handle, alpha, A, X, beta, Y, modeIsConjugate);
       } else if (modeIsTrans || modeIsConjugateTrans) {
+        Kokkos::Profiling::popRegion();
         return Bsr::spMatVec_transpose(space, handle, alpha, A, X, beta, Y, modeIsConjugateTrans);
       }
     }
@@ -124,17 +128,21 @@ struct SPMV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, false, 
     if (KokkosKernels::Impl::is_gpu_exec_space_v<ExecutionSpace> || handle->algo == SPMV_BSR_V42) {
       if (modeIsNoTrans) {
         ::KokkosSparse::Impl::apply_v42(space, alpha, A, X, beta, Y);
+        Kokkos::Profiling::popRegion();
         return;
       }
     }
 
     // fall back to V41 all else fails
     if (modeIsNoTrans || modeIsConjugate) {
+      Kokkos::Profiling::popRegion();
       return Bsr::spMatVec_no_transpose(space, handle, alpha, A, X, beta, Y, modeIsConjugate);
     } else if (modeIsTrans || modeIsConjugateTrans) {
+      Kokkos::Profiling::popRegion();
       return Bsr::spMatVec_transpose(space, handle, alpha, A, X, beta, Y, modeIsConjugateTrans);
     }
 
+    Kokkos::Profiling::popRegion();
     {
       std::stringstream ss;
       ss << __FILE__ << ":" << __LINE__ << " ";
@@ -157,6 +165,8 @@ struct SPMV_MV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, fals
 
   static void spmv_mv_bsrmatrix(const ExecutionSpace &space, Handle *handle, const char mode[], const YScalar &alpha,
                                 const AMatrix &A, const XVector &X, const YScalar &beta, const YVector &Y) {
+    std::string label = "KokkosSparse::spmv[NATIVE,MV,BSMATRIX," + KokkosKernels::ArithTraits<YScalar>::name() + "]";
+    Kokkos::Profiling::pushRegion(label);
 #if defined(KOKKOS_ENABLE_CUDA) && (defined(KOKKOS_ARCH_AMPERE) || defined(KOKKOS_ARCH_VOLTA))
     Method method = Method::Fallback;
     {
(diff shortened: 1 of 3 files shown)


sparse/src/KokkosSparse_spmv.hpp
```diff
@@ -284,17 +287,23 @@ void spmv(const ExecutionSpace& space, Handle* handle, const char mode[], const
...
-                                                                                                             alpha, A_i,
-                                                                                                             x_i, beta,
-                                                                                                             y_i);
+        auto ret =
+            Impl::SPMV_MV<ExecutionSpace, HandleImpl, AMatrix_Internal, XVector_Internal, YVector_Internal,
+                          std::is_integral<typename AMatrix_Internal::value_type>::value, false>::spmv_mv(space, handle,
+                                                                                                          mode, alpha,
+                                                                                                          A_i, x_i,
+                                                                                                          beta, y_i);
         Kokkos::Profiling::popRegion();
```
cwpearson: This `popRegion` was never called.

brian-kelley (approved):

cwpearson:
re docs: This PR doesn't change the spmv interface

lucbv:
I am not sure I'm 100% happy with this, should we not add the region and label in the spec file around the dispatch call inside the `struct SPMV<..., false, KOKKOSKERNELS_IMPL_COMPILE_LIBRARY>` [here](https://github.com/kokkos/kokkos-kernels/blob/62be7921c5387239be0f67b4eb62848363c813cf/sparse/impl/KokkosSparse_spmv_spec.hpp#L134)?
This would make it similar to how we do this for TPL calls, in which case we could make it a design rule that we have to create profiling regions in our `impl/KokkosSprase_spmv_spec.hpp` and `tpls/KokkosSparse_spmv_tpl_spec_decl.hpp`

cwpearson:
@lucbv is this what you had in mind?

sparse/src/KokkosSparse_spmv.hpp
```diff
@@ -351,13 +334,9 @@ void spmv(const ExecutionSpace& space, Handle* handle, const char mode[], const
 #endif
       if (useNative) {
         // Explicitly call the non-TPL SPMV_BSRMATRIX implementation
-        std::string label = "KokkosSparse::spmv[NATIVE,MV,BSMATRIX," +
```
cwpearson, line 354: These matrices are definitely some BS 😆

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
```diff
@@ -114,8 +116,10 @@ struct SPMV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, false,
     // use V41 if requested
     if (handle->algo == SPMV_BSR_V41) {
       if (modeIsNoTrans || modeIsConjugate) {
+        Kokkos::Profiling::popRegion();
```
lucbv, line 120: I would actually do something like this:
```c++
Bsr::spMatVec_no_transpose(...);
Kokkos::Profiling::popRegion();
return;
```
so that the profiling region actually encapsulate the spmv kernel, currently I think the kernels is not captured which will look weird when looking at timing or to detect where a bug might be happening

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
```diff
@@ -114,8 +116,10 @@ struct SPMV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, false,
...
         return Bsr::spMatVec_no_transpose(space, handle, alpha, A, X, beta, Y, modeIsConjugate);
       } else if (modeIsTrans || modeIsConjugateTrans) {
+        Kokkos::Profiling::popRegion();
```
lucbv, line 124: Same as above

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
```diff
@@ -124,17 +128,21 @@ struct SPMV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, false,
...
       if (modeIsNoTrans) {
         ::KokkosSparse::Impl::apply_v42(space, alpha, A, X, beta, Y);
+        Kokkos::Profiling::popRegion();
         return;
       }
     }
 
     // fall back to V41 all else fails
     if (modeIsNoTrans || modeIsConjugate) {
+      Kokkos::Profiling::popRegion();
```
lucbv, line 141: Same as above

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
```diff
@@ -124,17 +128,21 @@ struct SPMV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, false,
...
       return Bsr::spMatVec_no_transpose(space, handle, alpha, A, X, beta, Y, modeIsConjugate);
     } else if (modeIsTrans || modeIsConjugateTrans) {
+      Kokkos::Profiling::popRegion();
```
lucbv, line 145: Same as above

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
```diff
@@ -124,17 +128,21 @@ struct SPMV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, false,
...
       return Bsr::spMatVec_transpose(space, handle, alpha, A, X, beta, Y, modeIsConjugateTrans);
     }
 
+    Kokkos::Profiling::popRegion();
```
lucbv, line 149: What are we popping here?

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
```diff
@@ -238,8 +253,10 @@ struct SPMV_MV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, fals
     // use V41 if requested
     if (handle->algo == SPMV_BSR_V41) {
       if (modeIsNoTrans || modeIsConjugate) {
+        Kokkos::Profiling::popRegion();
```
lucbv, line 261: Same as above

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
```diff
@@ -238,8 +253,10 @@ struct SPMV_MV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, fals
...
         return Bsr::spMatMultiVec_no_transpose(space, handle, alpha, A, X, beta, Y, modeIsConjugate);
       } else if (modeIsTrans || modeIsConjugateTrans) {
+        Kokkos::Profiling::popRegion();
```
lucbv, line 265: Same as above

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
```diff
@@ -248,17 +265,21 @@ struct SPMV_MV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, fals
...
       if (modeIsNoTrans) {
         ::KokkosSparse::Impl::apply_v42(space, alpha, A, X, beta, Y);
+        Kokkos::Profiling::popRegion();
         return;
       }
     }
 
     // use V41 as the ultimate fallback
     if (modeIsNoTrans || modeIsConjugate) {
+      Kokkos::Profiling::popRegion();
```
lucbv, line 282: Same as above

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
```diff
@@ -248,17 +265,21 @@ struct SPMV_MV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, fals
...
       return Bsr::spMatMultiVec_no_transpose(space, handle, alpha, A, X, beta, Y, modeIsConjugate);
     } else if (modeIsTrans || modeIsConjugateTrans) {
+      Kokkos::Profiling::popRegion();
```
lucbv, line 286: Same as above

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
```diff
@@ -248,17 +265,21 @@ struct SPMV_MV_BSRMATRIX<ExecutionSpace, Handle, AMatrix, XVector, YVector, fals
...
       return Bsr::spMatMultiVec_transpose(space, handle, alpha, A, X, beta, Y, modeIsConjugateTrans);
     }
 
+    Kokkos::Profiling::popRegion();
```
lucbv, line 290: What are we popping?

lucbv (changes requested):
I am not convinced that the push/pop regions are in the right location and fully matching though out the code...

cwpearson:
I also added the integer special-cased MV spmvs that I missed

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
cwpearson: Still popping the region in the case where we trigger the internal error

sparse/impl/KokkosSparse_spmv_bsrmatrix_spec.hpp
cwpearson: Still popping the region in the case where we trigger the internal error

lucbv (approved):
Thanks for all the improvements to our timers : )

PR #2956, opened 2026-02-22
Fix FindTPLCUBLAS.cmake and friends

### LINKED ISSUE #2955: FindTPLCUBLAS.cmake only checks first CUDA::cublas include directory
FindTPLCUBLAS.cmake does:
```
get_target_property(kk_cublas_include_dir_list CUDA::cublas INTERFACE_INCLUDE_DIRECTORIES)
list(GET kk_cublas_include_dir_list 0 kk_cublas_include_dir)
```

Since INTERFACE_INCLUDE_DIRECTORIES is a list, using only the first entry can miss valid cuBLAS headers located in later include directories. In that case, configuration fails with:
```
Could NOT find TPLCUBLAS (missing: CUBLAS_INCLUDE_DIRS)
```
even though find_package(CUDAToolkit) succeeds and cublas_v2.h exists in one of the include directories of CUDA::cublas.

From some quick experiments, I believe it's sufficient to replace the second line with
```
set(kk_cublas_include_dir ${kk_cublas_include_dir_list})
```
so that the full list is transferred.

wcwitt:
Fixes #2955.

### DIFF
--- a/cmake/Modules/FindTPLCUBLAS.cmake
+++ b/cmake/Modules/FindTPLCUBLAS.cmake
@@ -2,7 +2,7 @@ find_package(CUDAToolkit)
 
 if(CUDAToolkit_FOUND)
   get_target_property(kk_cublas_include_dir_list CUDA::cublas INTERFACE_INCLUDE_DIRECTORIES)
-  list(GET kk_cublas_include_dir_list 0 kk_cublas_include_dir)
+  set(kk_cublas_include_dir ${kk_cublas_include_dir_list})
   get_target_property(kk_cublas_library CUDA::cublas IMPORTED_LOCATION)
   get_filename_component(kk_cublas_library_dir ${kk_cublas_library} DIRECTORY)
   kokkoskernels_find_imported(CUBLAS INTERFACE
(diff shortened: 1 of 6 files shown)


lucbv:
@wcwitt thanks for this pull request, do you know if CMake prioritizes the list that gets passed in some way?
The reason for picking the first item on the list was to ensure that we use the Release build and not the Debug build which is typically listed second...

Also could you sign your commits with `git commit -s` so that our DCO check can pass.

wcwitt:
> do you know if CMake prioritizes the list that gets passed in some way?

From what I can find, the compiler searches the paths in the order they appear, but not 100% sure.

The context is that I have a Kokkos codebase that recently stopped building on several machines after upgrading Kokkos and Kokkos Kernels. I'm not sure exactly which precise version introduced the problem, apologies.

One of those machines is Perlmutter with CUDA 12.9 loaded. Early during the CMake configuration, I see
```
-- Found CUDAToolkit: 
   /opt/nvidia/hpc_sdk/Linux_x86_64/25.5/cuda/12.9/targets/x86_64-linux/include;
   /opt/nvidia/hpc_sdk/Linux_x86_64/25.5/math_libs/12.9/include
   (found version "12.9.41")
```
but then later the configuration fails with
```
Could NOT find TPLCUBLAS (missing: CUBLAS_INCLUDE_DIRS)
```
and related errors. From what I can tell, that second CUDAToolkit include directory (where the math_libs live) is getting dropped.

My fix allows the builds to succeed again, but open to other ideas.

lucbv:
I am waiting a bit to see if our V100 Kumquat machine can get back online and test this otherwise I think we are good to merge. This was tested at NERSC and in our cuda13.1 image and worked well.

lucbv (approved):

wcwitt:
Thanks for merging! Any guesses as to when this will make it into a release?

JBludau:
it seems to be in the release candidate branch for 5.1.0

PR #2960, opened 2026-02-23
Implement batched serial/team/teamvector rot

yasahi-hpc:
This PR implements [rot](https://www.netlib.org/lapack/explore-html/d1/d45/group__rot_gae48ef017306866ac2d5a8c5a52617858.html#gae48ef017306866ac2d5a8c5a52617858) function.

`KokkosBatched_Rot.hpp` has also been slightly modified to support all of `{s,d,cs,zd}rot` or `{c,z}rot` operations.

Following files are added:
1. `KokkosBatched_Rot_Impl.hpp`: Internal interfaces
2. `KokkosBatched_Rot_Internal.hpp`: Implementation details
3. `KokkosBatched_Rot.hpp`: APIs
4. `Test_Batched_Rot.hpp`: Unit tests for that

## Detailed description

It performs the rank 1 operation `{s,d,cs,zd}rot` or `{c,z}rot`
```
x(i) := c*x(i) + s*y(i)
y(i) := c*y(i) - s*x(i) // {s,d,cs,zd}rot
y(i) := c*y(i) - conj(s)*x(i) // {c,z}rot
```

We check that `s` is real for `{s,d,cs,zd}rot` and `s` is complex for `{c,z}rot`. `c` is always real.

```C++
Kokkos::parallel_for('rot', 
    [=](const int k) {
      auto sub_y = Kokkos::subview(m_y, k, Kokkos::ALL());

      KokkosBatched::SerialRot<ArgTrans>::invoke(sub_x, sub_y, c, s);
    });
```

## Tests
1.  Make a random `x` and `y`, while copying `x` and `y` into `x_ref` and `y_ref`. The reference `x_ref` and `y_ref` are computed by the formula at host.  Finally, we confirm `x` and `y` computed by serial `rot` are identical to the reference.
1.  Simple and small analytical test, i.e. choose `x`,  and `y` as follows to confirm `x_ref` and `y_ref` are updated as expected.

```bash
c = 0.6, s = 0.8
x = [1, 2, 3, 4]
y = [5, 6, 7, 8]
x_ref = [4.6, 6.0, 7.4, 8.8]
y_ref = [2.2, 2.0, 1.8, 1.6]
```


batched/dense/src/KokkosBatched_Rot.hpp
```diff
@@ -0,0 +1,112 @@
...
+/// \tparam SType: Input type for the sine s (real or complex)
+///
+/// \param[inout] x: x is a length n vector, a rank 1 view
+/// \param[inout] y: y is a length n vector, a rank 1 view
+/// \param[in] c: cosine of the rotation (real scalar)
+/// \param[in] s: sine of the rotation (real or complex scalar)
+///
+/// No nested parallel_for is used inside of the function.
+///
+template <typename ArgTrans>
```
lucbv: I do not like calling this ArgTrans, would ArgConj be better since that is actually the only option you have here: regular apply or conjugate apply?
Also it then probably makes sense to default it on `NoTranspose`?
```suggestion
template <typename ArgTrans = KokkosBlas::Trans::NoTranspose>
```

batched/dense/src/KokkosBatched_Rot.hpp
```diff
@@ -0,0 +1,112 @@
...
+struct SerialRot {
+  static_assert(std::is_same_v<ArgTrans, Trans::Transpose> || std::is_same_v<ArgTrans, Trans::ConjTranspose>,
```
lucbv:
```suggestion
  static_assert(std::is_same_v<ArgTrans, Trans::NoTranspose> || std::is_same_v<ArgTrans, Trans::ConjTranspose>,
```

batched/dense/src/KokkosBatched_Rot.hpp
```diff
@@ -0,0 +1,112 @@
...
+///
+/// \param[in] member: TeamPolicy member
+/// \param[inout] x: x is a length n vector, a rank 1 view
+/// \param[inout] y: y is a length n vector, a rank 1 view
+/// \param[in] c: cosine of the rotation (real scalar)
+/// \param[in] s: sine of the rotation (real or complex scalar)
+///
+/// A nested parallel_for with TeamThreadRange is used.
+///
+template <typename MemberType, typename ArgTrans>
```
lucbv: See above on `ArgTrans`

batched/dense/src/KokkosBatched_Rot.hpp
```diff
@@ -0,0 +1,112 @@
...
+///
+/// \param[in] member: TeamPolicy member
+/// \param[inout] x: x is a length n vector, a rank 1 view
+/// \param[inout] y: y is a length n vector, a rank 1 view
+/// \param[in] c: cosine of the rotation (real scalar)
+/// \param[in] s: sine of the rotation (real or complex scalar)
+///
+/// A nested parallel_for with TeamVectorRange is used.
+///
+template <typename MemberType, typename ArgTrans>
```
lucbv: See above regarging `ArgTrans`

lucbv (changes requested):
I am not quite happy with asking for a Transpose mode when we really only want `conj` so this needs some re-work

batched/dense/src/KokkosBatched_Rot.hpp
yasahi-hpc: > I do not like calling this ArgTrans, would ArgConj be better since that is actually the only option you have here: regular apply or conjugate apply? Also it then probably makes sense to default it on `NoTranspose`?

That is true.
I just use the boolean to specify the conj operation.

yasahi-hpc:
Thank you for the review @lucbv 
I have updated based on your comments
Can I have another review

lucbv (approved):

PR #2968, opened 2026-02-25
generate_eti: A new script for generating boilerplate for new kernels

jgfouca:
Main change: creates a new script generate_eti and uses it to generate stubs for Cholesky factorization (lapack::potrf). All the lapack changes (except a couple Cmake changes) were done exclusively with generate_eti via this command:

`./scripts/generate_eti lapack potrf 'ci_uplo,0dii_n,2dsio_A,0dii_lda'`

The goal was to run generate_eti and be able to build kokkos-kernels without having to change any C++ code.

Other changes:
1) A minor cleanup of lapack/CMakeLists.txt
2) Delete some old, unused stuff from scripts
3) Update scripts/README
4) Adds a test for generate_eti since it is a fairly complex program

NOTE: This PR is a stub-only implementation of potrf and it's test.


lapack/impl/KokkosLapack_potrf_impl.hpp
```diff
@@ -0,0 +1,26 @@
...
+namespace KokkosLapack {
+namespace Impl {
+
+// Implementation struct for potrf
+template <class AViewType>
+struct PotrfImpl {
+  // TODO: Add your implementation here
+  static void potrf(const char uplo[], const int& n, AViewType& A, const int& lda) {
+    // TODO: Implement your kernel here
+    // Kokkos::abort("KokkosLapack::potrf: Not yet implemented");
```
jgfouca, line 20: Normally, this abort would be uncommented since this function has not been implemented, but that would break CI testing for this PR.

scripts/generate_eti
```diff
@@ -0,0 +1,977 @@
...
+            cpp_type = "const char[]"
+            template_param = None
+        elif dim == 0:
+            # Scalar parameter (real/complex or integer)
+            if dtype == 'i':
+                cpp_type = "int"
+            else:
+                cpp_type = "ScalarType"
+            template_param = None  # Will use view's value_type
+        elif dim == 1:
```
lucbv: For `dim == 1`, `dim == 2` and higher order cases, I do not believe the `cpp_type` makes complete sense. Why same for 1 and 2 but different otherwise? Don't we have MatrixType in some cases and VectorType in others?

scripts/generate_eti
```diff
@@ -0,0 +1,977 @@
...
+
+###############################################################################
+def build_template_params_list(params):
+###############################################################################
+    """Build C++ template parameter list from parsed params."""
+    template_params = [p for p in params if p.template_param is not None]
+    if template_params:
+        return ', '.join([f"class {p.template_param}" for p in template_params])
+    else:
+        return "class DummyType = void"  # For functions with no view parameters
```
lucbv: Not sure why we would go with `template <class DummyType = void>` instead of no template parameter list but hopefully we catch that case later on...

scripts/generate_eti
```diff
@@ -0,0 +1,977 @@
...
+    param_list_parts = []
+    for p in params:
+        if p.is_char:
+            param_list_parts.append(f"const char {p.name}[]")
+        elif p.is_scalar:
+            # Use cpp_type directly for int scalars, otherwise use view's value_type
+            if p.cpp_type == "int":
+                param_list_parts.append(f"const {p.cpp_type}& {p.name}")
+            elif first_view:
+                param_list_parts.append(f"const typename {first_view.cpp_type}::value_type& {p.name}")
```
lucbv: This surely will be wrong with complex types where we may or may not want the magnitude type and not the actual value_type. Note that you cannot guess this mechanically, this is a case by case choice based on mathematical understanding of the function...

scripts/generate_eti
```diff
@@ -0,0 +1,977 @@
...
+            else:
+                # Fallback if no views (should be rare)
+                param_list_parts.append(f"const double& {p.name}")
```
lucbv: This is obviously not always double and actually not always const...

scripts/generate_eti
```diff
@@ -0,0 +1,977 @@
...
+        return False
+
+    impl_dir      = package_root / "impl"
+    src_dir       = package_root / "src"
+    tpls_dir      = package_root / "tpls"
+    unit_test_dir = package_root / "unit_test"
+    eti_cpp_dir   = package_root / "eti" / "generated_specializations_cpp" / function_lower
+    eti_hpp_dir   = package_root / "eti" / "generated_specializations_hpp"
+
+    # Create directories if they don't exist
```
lucbv: Nope, let us not create new directory structure via script, if this is necessary I want someone to do it manually

lapack/unit_test/Test_Lapack_potrf.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include <gtest/gtest.h>
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Random.hpp>
+#include <KokkosLapack_potrf.hpp>
+#include <KokkosKernels_TestUtils.hpp>
+
+namespace Test {
+
+template <class AViewType, class Device>
+void impl_test_potrf(int N) {
```
lucbv, line 13: This does not test anything... cannot be included in this PR

lapack/unit_test/Test_Lapack_potrf.hpp
```diff
@@ -0,0 +1,110 @@
...
+  // using ScalarType = typename AViewType::value_type;
+  // using AT = KokkosKernels::ArithTraits<ScalarType>;
+  // using MagnitudeType = typename AT::mag_type;
+
+  // const MagnitudeType eps = AT::epsilon() * 1000;
+  // const MagnitudeType max_val = 10.0;
+
+  view_stride_adapter<AViewType> A("A", N, N);  // TODO: Adjust dimensions
+  const char uplo[] = "N";                      // TODO: Set appropriate value
```
lucbv: So uplo is for upper or lower which we denote by `L`, `l`, `U` or `u`. `N` does not make sense and should actually be leading to some runtime failure of the test.

jgfouca:
Bump.

lapack/CMakeLists.txt
lucbv, line 1: I would refrain from this change, it is out of scope and makes the PR longer and harder to review.

lapack/tpls/KokkosLapack_potrf_tpl_spec_decl.hpp
lucbv, line 1: No TPL implementation?

lucbv (changes requested):
How are the potrf tests supposed to do anything useful if there is no implementation anywhere? Are you trying to do the PR for the script or for the new LAPACK kernel or both?
Also I would not do additional changes beyond those required in this PR, for instance the macro for LAPACK kernels instantiation should be its own PR which could easily merge before this one actually...

lapack/tpls/KokkosLapack_potrf_tpl_spec_decl.hpp
jgfouca: Correct, I was thinking this PR would be all stubs for now, just demonstrating the tool, but I can go ahead and add a TPL impl.

lapack/unit_test/Test_Lapack_potrf.hpp
jgfouca: Yes, impl is a just a stub, so there's nothing to test. I can go ahead and fill these out in this PR if you want.

lapack/unit_test/Test_Lapack_potrf.hpp
jgfouca: Ah thanks, this was a typo.

scripts/generate_eti
jgfouca: Good catch. The AI was hallucinating a bit here. It seems simplest to just always do `${name}ViewType` regardless of dimension, so I will change this.

scripts/generate_eti
jgfouca: Ok, I will change these to check exists.

scripts/generate_eti
jgfouca: Yeah, this is probably the best we can do here without further expanding the argument syntax.

scripts/generate_eti
jgfouca: Good catch, there's a few places here where it's assuming const and shouldn't be. I will fix.

scripts/generate_eti
jgfouca: I think functions with no Views should at least take a ScalarType template arg, so I changed this.

jgfouca:
@lucbv , I think I made all the changes you wanted. The original idea for this PR is that it would just have stubs for the potrf impl and test (to help keep the PR smaller). Since this PR is already huge, I'm thinking I should go ahead and make a TPL impl and test.

lucbv:
@jgfouca it is okay to not have the tpl or impl but the there should be none of the generate headers. I understand why you put them in but we cannot merge it as is. I would recommend removing them if you want to keep things simple. I have seen how it works and could go back to previous commits to have a look at them again. I would argue against making the PR even longer. If you have already started and made some commits, I can show you how to get that sorted out using git tomorrow

jgfouca:
@lucbv , sounds good. I will split this PR into two PRs:
1) Only scripts stuff
2) potrf with full test and impl

PR #2969, opened 2026-02-26
Benchmark - Spmv: fixing a small bug in input parsing

lucbv:
Without the proper parsing for --mode=manual it is hard to do simple runs with matrix market files. This should remedy that problem.

### DIFF
--- a/benchmarks/sparse/KokkosSparse_spmv_benchmark.cpp
+++ b/benchmarks/sparse/KokkosSparse_spmv_benchmark.cpp
@@ -63,7 +63,7 @@ void parse_inputs(int argc, char** argv, spmv_parameters& params) {
   for (int i = 1; i < argc; ++i) {
     if (benchmark::check_arg_int(i, argc, argv, "-n", params.N)) {
       ++i;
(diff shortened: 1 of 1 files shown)


benchmarks/sparse/KokkosSparse_spmv_benchmark.cpp
```diff
@@ -64,7 +64,7 @@ void parse_inputs(int argc, char** argv, spmv_parameters& params) {
     if (benchmark::check_arg_int(i, argc, argv, "-n", params.N)) {
       ++i;
     } else if (benchmark::check_arg_str(i, argc, argv, "--mode", params.alg)) {
```
cwpearson: probably need to change `params.alg` here too? (to `params.mode`?)

cwpearson (changes requested):

benchmarks/sparse/KokkosSparse_spmv_benchmark.cpp
lucbv: Yeah...

cwpearson (approved):

PR #2970, opened 2026-02-26
AT2 - Kumquat: disabling the CI on V100

lucbv:
Since we have maintenance on this machine for an undefinite amount of time we are disabling these checks for the time being.


cwpearson (approved):
Sad

PR #2990, opened 2026-03-09
Fix for illegal instruction error on AMD gfx1000+ in debug (-O0) builds

RL-S:
When trying a debug build on my RX 9070 XT, I got the output

```
error: Illegal instruction detected: 
Invalid dpp_ctrl value: wavefront shifts are not supported on GFX10+
```

which matches a known ROCm issue: https://github.com/ROCm/ROCm/issues/5826
The workaround proposed there is to change `-O0` to `-Og`, which is exactly what this PR does: For HIP builds with `Kokkos_ARCH=AMD_GFX1000` or higher, the CMAKE_CXX_FLAGS_DEBUG are appended with `-0g -g`.

I hope this workaround becomes obsolete soon, as this should really be fixed in ROCm, rather than in downstream code.

### DIFF
--- a/cmake/kokkos_backends.cmake
+++ b/cmake/kokkos_backends.cmake
@@ -16,3 +16,5 @@ check_kokkos_backend(OPENMPTARGET)
 check_kokkos_backend(CUDA)
 check_kokkos_backend(HIP)
 check_kokkos_backend(SYCL)
+
+include(cmake/kokkoskernels_backendFixes.cmake)
--- a/cmake/kokkoskernels_backendFixes.cmake
+++ b/cmake/kokkoskernels_backendFixes.cmake
@@ -1,0 +1,21 @@
+
+# On recent AMD GPUs (gfx1000+), the following error can occur 
+# for builds with optimisation level 0 (-O0, like debug builds):
+#   error: Illegal instruction detected: 
+#   Invalid dpp_ctrl value: wavefront shifts are not supported on GFX10+
+# The ROCm team is aware, but an upstream fix may be far off.
+# See https://github.com/ROCm/ROCm/issues/5826
+# The recommended workaround is to replace -O0 with -Og,
+# but since that can negatively affect debugging in some instances
+# (see, for example, https://stackoverflow.com/q/63386189),
+# we restrict this change specifically to the affected architectures.
+if(KOKKOS_ENABLE_HIP)
+	if(Kokkos_ARCH MATCHES "AMD_GFX([0-9]+).*")
+		# the issue affects "gfx10+" according to the error message
+		if(CMAKE_MATCH_1 GREATER 1000)
+			message(STATUS "Detected gfx10+ arch (${Kokkos_ARCH}).")
+			message(STATUS "Appending -Og -g to debug build flags (see https://github.com/ROCm/ROCm/issues/5826).")
+			set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Og -g")
+		endif()
+	endif()
+endif()


cwpearson:
Thanks @RL-S!

@lucbv would this be more suitable in Kokkos core?

RL-S:
> Thanks @RL-S!
> 
> @lucbv would this be more suitable in Kokkos core?

If so, then it probably doesn't work via `CMAKE_CXX_FLAGS_DEBUG`, but maybe something like

```
target_compile_options(Kokkos::kokkos INTERFACE 
	$<$<CONFIG:Debug>:-Og;-g>
)
```

Let me know which project you find more appropriate. Kokkos itself compiles fine without the fix, but any downstream projects could benefit from putting it into core.

lucbv:
Hi @RL-S could you give us a bit more information, we are still trying to decide if Kokkos Core or Kernels is the right place for a fix.
Can you send us your configuration for both Core and Kernels and let us know if the issue arrises in a unit-test or your own code?

lucbv:
@RL-S after discussing it with the development team for Kokkos Core and Kernels, we will add an entry about this problem in our documentation under "known issues" explaining how to work around this problem. 

We will not accept the change you proposed, since the issue does not come from the code base and we want each user to be able to set the flags the way they want/need for each combination of toolchain and gpu themselves.

RL-S:
Oh. I thought part of Kokkos' purpose was to abstract away the platform-dependent flags, as without this fix (wherever it's applied), basically any downstream code is no longer portable.

It's an imperfect workaround, and temporarily* including it feels iffy, I get that. I would also understand the argument that it creates unexpected/unintuitive behaviour if -O0 is overridden with -Og. But I thought of Kokkos as setting many flags for the user for the purpose of code portability.

*whatever that means with ROCm, could still be years I guess.

I'd now have to include this fix in every project that uses Kokkos or KokkosKernels. So does every other project that wants the portability of their code to include modern AMD GPUs. It's from that standpoint that I don't understand the decision.

If user choice is your concern, why not wrap this in an opt-out cmake option? Like,

```
option(
    Kokkos_DONT_USE_GFX1000_DEBUG_WORKAROUND 
    "Don't override -O0 flags on HIP builds for GFX1000+ with -0g -g"
    OFF
)
```

and then include that in the "Known issues". Or have it opt-in if you prefer.

EDIT: My point is that on the affected hardware, there just isn't that much "user choice". You either implement some workaround, or your code won't compile.

JBludau:
@RL-S yes, Kokkos tries to abstract things away for users. But we also need to keep our own code-base clean.

As ROCm lists the fix in https://rocm.docs.amd.com/en/7.11.0-preview/about/release-notes.html#clang-illegal-instruction-error-on-radeon-gpus, and it only concerns debug builds, we decided to favor keeping our code base maintainable. My personal opinion is that downstream projects should also point at ROCm and not silently work around the issue as this might only lead to confusion why it is necessary for some libraries but not others. Also we want ROCm to fix it

RL-S:
Fair enough. As a library consumer, it's indeed hard to appreciate the decision of the ROCm team to leave this issue unaddressed. ROCm itself would definitely be the best place to address this.

Btw, I have now confirmed that this issue is present in Kokkos Core: with Kokkos_ENABLE_TESTS=ON, the same "illegal instruction" error occurs. That means four out of four libraries that I tested experience the issue. Maybe I'm misunderstanding what you mean with confusion and "some libraries but not others", because I've had a hit rate of 100%.

Rombur:
I can reproduce the error in Kokkos Core on a Radeon PRO W6800 (gfx1030).

JBludau:
> That means four out of four libraries that I tested experience the issue. Maybe I'm misunderstanding what you mean with confusion and "some libraries but not others", because I've had a hit rate of 100%.

I was referring to "If e.g. Kokkos decides to fix the flag but other libraries don't": users might be confused why a program linking both libraries together works but programs that rely only on libraries that are not fixing the flag don't.
This is likely to create false-positive bug reports.
In contrast: if no library fixes the flag, then users that have an affected gpu will see it in all their programs, which indicates it is something hardware/compiler specific.

Does that make sense?

RL-S:
Yep, now I understand the intention. 
From my perspective, I can't really agree with it, because it means I have to apply the same fix to multiple projects, plus forwarding it to KokkosKernels, instead of having a single addition in the shared dependency. Even as _opt-in_, it would still be useful. And by putting the workaround into a separate file, the changes to Kokkos could be a single inclusion line. Then I don't think the maintenance burden is too high - particularly because I'm already an interested party willing to chip in.
But well, my perspective is just that.

lucbv:
@RL-S we are still looking into this issue and think that we found which calls are triggering the compiler issue. Currently we are looking into a more localized approach to resolve this. Would you be okay opening an issue with Kokkos Core to track the resolution?

Issue #2966, opened 2026-02-25
Lapack test failure


cgcgcg:
Seen in Trilinos PR testing in `KokkosKernels_lapack_serial_MPI_1`:
```
[ RUN      ] serial.svd_double
../../runner/_work/Trilinos/Trilinos/packages/kokkos-kernels/test_common/KokkosKernels_TestUtils.hpp:127: Failure
Expected: ((double)AT1::abs(val1 - val2)) <= ((double)AT3::abs(tol)), actual: 2.98383e-14 vs 1.93175e-14
[  FAILED  ] serial.svd_double (17 ms)
```
https://sems-cdash-son.sandia.gov/cdash/tests/31120732#

Trilinos detected LAPACK as
```
Processing enabled external package/TPL: LAPACK (enabled explicitly, disable with -DTPL_ENABLE_LAPACK=OFF)
-- LAPACK_LIBRARY_NAMES='lapack lapack_win32 openblas'
-- TPL_LAPACK_LIBRARIES='-L/home/runner/spack/opt/spack/linux-x86_64/netlib-lapack-3.12.1-z2x6vobqgq33k4kxoolji4hxa6zo7i4x/lib;-llapack;-lgfortran;-lgomp'
```
(See https://sems-cdash-son.sandia.gov/cdash/builds/1116566)

I believe that the LAPACK library has not changed from previous runs, nor has the code for KokkosKernels.

ndellingwood:
@cgcgcg is the failure consistent or spurious?

ndellingwood:
The matrix for the test is initialized with random values https://github.com/kokkos/kokkos-kernels/blob/1cae3f796a5f5997cb344f07703f89a63d6aba61/lapack/unit_test/Test_Lapack_svd.hpp#L453-L456 , maybe the tolerance on the check needs adjustment?

cgcgcg:
I reran and did not see it anymore.

brian-kelley:
@cgcgcg #3008 should take care of this. I didn't have to increase the tolerance directly, but instead used a larger interval where absolute tolerance is used.

lucbv:
I asked @ndellingwood to cherry-pick this PR for 5.1.1, it is fairly low risk as it only involves a test and it will be nice to have it run in Trilinos' CI to observe if the test gets more stable with this. If so we can close this issue

PR #3008, opened 2026-03-24
LAPACK SVD test 

brian-kelley:
SVD tests: expand the interval around 0 where we switch to an absolute tolerance, instead of relative tolerance, for checking that the triple product U*S*V^t equals the original matrix. Values close to 0 in this product can be the result of summing larger magnitude values that happen to cancel out, so they are less accurate according to relative tolerance.

With my system's host LAPACK, I replicated a failure like #2966 after about ~1000 repetitions of the test. After this change, I ran for ~200k repetitions and it didn't fail, so this should be a reasonable fix.


lucbv (approved):

lucbv:
@ndellingwood if there is time to include this in the release it might not be a bad idea as it should stabilize the test in Trilinos

PR #3020, opened 2026-04-01
Remove KOKKOS_ENABLE_CUDA_LAMBDA checks

lucbv:
Kokkos always enables cuda lambda when the CUDA backend is enabled so we do not need to check for it anymore.
This was used in batched tests and graph coarsening.


ndellingwood (approved):
Thanks!

ndellingwood:
@lucbv should we cherry-pick to rc 5.1.1?

lucbv:
No, 5.1.1 still has the `Kokkos_ENABLE_CUDA_LAMBDA` option so we should hold off on this change and only include it when Kokkos Core removes the option and macro from its release.

PR #3032, opened 2026-04-13
Improve matrix market and batched sparse file readers to handle bad I/O

janciesko:
Validates file opens, header/size lines, and stream extraction before using parsed dimensions or indexing into empty containers. This removes undefined behavior (e.g. empty vectors, empty strings, failed fgets) that segfaults when inputs are missing or malformed.

### DIFF
--- a/perf_test/test_crsmatrix.cpp
+++ b/perf_test/test_crsmatrix.cpp
@@ -2,6 +2,8 @@
 // SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
 #include <cstdio>
 #include <cstddef>
+#include <array>
+#include <memory>
 
 #include <ctime>
 #include <cstring>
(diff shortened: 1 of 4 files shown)


perf_test/test_crsmatrix.cpp
```diff
@@ -44,15 +49,33 @@ int SparseMatrix_MatrixMarket_read(const char* filename, OrdinalType& nrows, Ord
     newFileSize = fread (buffer,1,fileSize,file);*/
 
   while (line[0] == '%') {
-    fgets(line, 511, file);
+    if (!fgets(line, 511, file)) {
```
lucbv: That might just be my lack of knowledge but any reasons behind the choice of 511 characters?

lucbv (commented):
Just a question, this should be good to merge after clang-format is applied

perf_test/test_crsmatrix.cpp
janciesko: We read 511 chars +`\0` to fit into the static array `line`.  This is archaic, should I modernize and reuse this PR?

lucbv (approved):

PR #3034, opened 2026-04-15
KokkosBatched_Util: allow DynRankView in rank checks

japlews:
Conditionally disable some compile-time checking

### DIFF
--- a/batched/KokkosBatched_Util.hpp
+++ b/batched/KokkosBatched_Util.hpp
@@ -23,6 +23,7 @@
 
 #include "Kokkos_Complex.hpp"
 
+#include "Kokkos_DynRankView.hpp"
 #include "KokkosKernels_config.h"
 #include "KokkosKernels_Macros.hpp"
 #include "KokkosKernels_SimpleUtils.hpp"
@@ -677,9 +678,14 @@ KOKKOS_INLINE_FUNCTION void fma_bounds_check(ViewType v, SizeType m, SizeType n,
 namespace Impl {
 template <typename ViewType>
 KOKKOS_INLINE_FUNCTION int get_extent_int(const ViewType &v, const int r) {
-  static_assert(Kokkos::is_view_v<ViewType>, "KokkosBatched: ViewType is not a Kokkos::View.");
-  constexpr std::size_t V_rank = ViewType::rank();
-  static_assert(V_rank <= 2, "KokkosBatched: ViewType must have rank 0, 1 or 2.");
+  // Only allow View and DynRankView objects
+  if constexpr (Kokkos::is_view_v<ViewType>) {
+    static_assert(ViewType::rank() == 2, "KokkosBatched: ViewType must be rank 0, 1, or 2");
+  } else {
+    static_assert(Kokkos::is_dyn_rank_view_v<ViewType>, "KokkosBatched: ViewType must be a DynRankView or a View");
+  }
+
+  const std::size_t V_rank = v.rank();
 
   if (r == 0) {
     int V_extent_0 = V_rank < 1 ? 1 : v.extent_int(0);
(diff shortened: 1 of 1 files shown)


japlews:
@lucbv

batched/KokkosBatched_Util.hpp
```diff
@@ -677,9 +678,14 @@ KOKKOS_INLINE_FUNCTION void fma_bounds_check(ViewType v, SizeType m, SizeType n,
 namespace Impl {
 template <typename ViewType>
 KOKKOS_INLINE_FUNCTION int get_extent_int(const ViewType &v, const int r) {
-  static_assert(Kokkos::is_view_v<ViewType>, "KokkosBatched: ViewType is not a Kokkos::View.");
-  constexpr std::size_t V_rank = ViewType::rank();
-  static_assert(V_rank <= 2, "KokkosBatched: ViewType must have rank 0, 1 or 2.");
+  // Only allow View and DynRankView objects
+  if constexpr (Kokkos::is_view_v<ViewType>) {
+    static_assert(ViewType::rank() == 2, "KokkosBatched: ViewType must be rank 0, 1, or 2");
```
lucbv, line 683:
```suggestion
    static_assert(ViewType::rank() <= 2, "KokkosBatched: ViewType must be rank 0, 1, or 2");
```

batched/KokkosBatched_Util.hpp
```diff
@@ -694,9 +700,14 @@ KOKKOS_INLINE_FUNCTION int get_extent_int(const ViewType &v, const int r) {
 
 template <typename ViewType>
 KOKKOS_INLINE_FUNCTION std::size_t get_stride(const ViewType &v, const int r) {
-  static_assert(Kokkos::is_view_v<ViewType>, "KokkosBatched: ViewType is not a Kokkos::View.");
-  constexpr std::size_t V_rank = ViewType::rank();
-  static_assert(V_rank <= 2, "KokkosBatched: ViewType must have rank 0, 1 or 2.");
+  // Only allow View and DynRankView objects
+  if constexpr (Kokkos::is_view_v<ViewType>) {
+    static_assert(ViewType::rank() == 2, "KokkosBatched: ViewType must be rank 0, 1, or 2");
```
lucbv, line 705:
```suggestion
    static_assert(ViewType::rank() <= 2, "KokkosBatched: ViewType must be rank 0, 1, or 2");
```

batched/KokkosBatched_Util.hpp
```diff
@@ -677,9 +678,14 @@ KOKKOS_INLINE_FUNCTION void fma_bounds_check(ViewType v, SizeType m, SizeType n,
...
+  } else {
+    static_assert(Kokkos::is_dyn_rank_view_v<ViewType>, "KokkosBatched: ViewType must be a DynRankView or a View");
+  }
+
+  const std::size_t V_rank = v.rank();
```
lucbv, line 688: Should this be the dynamic rank when using a dyn_rank_view?
In which case we should define the variable on line 682 and then assign it the `if constexpr` expression.

batched/KokkosBatched_Util.hpp
```diff
@@ -694,9 +700,14 @@ KOKKOS_INLINE_FUNCTION int get_extent_int(const ViewType &v, const int r) {
...
```
lucbv, line 710: Same question as above

lucbv (changes requested):
We should have a look at how the dynamic view is implemented and see if it returns extents for the dynamic ranks or static ranks only?

batched/KokkosBatched_Util.hpp
japlews: Hmm, I like this pattern better because we don't have to repeat ourselves, `V_rank` can be `const`, and it works for both `DynRankView` and `View`. I'm following the pattern in `KokkosBatched_Trsm_Team_Impl.hpp`.

ndellingwood:
@lucbv when this PR is finalized should it be include with 5.1.1?

batched/KokkosBatched_Util.hpp
lucbv: Yeah, after re-reading the documentation on dyn_rank_view, I agree you are right and we should keep this pattern.

lucbv:
Closing this since #3036 merged

PR #3049, opened 2026-04-28
Add batched rotg

yasahi-hpc:
This PR aims at adding a batched interface for [rotg](https://www.netlib.org/lapack/explore-html/d7/dc5/group__rotg_gaafa91c51f75df6c3f2182032a221c2db.html#gaafa91c51f75df6c3f2182032a221c2db). 
Discussing points

1. Should we update the `b` value? It is updated only by (s/d)rotg but not with (c/z)rotg. It is unclear this value can be used in reality. 
2. Should we manage more carefully the cases for extremely small `a` and `b`?  This kernel can overflow in the current implementation.

- [x] Refactor `rotg_impl` under blas
- [x] Add batched interface `Rotg`. `SerialRotg` and `TeamRotg` are not added because this kernel works only on scalars without any parallelization
- [x] Tests introduced under batched (under blas tests are commented out for some reason)

### DIFF
--- a/batched/dense/impl/KokkosBatched_Rotg_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Rotg_Impl.hpp
@@ -12,13 +12,15 @@ namespace KokkosBatched {
 /// \brief invoke Rotg for scalars
 /// constructs a plane rotation such that
 /// [[c,         s],  * [[a],  = [[r],
-///  [-conjg(s), c]]    [b],]     [0]]
+///  [-conjg(s), c]]     [b]]     [0]]
 ///
 /// \tparam SViewType 0-D View type containing a nonconst real or complex scalar
 /// \tparam MViewType 0-D View type containing a nonconst real/magnitude-typed scalar
 ///
 /// \param[in,out] a: On entry, the scalar a. On exit, it is overwritten by r, the nonzero element of the rotated
-/// vector. \param[in,out] b: The scalar b (real or complex). \param[out] c: cosine of the rotation (real scalar)
+/// vector.
+/// \param[in,out] b: The scalar b (real or complex).
+/// \param[out] c: cosine of the rotation (real scalar)
 /// \param[out] s: sine of the rotation (real or complex scalar)
 ///
 template <class SViewType, class MViewType>
--- a/batched/dense/impl/KokkosBatched_Rotg_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Rotg_Impl.hpp
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
+
+#ifndef KOKKOSBATCHED_ROTG_IMPL_HPP_
+#define KOKKOSBATCHED_ROTG_IMPL_HPP_
+
+#include <KokkosBlas_util.hpp>
+#include <KokkosBatched_Util.hpp>
+#include "KokkosBlas1_rotg_impl.hpp"
+
+namespace KokkosBatched {
+/// \brief invoke Rotg for scalars
+/// constructs a plane rotation such that
+/// [[c,         s],  * [[a],  = [[r],
+///  [-conjg(s), c]]    [b],]     [0]]
+///
+/// \tparam SViewType 0-D View type containing a nonconst real or complex scalar
+/// \tparam MViewType 0-D View type containing a nonconst real/magnitude-typed scalar
+///
+/// \param[in,out] a: On entry, the scalar a. On exit, it is overwritten by r, the nonzero element of the rotated
+/// vector. \param[in,out] b: The scalar b (real or complex). \param[out] c: cosine of the rotation (real scalar)
+/// \param[out] s: sine of the rotation (real or complex scalar)
+///
+template <class SViewType, class MViewType>
+KOKKOS_INLINE_FUNCTION int Rotg::invoke(const SViewType &a, const SViewType &b, const MViewType &c,
+                                        const SViewType &s) {
+  static_assert(Kokkos::is_view_v<SViewType>, "KokkosBatched::rotg: SViewType is not a Kokkos::View.");
+  static_assert(Kokkos::is_view_v<MViewType>, "KokkosBatched::rotg: MViewType is not a Kokkos::View.");
+  static_assert(SViewType::rank() == 0, "KokkosBatched::rotg: SViewType must have rank 0.");
+  static_assert(MViewType::rank() == 0, "KokkosBatched::rotg: MViewType must have rank 0.");
+  static_assert(std::is_same_v<typename SViewType::value_type, typename SViewType::non_const_value_type>,
+                "KokkosBatched::rotg: SViewType must have non-const value type.");
+  static_assert(std::is_same_v<typename MViewType::value_type, typename MViewType::non_const_value_type>,
+                "KokkosBatched::rotg: MViewType must have non-const value type.");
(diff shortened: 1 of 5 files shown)


blas/impl/KokkosBlas1_rotg_impl.hpp
lucbv, line 1: While we should consider concepts for modernization of the library, I do not think that the require clause is the right way to go here, instead I suggest the simpler
```
if constexpr KokkosKernels::ArithTraits<Scalar>::is_complex {

} else {

}
```
It does not introduce a new API overload that needs to be kept in think and it achieves more code reuse so less maintenance.

batched/dense/impl/KokkosBatched_Rotg_Impl.hpp
```diff
@@ -0,0 +1,45 @@
...
+/// vector.
+/// \param[in,out] b: The scalar b (real or complex).
+/// \param[out] c: cosine of the rotation (real scalar)
+/// \param[out] s: sine of the rotation (real or complex scalar)
+///
+template <class SViewType, class MViewType>
+KOKKOS_INLINE_FUNCTION int Rotg::invoke(const SViewType &a, const SViewType &b, const MViewType &c,
+                                        const SViewType &s) {
+  static_assert(Kokkos::is_view_v<SViewType>, "KokkosBatched::rotg: SViewType is not a Kokkos::View.");
+  static_assert(Kokkos::is_view_v<MViewType>, "KokkosBatched::rotg: MViewType is not a Kokkos::View.");
```
lucbv, line 30: Why use views instead of scalars here? Is there a use case that demonstrates it?

lucbv (changes requested):
I think we can simplify the implementation quite a bit and I am also not sure about the API, i.e. rank-0 views vs scalar? Maybe both could be provided?

batched/dense/impl/KokkosBatched_Rotg_Impl.hpp
yasahi-hpc: I think it is better to make them `Views`
If they are values, we would have something like this. It is really easy to forget `&` and I am not sure how to check it from inside the API

```C++
  KOKKOS_INLINE_FUNCTION
  void operator()(const int k) const {
    auto& a = m_a(k);
    auto& b = m_b(k);
    auto& c = m_c(k);
    auto& s = m_s(k);
    KokkosBatched::Rotg::invoke(a, b, c, s);
  }
```

blas/impl/KokkosBlas1_rotg_impl.hpp
yasahi-hpc: We can do that.
I implemented this way because the APIs are different for real and complex as commented in the PR description.
Do you think we should still update `b` value only for the real case?

batched/dense/impl/KokkosBatched_Rotg_Impl.hpp
lucbv: Okay

blas/impl/KokkosBlas1_rotg_impl.hpp
lucbv: Isn't it the behavior of BLAS? What would we update `b` to?

yasahi-hpc: It is.
`b` is updated to `z` where `z` satisfies the following formula.

This allows c and s to be reconstructed from z as follows:
```
   If z = 1, set c = 0, s = 1.
   If |z| < 1, set c = sqrt(1 - z**2) and s = z.
   If |z| > 1, set c = 1/z and s = sqrt( 1 - c**2).
```

The reason why I think `z` value is less important is that 

1. `b` is not updated by the accurate version of [dlartg](https://www.netlib.org/lapack/explore-html/da/dd3/group__lartg_ga86f8f877eaea0386cdc2c3c175d9ea88.html#ga86f8f877eaea0386cdc2c3c175d9ea88) which is the accurate version of drotg
2. `b` is not updated by the complex version

yasahi-hpc: cuBLAS supports the original behavior
https://docs.nvidia.com/cuda/cublas/index.html?highlight=rotg#cublas-t-rotg

Maybe we should keep the original behavior?

If the behavior of `b` is different, I prefer the separate implementations for `real` and `complex`.
If you have a strong preference to merge them, I can do that though

yasahi-hpc: Merged version would look like

```C++
template <class Scalar, class Magnitude>
KOKKOS_INLINE_FUNCTION void rotg_impl(Scalar* KOKKOS_RESTRICT a, Scalar* KOKKOS_RESTRICT b,
                                      Magnitude* KOKKOS_RESTRICT c, Scalar* KOKKOS_RESTRICT s) {
  using mag_type     = typename KokkosKernels::ArithTraits<Scalar>::mag_type;
  const Scalar one   = KokkosKernels::ArithTraits<Scalar>::one();
  const Scalar zero  = KokkosKernels::ArithTraits<Scalar>::zero();
  const mag_type anorm = Kokkos::abs(*a);
  const mag_type bnorm = Kokkos::abs(*b);

  if (bnorm == zero) {
    *c = KokkosKernels::ArithTraits<Magnitude>::one();
    *s = zero;
    if constexpr (!KokkosKernels::ArithTraits<Scalar>::is_complex) {
      *b = zero;
    }
    return;
  }

  if (anorm == zero) {
    *c = KokkosKernels::ArithTraits<Magnitude>::zero();
    *a = Scalar(bnorm);
    if constexpr (KokkosKernels::ArithTraits<Scalar>::is_complex) {
      *s = Kokkos::conj(*b) / bnorm;   
    } else {
      *s = one;
      *b = one;
    }
    return;
  }

  const mag_type scale = anorm + bnorm;
  if constexpr (KokkosKernels::ArithTraits<Scalar>::is_complex) {
    const mag_type norm  = scale * Kokkos::hypot(anorm / scale, bnorm / scale);
    const Scalar unit_a  = *a / anorm;
    *c                   = anorm / norm;
    *s                   = unit_a * Kokkos::conj(*b) / norm;
    *a                   = unit_a * norm;
  } else {
    const Scalar sign  = anorm > bnorm ? *a : *b;
    Scalar norm        = scale * Kokkos::hypot(*a / scale, *b / scale);
    norm               = Kokkos::copysign(norm, sign);
  
    *c = *a / norm;
    *s = *b / norm;
    if (anorm > bnorm) {
      *b = *s;
    } else if (*c != zero) {
      *b = one / *c;
    } else {
      *b = one;
    }
    *a = norm;
  }
}
```

lucbv: Thanks for the clarification, then we can keep it separate as you like although from a user perspective the API will be the same. Regarding over/under flow, we can improve things especially for the complex version where norms are computed. I recommend doing it in a second PR as the scope is a bit different and it will be easier to review that way!

lucbv (approved):

blas/impl/KokkosBlas1_rotg_impl.hpp
yasahi-hpc: Thank you for approval.
I fully agree that over/under flow should be addressed in another PR. 
Maybe we can just implement [DLARTG](https://www.netlib.org/lapack/explore-html/da/dd3/group__lartg_ga86f8f877eaea0386cdc2c3c175d9ea88.html#ga86f8f877eaea0386cdc2c3c175d9ea88) and [ZLARTG](https://www.netlib.org/lapack/explore-html/da/dd3/group__lartg_ga1e266e78a9a8585a1fc67f2d23badc64.html#ga1e266e78a9a8585a1fc67f2d23badc64).

lucbv (approved):

Issue #3035, opened 2026-04-15
tranpose_matrix for complex values sparse matrices


cgcgcg:
For complex-valued sparse matrices it would be useful to have the option to not only transpose the matrix but also take the conjugate of all matrix entries. We could conjugate the values view after the transpose, but combining everything into one kernel launch would be preferable.

@japlews

lucbv:
Shall we close this since #3052 merged?

cgcgcg:
Closed via #3052. Follow-up via https://github.com/trilinos/Trilinos/pull/15201.

PR #3052, opened 2026-04-29
Add option to conjugate values to sparse matrix transpose

cgcgcg:
Adds an optional parameter to the sparse matrix transpose that allows to conjugate the values at the same time. The default behavior does not change. 

#3035

Also see https://github.com/trilinos/Trilinos/pull/15201

### DIFF
--- a/sparse/src/KokkosSparse_Utils.hpp
+++ b/sparse/src/KokkosSparse_Utils.hpp
@@ -4,6 +4,7 @@
 #define KOKKOSKERNELS_SPARSEUTILS_HPP
 #include <vector>
 
+#include "KokkosKernels_ArithTraits.hpp"
 #include "Kokkos_Core.hpp"
 #include "KokkosKernels_SimpleUtils.hpp"
 #include "KokkosKernels_IOUtils.hpp"
@@ -230,7 +231,8 @@ void kk_create_bsr_from_bsr_formatted_point_crs(int block_size, size_t num_rows,
 }
 
 template <typename in_row_view_t, typename in_nnz_view_t, typename in_scalar_view_t, typename out_row_view_t,
-          typename out_nnz_view_t, typename out_scalar_view_t, typename MyExecSpace, bool transpose_values>
+          typename out_nnz_view_t, typename out_scalar_view_t, typename MyExecSpace, bool transpose_values,
+          bool conjugate_values>
 struct TransposeMatrix {
   struct CountTag {};
   struct FillTag {};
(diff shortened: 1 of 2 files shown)


sparse/src/KokkosSparse_Utils.hpp
```diff
@@ -306,7 +309,11 @@ struct TransposeMatrix {
 
                              t_adj(pos) = row_index;
                              if constexpr (transpose_values) {
-                               t_vals(pos) = vals[adjind];
+                               if constexpr (!conjugate_values) {
+                                 t_vals(pos) = vals[adjind];
+                               } else {
+                                 t_vals(pos) = KAT::conj(vals[adjind]);
+                               }
```
lucbv: I am slightly more in favor of this since it makes is a little bit more straight forwards
```suggestion
                               if constexpr (conjugate_values) {
                                 t_vals(pos) = KAT::conj(vals[adjind]);
                               } else {
                                 t_vals(pos) = vals[adjind];
                               }
```

sparse/unit_test/Test_Sparse_Transpose.hpp
```diff
@@ -87,6 +87,36 @@ void testTranspose(int numRows, int numCols, bool doValues) {
                             ExactCompare<size_type, values_t>(input_mat.values, tt_values), valuesDiffs);
     EXPECT_EQ(size_type(0), valuesDiffs);
   }
+
+  using KAT = KokkosKernels::ArithTraits<scalar_t>;
+  if (doValues && KAT::isComplex) {
+    // sort the non-conjugated transpose
+    KokkosSparse::sort_crs_matrix<exec_space, c_rowmap_t, entries_t, values_t>(t_rowmap, t_entries, t_values);
+    auto j = KAT::sqrt(-KAT::one());
```
lucbv: While it's good that the sqrt(-1) works, I would rather do the following to have zero on the real part and one on the imaginary part.
```suggestion
    auto j = scalar_t(0, 1);
```

lucbv (changes requested):
Some small clean-up

cgcgcg:
@lucbv Done.

lucbv (approved):

cgcgcg:
Had to make sure that the constructor for the imaginary unit is actually valid. One more round please..

lucbv:
Let's see if it builds and run on the last 4 AT2 actions, then maybe do clang-format and re-run everything

lucbv (approved):

PR #3055, opened 2026-04-30
KokkosBlas axpby unification refactor

### LINKED ISSUE #2920: axpby native impl does unnecessary allocations and deep copies
In ``KokkosBlas::axpby``, when a and b (coefficients) are both views, it seems like all possible codepaths allocate a new view for each of them and fill or deep_copy to them. This isn't conditioned on whether the input a/b are accessible to the execution space. In general this should not be needed, since our implementation should be OK with a and b having any layout.

brian-kelley:
- avoid allocations and deep_copy in all cases (fix #2920)
- consolidate functors for different a,b type cases
- test uncommon cases only when TEST_ETI_ONLY is OFF
- route uncommon cases (2D X/Y, but with a and/or b being rank-1, extent == 1) through a more generic codepath that is slightly less efficient due to an extra runtime branch in the kernel. This PR also removes this branch from the main axpby_mv kernels in many cases.

All the cases that work before still work (tested locally with TEST_ETI_ONLY=OFF, though it takes ~1.5 hrs to build just the axpby tests!) For both axpby and axpby_mv, a and b can still be:
- scalar
- rank-0 host view
- rank-0 device view
- rank-1 device view of length ``X.extent(1)`` (one coefficient per column)
- rank-1 device view of length 1 (one coefficient for the whole matrix)

This PR adds a doc page for axpby (before there was just axpy) and explains the above stuff.


blas/impl/KokkosBlas1_axpby_unification.hpp
```diff
@@ -0,0 +1,181 @@
...
+#define KOKKOSBLAS1_AXPBY_UNIFICATION_HPP_
+
+#include <Kokkos_Core.hpp>
+#include <KokkosKernels_helpers.hpp>
+
+namespace KokkosBlas {
+namespace Impl {
+
+template <class T>
+constexpr KOKKOS_INLINE_FUNCTION std::enable_if<Kokkos::is_view_v<T>, bool>::type isRank1View() {
```
lucbv: Want to try the new fancy `requires` clause instead of `enable_if`
```suggestion
template <class T>
requires(Kokkos::is_view_v<T>)
constexpr KOKKOS_INLINE_FUNCTION isRank1View() {
```

blas/impl/KokkosBlas1_axpby_unification.hpp
```diff
@@ -0,0 +1,181 @@
...
+  return (int)T::rank == 1;
+}
+
+template <class T>
+constexpr KOKKOS_INLINE_FUNCTION typename std::enable_if<!Kokkos::is_view_v<T>, bool>::type isRank1View() {
```
lucbv:
```suggestion
template <class T>
requires(!Kokkos::is_view_v<T>)
constexpr KOKKOS_INLINE_FUNCTION isRank1View() {
```

lucbv (commented):
Just a quick suggestion to make the code easier to maintain and maybe prettier?

blas/impl/KokkosBlas1_axpby_unification.hpp
brian-kelley: Good suggestion, done.

lucbv (approved):

PR #3059, opened 2026-05-04
Add long double support to MagnitudeScalarType

jgfouca:
With this change, I was able to get the par_ilut benchmark to run with long doubles.

### DIFF
--- a/batched/dense/src/KokkosBatched_Vector.hpp
+++ b/batched/dense/src/KokkosBatched_Vector.hpp
@@ -199,6 +199,10 @@ struct MagnitudeScalarType<double> {
   typedef double type;
 };
 template <>
+struct MagnitudeScalarType<long double> {
+  typedef long double type;
+};
+template <>
 struct MagnitudeScalarType<Kokkos::complex<float>> {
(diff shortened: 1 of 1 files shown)


lucbv (commented):
Can you explain a little bit better why we are making this change? Where is it use and is it being tested?

jgfouca:
@lucbv , sure. When I was experimenting with par_ilut, trying to get it to work with the matrix from TF, I wanted to see if changing the scalar type to "long double" would help the algorithm run (I was overflowing the double type). When I tried this, 99% of KK built fine except for this one MagnitudeScalarType error.

jgfouca:
@lucbv , just realized I didn't really answer your question. It's clear from the build error I got that "long double" is not used or tested regularly in KK. That said, I don't think there's any reason we'd not want to support it? The ETI system does not support "long double", but you can still set up template types that for KK to use long doubles. For example, in the par_ilut benchmark, I just changed my `scalar_t` to `long double` and everything worked (except for this change I had to make to MagnitudeScalarType.

lucbv (approved):

PR #3062, opened 2026-05-05
cuSPARSE spmv/spmm preprocess

brian-kelley:
If we are using a spmv handle for reuse, and using cuSPARSE, call the optional preprocess function to optimize the future spmv/spmm calls. For SpMM this has been available since before CUDA 11.8 (oldest kokkos supported version) and for SpMV, since CUDA 12.4.0.

Also improve the kk_spmv performance test by adding a ``--new-handle`` flag. When this is passed in, a new spmv handle is created on every call to help measure the one-time setup cost of different algorithms. This was used to measure the setup values below.

These results are from Hops H100 with CUDA 12.9. They are all times in seconds and averaged over 1000 repetitions.
SpMV:

|                 | Setup Before | Setup With Preprocess | Reuse Before | Reuse With Preprocess |
|-----------------|--------------|-----------------------|--------------|-----------------------|
| Elasticity3D_60 | 0.00021713   | 0.000216869           | 0.000214423  | 0.000208639           |
| Laplace3D_100   | 5.65189e-05  | 5.65395e-05           | 5.40707e-05  | 4.70214e-05           |
| circuit5M       | 0.000398333  | 0.000398156           | 0.000394993  | 0.000381233           |
| momentum        | 0.000185225  | 0.000184659           | 0.000182083  | 0.000169173           |

SpMM with 4 vectors:
|                 | Setup Before | Setup With Preprocess | Reuse Before | Reuse With Preprocess |
|-----------------|--------------|-----------------------|--------------|-----------------------|
| Elasticity3D_60 | 0.00105127   | 0.00105669            | 0.00103953   | 0.00104006            |
| Laplace3D_100   | 0.000200441  | 0.000204137           | 0.000190426  | 0.000190091           |
| circuit5M       | 0.00180931   | 0.00181741            | 0.00180038   | 0.00180027            |
| momentum        | 0.000687568  | 0.000694562           | 0.000677078  | 0.000677038           |

To summarize this:
- the actual preprocess call costs basically nothing
- for SpMV it can make a nice improvement (almost 15% for Laplace3D_100)
- for SpMM it seems to make very little difference.

I used a printout to check that cusparseSpMM_preprocess is actually getting called in this 4-vector experiment.

### DIFF
--- a/perf_test/sparse/KokkosSparse_kk_spmv.cpp
+++ b/perf_test/sparse/KokkosSparse_kk_spmv.cpp
@@ -36,6 +36,7 @@ struct SPMVBenchmarking {
   std::string filename = "";
   bool flush_cache     = false;
   bool non_reuse       = false;
+  bool new_handle      = false;
 
   // Using the parameters above, run and time spmv where x and y use the given
   // memory layout.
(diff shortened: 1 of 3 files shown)


lucbv (approved):
lgtm

PR #3063, opened 2026-05-05
Changes to get find_package(MKL) working.

japlews:
We need to call `kokkoskernels_find_imported(...)` in order for depedendents to be able to find
  target `MKL::MKL`.


japlews:
@lucbv We are currently using this as a patch in Sierra. @vbrunini FYI.

lucbv:
@japlews can we have a discussion about your use case?
It would be good for me to understand how it breaks and why before making the changes.

PR #3064, opened 2026-05-06
check_api_updates.py: Enhance this tool a bit

jgfouca:
Change list:
* Add argument parsing capability
* Expect piped or redirected input instead of needing a saved file (modified_files.txt)
* Add verbose mode to produce lots more info
* CI will check against GITHUB_BASE_REF instead of origin/develop so that only files changed by the current PR will be processed.
* Add a basic test
* All paths in the src map should have full paths relative to repo root.
* Make sure all file loops process files in sorted order.

One thing we could do is, if no input pipe or redirect is detected, we could run a shell command through python (`git diff --name-only $(git merge-base origin/develop HEAD)`). Let me know if I should do that.


jgfouca:
Apologies for the churn, the GH action git state was a bit different than I expected. It looks like it's working now.

lucbv (approved):

jgfouca:
@lucbv , conflict fixed.

lucbv (approved):

PR #3067, opened 2026-05-07
Use Intel extension to query SYCL device free memory

### LINKED ISSUE #1062: Work around kk_get_free_total_memory not being available for all backends.
Some backends, e.g. `SYCL` don't allow querying the amount of free device space. Nevertheless, `KokkosSPGEMM::compute_num_pool_chunks`, https://github.com/kokkos/kokkos-kernels/blob/0a1740890e84a145d6567768a0102f90a5c26f16/src/sparse/impl/KokkosSparse_spgemm_impl.hpp#L795-L816
and https://github.com/kokkos/kokkos-kernels/blob/0a1740890e84a145d6567768a0102f90a5c26f16/src/sparse/impl/KokkosSparse_gauss_seidel_impl.hpp#L1029-L1045
use it. Thus, using these classes/functions doesn't work for all backends.

brian-kelley:
This way we don't have to call the level-zero API directly.

Resolves #1062.

### DIFF
--- a/common/src/KokkosKernels_ExecSpaceUtils.hpp
+++ b/common/src/KokkosKernels_ExecSpaceUtils.hpp
@@ -182,41 +182,22 @@ inline void kk_get_free_total_memory<Kokkos::HIPManagedSpace>(size_t& free_mem, 
 }
 #endif
 
-// FIXME_SYCL Use compiler extension instead of low level interface when
-// available. Also, we assume to query memory associated with the default queue.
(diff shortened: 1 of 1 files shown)


brian-kelley:
This should be ready for review now. The PV autotest build still has failing tests, but the native SpGEMM that uses ``kk_get_free_total_memory`` is passing.

lucbv (approved):

PR #3081, opened 2026-05-12
Move CPP stuff not derived directly from CMake vars out of KokkosKernels_config.h.in

### LINKED ISSUE #944: Move src/KokkosKernels_Macros.hpp to src/common/KokkosKernels_Macros.hpp

jgfouca:
KokkosKernels_config.h becomes the basic include that clients should include (it was already far more commonly included), not KokkosKernels_Macros.hpp

Fixes #944

### DIFF
--- a/batched/KokkosBatched_Util.hpp
+++ b/batched/KokkosBatched_Util.hpp
@@ -26,7 +26,6 @@
 #include "Kokkos_DynRankView.hpp"
 
 #include "KokkosKernels_config.h"
-#include "KokkosKernels_Macros.hpp"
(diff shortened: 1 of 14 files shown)


JBludau:
hmm I just wanted to say that I think Kokkos core does it the other way round and recommends to not include Kokkos_config.h https://github.com/kokkos/kokkos/blob/102df4a8f668f17688937f3f399f3d458738f134/cmake/KokkosCore_config.h.in#L2

Not sure how much Core and Kernels want to align

jgfouca:
@JBludau , thanks. I think it would be better to do it that way. I will change once I figure out why CI is failing.

PR #3088, opened 2026-05-15
Add batched rotmg

yasahi-hpc:
This PR aims at adding a batched interface for [rotmg](https://www.netlib.org/lapack/explore-html/d3/dd5/group__rotmg_gaebf62f1c90f0829a0a762a7f8918213f.html#gaebf62f1c90f0829a0a762a7f8918213f). 

- [x] Refactor `rotmg_impl` under blas. Fixed the behavior for `d1 < 0`.
- [x] Add batched interface `Rotmg`. `SerialRotmg` and `TeamRotmg` are not added because this kernel works only on scalars without any parallelization
- [x] Tests introduced under batched. Some corner case (`d1 < 0`) is not covered

### DIFF
--- a/batched/dense/impl/KokkosBatched_Rotmg_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_Rotmg_Impl.hpp
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
+
+#ifndef KOKKOSBATCHED_ROTMG_IMPL_HPP_
+#define KOKKOSBATCHED_ROTMG_IMPL_HPP_
+
+#include <KokkosBlas_util.hpp>
+#include <KokkosBatched_Util.hpp>
+#include "KokkosBlas1_rotmg_impl.hpp"
+
+namespace KokkosBatched {
+/// \brief invoke Rotmg
+/// constructs a modified Givens transformation matrix H which zeros the second component of the 2-vector [sqrt(d1)*x1,
+/// sqrt(d2)*y1]**T, where d1, d2, x1, y1 are scalars. The first component of the rotated vector is stored in d1. The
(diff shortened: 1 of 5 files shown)


blas/impl/KokkosBlas1_rotmg_impl.hpp
lucbv, line 1: Can you explain to me why you are removing all the views here?
Typically we use raw pointers so we can mark them as `KOKKOS_RESTRICT` however rotmg computes only four values and cannot be vectorized...
I would recommend keeping the view for this one so we can benefit from ref counting and debug bounds checking.

batched/dense/impl/KokkosBatched_Rotmg_Impl.hpp
```diff
@@ -0,0 +1,66 @@
...
+                    !KokkosKernels::ArithTraits<p_value_type>::is_complex,
+                "KokkosBatched::rotmg: Complex types are not supported for Rotmg.");
+#ifndef NDEBUG
+  // param should include flag, h11, h21, h12, h22
+  if (param.extent_int(0) != 5) {
+    Kokkos::printf("KokkosBatched::rotmg: param must have length 5: param length = %d\n", param.extent_int(0));
+    return 1;
+  }
+#endif
+  KokkosBlas::Impl::rotmg_impl(d1.data(), d2.data(), x1.data(), y1.data(), param.data(), param.stride(0));
```
lucbv: Keep the views here

lucbv (changes requested):
I am pretty much okay with this, I just do not want to change from View to raw pointers in the Blas impl file since we cannot vectorize in this function.

blas/impl/KokkosBlas1_rotmg_impl.hpp
yasahi-hpc: OK
I made them pointers in order to align with `rotg_impl` implementation where the values are represented by pointers. 
I do not have a strong preference, so I can make them views as you want

lucbv (approved):

PR #3098, opened 2026-05-18
nightly - alcf: similar update to olcf PR #3090

lucbv:
Should properly run the "update" phase, eventually the script logic will be updated to not fail the job when one unit-test returns with an error so we can always upload the results of the benchmark tests.


.gitlab/alcf-ctest.cmake
```diff
@@ -0,0 +1,29 @@
+set(CTEST_SOURCE_DIRECTORY $ENV{PWD})
+set(CTEST_BINARY_DIRECTORY $ENV{PWD}/kokkos-kernels_build)
+
+set(CTEST_SITE "ALCF-Polaris")
+set(CTEST_BUILD_NAME "Linux-g++")
+
+set(CTEST_START_WITH_EMPTY_BINARY_DIRECTORY TRUE)
+
+set(CTEST_UPDATE_COMMAND /usr/bin/git)
```
yasahi-hpc, line 9: Why you need this? Do we have any preference in git?

.gitlab/alcf-ctest.cmake
```diff
@@ -0,0 +1,29 @@
...
+			    -B kokkos-kernels_build \
+			    -DCMAKE_INSTALL_PREFIX=kokkos-kernels_install \
+			    -DBUILD_SHARED_LIBS=ON \
+			    -DCMAKE_BUILD_TYPE=Release \
+			    -DCMAKE_VERBOSE_MAKEFILE=OFF \
+			    -DKokkos_ROOT=kokkos_install \
+			    -DKokkosKernels_ENABLE_TESTS:BOOL=ON \
+			    -DKokkosKernels_ENABLE_EXAMPLES:BOOL=ON \
+			    -DKokkosKernels_ENABLE_BENCHMARKS:BOOL=ON \
+			    -DKokkosKernels_RUN_BENCHMARKS:BOOL=ON")
```
yasahi-hpc:
```suggestion
```

May be better to set using `ctest_configure`

.gitlab/alcf-ctest.cmake
```diff
@@ -0,0 +1,29 @@
...
+
+set(CTEST_BUILD_COMMAND "cmake --build $ENV{PWD}/kokkos-kernels/build -j 48")
+
+ctest_start(Nightly)
+ctest_update(SOURCE "${CTEST_SOURCE_DIRECTORY}")
+ctest_configure(BUILD "${CTEST_BINARY_DIRECTORY}")
+ctest_build(BUILD "${CTEST_BINARY_DIRECTORY}")
+ctest_test(BUILD "${CTEST_BINARY_DIRECTORY}")
```
yasahi-hpc:
```suggestion
# ------------------------------------------------------------------------------
# Update source tree
# ------------------------------------------------------------------------------

ctest_update(
  SOURCE "${CTEST_SOURCE_DIRECTORY}"
)

# ------------------------------------------------------------------------------
# Configure project
# ------------------------------------------------------------------------------

ctest_configure(
  BUILD "${CTEST_BINARY_DIRECTORY}"
  SOURCE "${CTEST_SOURCE_DIRECTORY}"
  OPTIONS
    "-DCMAKE_INSTALL_PREFIX=kokkos-kernels_install"
	"-DBUILD_SHARED_LIBS=ON"
	"-DCMAKE_BUILD_TYPE=Release"
	"-DCMAKE_VERBOSE_MAKEFILE=OFF"
	"-DKokkos_ROOT=kokkos_install"
	"-DKokkosKernels_ENABLE_TESTS:BOOL=ON"
	"-DKokkosKernels_ENABLE_EXAMPLES:BOOL=ON"
	"-DKokkosKernels_ENABLE_BENCHMARKS:BOOL=ON"
	"-DKokkosKernels_RUN_BENCHMARKS:BOOL=ON"
  RETURN_VALUE configure_result
)

# ------------------------------------------------------------------------------
# Build project
# ------------------------------------------------------------------------------

ctest_build(
  BUILD "${CTEST_BINARY_DIRECTORY}"
  NUMBER_ERRORS build_errors
  RETURN_VALUE build_result
)

# ------------------------------------------------------------------------------
# Run tests
# ------------------------------------------------------------------------------

ctest_test(
  BUILD "${CTEST_BINARY_DIRECTORY}"
  RETURN_VALUE test_result
)

# ------------------------------------------------------------------------------
# Submit results to CDash
# ------------------------------------------------------------------------------

ctest_submit()

# ------------------------------------------------------------------------------
# Failure handling
# ------------------------------------------------------------------------------

if(configure_result)
  message(FATAL_ERROR "Error during configuration! Exit code: ${configure_result}")
endif()

if(build_result)
  message(FATAL_ERROR "Error during build! Exit code: ${build_result}" with ${build_errors} errors)
endif()

if(test_result)
  message(FATAL_ERROR "Error during testing! Exit code: ${test_result}")
endif()
```

Actually, I am not sure we need to explicitly set `SOURCE` and `BUILD`. It seems that default values are sufficient. See [ctest_build](https://cmake.org/cmake/help/latest/command/ctest_build.html#command:ctest_build) for example

yasahi-hpc (changes requested):
From documentation, we would have something like this. However, I am not quite sure this is correct. So, let's give it a try

.gitlab/alcf-ctest.cmake
lucbv: It failed on my laptop without it so I added it here and we know that on alcf machines that is what we are using.

yasahi-hpc: As long as `git` command is an alias to `/usr/bin/git`, this can be suppressed. But, we can keep this for the moment.

.gitlab/alcf-ctest.cmake
```diff
@@ -0,0 +1,85 @@
...
+set(CTEST_START_WITH_EMPTY_BINARY_DIRECTORY TRUE)
+
+set(CTEST_UPDATE_COMMAND /usr/bin/git)
+
+set(CTEST_BUILD_COMMAND "cmake --build $ENV{PWD}/kokkos-kernels/build -j 48")
+
+ctest_start(Nightly)
+ctest_update(SOURCE "${CTEST_SOURCE_DIRECTORY}")
+ctest_configure(BUILD "${CTEST_BINARY_DIRECTORY}")
+ctest_build(BUILD "${CTEST_BINARY_DIRECTORY}")
```
yasahi-hpc:
```suggestion
```

yasahi-hpc (approved):
Looks good. Please suppress the duplicated commands

yasahi-hpc (approved):

PR #3103, opened 2026-05-18
SYCL: disabling device blas tests

lucbv:
For now we skip {Team,TeamVector,Serial}Gemv to avoid runtime failures. Further investigation is needed to understand what triggers the failure and come up with a permanent fix!


lucbv:
Unfortunately this does not fix everything, it actually hides it a bit but getting this build cleaner will allow us to actually detect new regressions in PRs so I think it's still worth the effort...

blas/unit_test/Test_Blas2_gemv.hpp
```diff
@@ -225,6 +225,23 @@ TEST_F(TestCategory, gemv_double) {
 #if defined(KOKKOSKERNELS_INST_COMPLEX_DOUBLE) || \
     (!defined(KOKKOSKERNELS_ETI_ONLY) && !defined(KOKKOSKERNELS_IMPL_CHECK_ETI_CALLS))
 TEST_F(TestCategory, gemv_complex_double) {
+#if defined(KOKKOS_ENABLE_SYCL)
```
brian-kelley, line 228: Maybe this should be
```
#if defined(KOKKOS_ENABLE_SYCL)
constexpr bool testComplex = !std::is_same_v<typename TestDevice::execution_space, Kokkos::Experimental::SYCL>;
#else
constexpr bool testComplex = true;
#endif

if constexpr(testComplex) {
...
}
else {
    GTEST_SKIP();
}
```
I had to read it carefully to see that the same test cases were in the ``else`` branch on 231 and the ``#else`` on 244.

blas/unit_test/Test_Blas2_gemv_util.hpp
```diff
@@ -90,7 +90,18 @@ static void fill_random_view(
...
 template <class GemvFunc, class ScalarA, class ScalarX, class ScalarY, class Device, class ScalarCoef = void>
 struct GEMVTest {
-  static void run(const char *mode) { run_algorithms<0, typename GemvFunc::algorithms>(mode); }
+  static void run(const char *mode) {
+#if defined(KOKKOS_ENABLE_SYCL)
+    if constexpr (std::is_same_v<typename Device::execution_space, Kokkos::Experimental::SYCL> &&
+                  KokkosKernels::ArithTraits<ScalarA>::is_complex) {
+      GTEST_SKIP();
+    } else {
+      run_algorithms<0, typename GemvFunc::algorithms>(mode);
```
brian-kelley, line 103: Here it's OK since the actual test code is just one line.

blas/unit_test/Test_Blas3_gemm.hpp
```diff
@@ -381,9 +381,19 @@ TEST_F(TestCategory, gemm_double) {
 #if defined(KOKKOSKERNELS_INST_COMPLEX_DOUBLE) || \
     (!defined(KOKKOSKERNELS_ETI_ONLY) && !defined(KOKKOSKERNELS_IMPL_CHECK_ETI_CALLS))
 TEST_F(TestCategory, gemm_complex_double) {
+#if defined(KOKKOS_ENABLE_SYCL)
+  if constexpr (std::is_same_v<typename TestDevice::execution_space, Kokkos::Experimental::SYCL>) {
```
brian-kelley: Might want to un-duplicate it here too

brian-kelley (changes requested):
Especially in gemv, this shouldn't duplicate the test code in the 2 possible branches where we don't ``GTEST_SKIP()``.

blas/unit_test/Test_Blas2_gemv.hpp
lucbv: Yeah, that's easier to read and maintain, I implemented it as `constexpr bool skip_complex` to avoid the negations which I think makes it also easier to understand.

brian-kelley (approved):

brian-kelley (approved):

PR #3104, opened 2026-05-19
Lapack - geqrf: fixing test that assumes Kokkos::Serial

lucbv:
Kokkos::Serial is not always enabled, unlike the Kokkos::HostSpace since we may have Kokkos::OpenMP as host execution space instead. This fix will allow the test to compile correctly and pick the default host execution space, which ever is enabled.


vqd8a (approved):
LGTM!

PR #3105, opened 2026-05-20
Fixing issues of alcf-ctest.cmake

lucbv:
These changes are required to fix issues in the nightly on Polaris


lucbv:
For reference the current CDash failing at all stages

PR #3111, opened 2026-05-21
Host space instantiation

### LINKED ISSUE #2231: trtri undefined references in cuda/11.2.2 build with no eti
Builds (at least with cuda/11.2.2) with no eti fail to link trtri:

```
17:34:43 [ 63%] Linking CXX executable KokkosKernels_lapack_serial
17:34:45 CMakeFiles/KokkosKernels_lapack_serial.dir/backends/Test_Serial_Lapack.cpp.o: In function `int KokkosLapack::trtri<Kokkos::View<double**, Kokkos::LayoutLeft, Kokkos::Serial> >(char const*, char const*, Kokkos::View<double**, Kokkos::LayoutLeft, Kokkos::Serial> const&)':
17:34:45 tmpxft_00007941_00000000-6_Test_Serial_Lapack.cudafe1.cpp:(.text._ZN12KokkosLapack5trtriIN6Kokkos4ViewIPPdJNS1_10LayoutLeftENS1_6SerialEEEEEEiPKcS9_RKT_[_ZN12KokkosLapack5trtriIN6Kokkos4ViewIPPdJNS1_10LayoutLeftENS1_6SerialEEEEEEiPKcS9_RKT_]+0x460): undefined reference to `KokkosLapack::Impl::TRTRI<Kokkos::View<int, Kokkos::LayoutRight, Kokkos::HostSpace, Kokkos::MemoryTraits<1u> >, Kokkos::View<double**, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<1u> >, false, false>::trtri(Kokkos::View<int, Kokkos::LayoutRight, Kokkos::HostSpace, Kokkos::MemoryTraits<1u> > const&, char const*, char const*, Kokkos::View<double**, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<1u> > const&)'
17:34:45 CMakeFiles/KokkosKernels_lapack_serial.dir/backends/Test_Serial_Lapack.cpp.o: In function `int KokkosLapack::trtri<Kokkos::View<Kokkos::complex<double>**, Kokkos::LayoutLeft, Kokkos::Serial> >(char const*, char const*, Kokkos::View<Kokkos::complex<double>**, Kokkos::LayoutLeft, Kokkos::Serial> const&)':
17:34:45 tmpxft_00007941_00000000-6_Test_Serial_Lapack.cudafe1.cpp:(.text._ZN12KokkosLapack5trtriIN6Kokkos4ViewIPPNS1_7complexIdEEJNS1_10LayoutLeftENS1_6SerialEEEEEEiPKcSB_RKT_[_ZN12KokkosLapack5trtriIN6Kokkos4ViewIPPNS1_7complexIdEEJNS1_10LayoutLeftENS1_6SerialEEEEEEiPKcSB_RKT_]+0x460): undefined reference to `KokkosLapack::Impl::TRTRI<Kokkos::View<int, Kokkos::LayoutRight, Kokkos::HostSpace, Kokkos::MemoryTraits<1u> >, Kokkos::View<Kokkos::complex<double>**, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<1u> >, false, false>::trtri(Kokkos::View<int, Kokkos::LayoutRight, Kokkos::HostSpace, Kokkos::MemoryTraits<1u> > const&, char const*, char const*, Kokkos::View<Kokkos::complex<double>**, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<1u> > const&)'
17:34:45 collect2: error: ld returned 1 exit status
```

**Reproducer**: (weaver rhel8)
```bash
bsub -Is -n 1 -q rhel8 -gpu "num=1" bash

source /etc/profile.d/modules.sh
source /projects/ppc64le-pwr9-rhel8/legacy-env.sh

module load cuda/11.2.2/gcc/8.3.1 cmake/3.23.1

${KOKKOSKERNELS_PATH}/cm_generate_makefile.bash --with-cuda --with-serial --compiler=${KOKKOS_PATH}/bin/nvcc_wrapper --arch=Volta70,Power9 --with-cuda-options=enable_lambda --kokkos-path=${KOKKOS_PATH} --kokkoskernels-path=${KOKKOSKERNELS_PATH} --with-scalars='double,complex_double' --with-ordinals=int --with-offsets=int,size_t --with-layouts=LayoutLeft --cxxstandard=17 --no-default-eti

make -j16
```

lucbv:
Fixes #2231 
@JBludau

For background information on the issue being fixed here, please have a look at the discussion in #2231 
In this PR there are two required changes and I am open to additional work as detailed below

**Required changes:**
 - [x] Removing OpenMPTarget from the ETI set, the removal occurred in Kokkos Core and this is just more confusing than helpful to anyone!
 - [x] Modifying the logic to instantiate the HostSpace, the minimum requirement is to turn it on when on Host execution space is being instantiated.
 
**Optional changes:**
 - [x]  Remove [kokkoskernels_eti_memspaces.cmake](https://github.com/kokkos/kokkos-kernels/blob/develop/cmake/kokkoskernels_eti_memspaces.cmake), I am sure I don't need to explain that one...
 - [ ] Align the logic between the device and host spaces instantiation with a common logic that could become a macro?
 - [ ] Remove [kokkos_backends.cmake]() and associated usage in Kokkos Kernels, that one could just spawn its own issue and PR depending on how pervasive it is...


lucbv:
Using the following configuration:
```
cmake \
    -S ${KOKKOSKERNELS_SOURCE} \
    -B ${KOKKOSKERNELS_BUILD} \
    -D CMAKE_CXX_COMPILER:STRING="clang++" \
    -D CMAKE_CXX_STANDARD:STRING="20" \
    -D CMAKE_BUILD_TYPE:STRING="Release" \
    -D Kokkos_ROOT:PATH="${KOKKOS_INSTALL}" \
    -D CMAKE_INSTALL_PREFIX:PATH="${KOKKOSKERNELS_INSTALL}" \
    -D KokkosKernels_ADD_DEFAULT_ETI:BOOL=OFF \
    -D KokkosKernels_INST_ORDINAL_INT:BOOL=ON \
    -D KokkosKernels_INST_OFFSET_INT:BOOL=ON \
    -D KokkosKernels_INST_LAYOUTLEFT:BOOL=ON \
    -D KokkosKernels_INST_FLOAT:BOOL=ON \
    -D KokkosKernels_INST_FLOAT:BOOL=ON \
    -D KokkosKernels_INST_FLOAT:BOOL=ON \
    -D KokkosKernels_ENABLE_TESTS:BOOL=ON
```
the configure stage output looks as follows, when building against a Kokkos Core that has both serial and threads execution spaces enabled.

### Before proposed change
```
Kokkos Kernels ETI Types
   Devices:  
   Scalars:  float
   Ordinals: int
   Offsets:  int
   Layouts:  LayoutLeft
```

### After proposed change
```
Kokkos Kernels ETI Types
   Devices:  <Threads,HostSpace>;<Serial,HostSpace>
   Scalars:  float
   Ordinals: int
   Offsets:  int
   Layouts:  LayoutLeft
```

The new logic turns on instantiation of the HostSpace which we have some device ETI and the build can complete without issues.

> Note: that without any changes to our instantiation logic, the build could be successful if `-D KokkosKernels_INST_MEMSPACE_HOSTSPACE:BOOL=ON` is manually specified at configure time.

cmake/kokkoskernels_eti_devices.cmake
```diff
@@ -104,26 +100,20 @@ if(KOKKOS_ENABLE_SYCL)
...
-    "Whether to pre instantiate kernels for the memory space Kokkos::Experimental::OpenMPTargetSpace.  Disabling this when Kokkos_ENABLE_OPENMPTARGET is enabled may increase build times. Default: ON if Kokkos is OpenMPTarget-enabled, OFF otherwise.")
-
-  if(KOKKOSKERNELS_INST_EXECSPACE_OPENMPTARGET AND KOKKOSKERNELS_INST_MEMSPACE_OPENMPTARGETSPACE)
-    list(APPEND DEVICE_LIST "<OpenMPTarget,OpenMPTargetSpace>")
-  endif()
-
-  if(Trilinos_ENABLE_COMPLEX_DOUBLE AND ((NOT DEFINED CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS) OR (NOT CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS)))
-    message(WARNING
-      "The CMake option CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS is either undefined or OFF.  Please set CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS:BOOL=ON when building with OpenMPTarget and complex double enabled.")
-  endif()
```
JBludau, line 121: This change seems unrelated to the rest to me, but I might not know enough about the interaction with Trilinos here.

cmake/kokkoskernels_eti_devices.cmake
```diff
@@ -104,26 +100,20 @@ if(KOKKOS_ENABLE_SYCL)
...
+#################
+#               #
+#  Host Spaces  #
+#               #
+#################
+
+if(KOKKOS_ENABLE_SERIAL OR KOKKOS_ENABLE_OPENMP OR KOKKOS_ENABLE_THREADS)
```
JBludau: Since we always require in core that at least one host space is enabled, this condition would only be **not** fulfilled for HPX. Is that the only situation that we want to use `KOKKOSKERNELS_ADD_DEFAULT_ETI`?
Would it not make sense to just always have the instantiation for `HostSpace` since it is also used by HPX?

Basically that would translate to "Always instantiate for (Memoryspace) HostSpace if any host backend is enabled"

JBludau (commented):
I think this is the right way to make sure the ETI is created for all enabled host backends.
But I think the logic can be simplified since Kokkos always requires at least one host backend to be enabled and all of them have the same memory space

cmake/kokkoskernels_eti_devices.cmake
lucbv: That's true, I am using this PR to delete some OpenMPTarget references since it was removed from Core. It could have been split in a second PR.

cmake/kokkoskernels_eti_devices.cmake
lucbv: Yes, that's a fair comment, I am updating the PR to enable unconditionally the HostSpace by default, of course that can still be manually set by the user at configure time to override this default.

JBludau (approved):

brian-kelley (approved):
After some discussion we agree that this is the best solution. It is basically making the default behavior for ``INST_MEMSPACE_HOSTSPACE`` consistent with the other device memory spaces. For example, if Kokkos has Cuda enabled then ``INST_MEMSPACE_CUDASPACE`` will also be on by default, even when ``ADD_DEFAULT_ETI=OFF``.

Issue #3073, opened 2026-05-09
[Windows][MPI] Build with shared libs failed with LNK1189


ViNN280801:
## Disclaimer

> Partlitially copied from [Kokkos-Issue-9168](https://github.com/kokkos/kokkos/issues/9168)

## Description

Build **shared** Kokkos libs with occurs linker error. But with **static** everything is ok. And as we can see, the static lib is quite heavy.

<img width="1070" height="542" alt="Image" src="https://github.com/user-attachments/assets/7466ba1e-b105-4e6d-a792-96f386edfdbf" />

<details><summary>More details here👇</summary>

```log
D:\Develop\Trilinos>powershell -ExecutionPolicy ByPass -File .\build.ps1
PowerShell 5.1.26100.8115

========================================================================================
  System info
========================================================================================
Windows : Microsoft Windows 11 Pro (Build 26100)
CPU     : 12th Gen Intel(R) Core(TM) i9-12900H  (20 logical cores)
CMake   : cmake version 3.31.2
MSVC    : 19.44.35211 for x64  (C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\bin\HostX64\x64\cl.exe)
Ninja   : 1.12.1
----------------------------------------------------------------------------------------


========================================================================================
  Windows long path check
========================================================================================
Trilinos ETI file names are very long and often exceed the Windows
MAX_PATH limit (260 chars), causing C1083 build errors.

Long paths: Enabled  (up to ~32767 chars)
No action needed.
----------------------------------------------------------------------------------------


========================================================================================
  Existing build(s) detected
========================================================================================
  [1] D:\Develop\Trilinos\bmpi  (last modified: 2026-05-09 20:14)
       source: D:/Develop/Trilinos
       project: Trilinos
       cmake command (saved):
         cmake `

           # --- Generator, source tree, build directory, configuration type ---
           -G `
           Ninja `
           '-DCMAKE_C_COMPILER=C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.44.35207/bin/HostX64/x64/cl.exe' `
           '-DCMAKE_CXX_COMPILER=C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.44.35207/bin/HostX64/x64/cl.exe' `
           -S `
           D:/Develop/Trilinos `
           -B `
           D:/Develop/Trilinos/bmpi `
[...]

PR #3113, opened 2026-05-24
CMake: Fix for warning C4530, error C1128, error LNK2001 and undefined reference to 'rank()' on Windows

ViNN280801:
## Motivation

Exactly same reasons as in [Kokkos PR-9181](https://github.com/kokkos/kokkos/pull/9181) and [Trilinos PR-15279](https://github.com/trilinos/Trilinos/pull/15279).
This patch adds specific .cmake file for Windows to correctly support SEH and adds `/bigobj` only for CUDA build because of heavy `.cudafe1.cpp` files which are exceeds number of sections `.obj` limit of 16-bit integer.

> In all fairness, I should have started my journey to fully compiling Trilinos on Windows with this, but I already added the `/EHsc` flag to my `build.ps1` because it's standard practice, and I did it automatically, without even thinking about compiling with a pure build without my own flags.

As I said in [Kokkos PR-9181](https://github.com/kokkos/kokkos/pull/9181) and [Trilinos PR-15279](https://github.com/trilinos/Trilinos/pull/15279) clang-cl does not compile code with exceptions but without this specific flag `/EHsc` when MSVC for some reasons allow to compile such code, what in my opinion is wrong permission because of wrong behaviour of program.

## Environment

| Item       | Value                                            |
| ---------- | ------------------------------------------------ |
| OS         | Windows 11 Pro x64, Build 26100                  |
| CPU        | Intel Core i9-12900H (20 logical cores)          |
| CMake      | 3.31.2                                           |
| MSVC       | 19.44.35211 (VS 2022 Community)                  |
| clang-cl   | 21.1.5;x86_64-pc-windows-msvc;thread-model:posix |
| Ninja      | 1.12.1                                           |
| Build type | RelWithDebInfo, shared libs, C++20               |

## Related Issues

- Root issue - [Trilinos Issue-14816](https://github.com/trilinos/Trilinos/issues/14816)
- [Kokkos PR-9181](https://github.com/kokkos/kokkos/pull/9181) and [Trilinos PR-15279](https://github.com/trilinos/Trilinos/pull/15279)
- [Trilinos PR-15279](https://github.com/trilinos/Trilinos/pull/15279)

## Related Links

- [Exception Handling Model](https://learn.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-170)
- [Bigger number of sections in .obj file](https://learn.microsoft.com/en-us/cpp/build/reference/bigobj-increase-number-of-sections-in-dot-obj-file?view=msvc-170)
- [Fatal Error C1128](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1128?view=msvc-170)
- [COFF Object File Format](https://en.wikipedia.org/wiki/COFF)
- [CMake Unity build](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html)

### MSVC (`/bigobj` problem)

- Configuration output: [configuration-output-msvc2022-ninja-cuda.log](https://github.com/user-attachments/files/28183199/configuration-output-msvc2022-ninja-cuda.log)
- Compilation output: [compilation-output-msvc2022-ninja-cuda.log](https://github.com/user-attachments/files/28183204/compilation-output-msvc2022-ninja-cuda.log)
- Compilation output about warnings: [compilation-output-msvc2022-ninja-cuda.log.warn.log](https://github.com/user-attachments/files/28183212/compilation-output-msvc2022-ninja-cuda.log.warn.log)
- Compilation output about errors: [compilation-output-msvc2022-ninja-cuda.log.err.log](https://github.com/user-attachments/files/28183214/compilation-output-msvc2022-ninja-cuda.log.err.log)

### Clang-cl (`/EHsc` problem)

- Configuration output: [configuration-output-clangcl-21.1.5-ninja-cuda.log](https://github.com/user-attachments/files/28183220/configuration-output-clangcl-21.1.5-ninja-cuda.log)
- Compilation output: [compilation-output-clangcl-21.1.5-ninja-cuda.log](https://github.com/user-attachments/files/28183228/compilation-output-clangcl-21.1.5-ninja-cuda.log)
- Compilation output about warnings: [compilation-output-clangcl-21.1.5-ninja-cuda.log.warn.log](https://github.com/user-attachments/files/28183249/compilation-output-clangcl-21.1.5-ninja-cuda.log.warn.log)
- Compilation output about errors: [compilation-output-clangcl-21.1.5-ninja-cuda.log.err.log](https://github.com/user-attachments/files/28183241/compilation-output-clangcl-21.1.5-ninja-cuda.log.err.log)

### If don't want to download logs

image

## Testing

After this change I made clean build, here are the results:

- Configuration output: [configuration-output-clangcl-21.1.5-ninja-cuda.log](https://github.com/user-attachments/files/28184016/configuration-output-clangcl-21.1.5-ninja-cuda.log)
- Compilation output: [compilation-output-clangcl-21.1.5-ninja-cuda.log](https://github.com/user-attachments/files/28184263/compilation-output-clangcl-21.1.5-ninja-cuda.log)
- Compilation output about warnings: [compilation-output-clangcl-21.1.5-ninja-cuda.log.warn.log](https://github.com/user-attachments/files/28184268/compilation-output-clangcl-21.1.5-ninja-cuda.log.warn.log)
- Compilation output about new linking error `__atomic_compare_exchange_16`: [compilation-output-clangcl-21.1.5-ninja-cuda.log.err.log](https://github.com/user-attachments/files/28184273/compilation-output-clangcl-21.1.5-ninja-cuda.log.err.log)

### `__atomic_compare_exchange_16`

After this unsuccessfull build because of:

```log
1845: Sparse_spmv_eti_COMPLEX_DOUBLE_ORDINAL_INT_OFFSET_INT_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp.obj : error LNK2001: unresolved external symbol __atomic_compare_exchange_16
1846: Sparse_spmv_mv_eti_COMPLEX_DOUBLE_ORDINAL_INT_OFFSET_INT_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp.obj : error LNK2001: unresolved external symbol __atomic_compare_exchange_16
1847: Sparse_sptrsv_solve_eti_COMPLEX_DOUBLE_ORDINAL_INT_OFFSET_INT_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp.obj : error LNK2001: unresolved external symbol __atomic_compare_exchange_16
1848: Sparse_gauss_seidel_apply_eti_COMPLEX_DOUBLE_ORDINAL_INT_OFFSET_INT_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp.obj : error LNK2001: unresolved external symbol __atomic_compare_exchange_16
1849: Sparse_spmv_bsrmatrix_eti_COMPLEX_DOUBLE_ORDINAL_INT_OFFSET_INT_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp.obj : error LNK2001: unresolved external symbol __atomic_compare_exchange_16
1850: Sparse_spmv_mv_bsrmatrix_eti_COMPLEX_DOUBLE_ORDINAL_INT_OFFSET_INT_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp.obj : error LNK2001: unresolved external symbol __atomic_compare_exchange_16
1851: Sparse_spmv_eti_DOUBLE_ORDINAL_INT_OFFSET_INT_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp.obj : error LNK2001: unresolved external symbol __atomic_compare_exchange_16
1852: Sparse_spmv_eti_COMPLEX_FLOAT_ORDINAL_INT_OFFSET_INT_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp.obj : error LNK2001: unresolved external symbol __atomic_compare_exchange_16
1853: Blas1_dot_mv_eti_COMPLEX_DOUBLE_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp.obj : error LNK2001: unresolved external symbol __atomic_compare_exchange_16
1854: Blas1_sum_mv_eti_COMPLEX_DOUBLE_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp.obj : error LNK2001: unresolved external symbol __atomic_compare_exchange_16
```

I proposed another solution in commit for MSVC show hint to explicitly use `-A x64` (but by default from Visual Studio 2019 default arch is x64, so, I don't think that someone will face with this error while compiling on MSVC), for clang-cl use `/clang:-mcx16`. And I really don't think that anyone on this planet has x86 with GPUs who wants compile code for HPC purposes.
After this change/commit build was successful.

Also, dear reviewer, I decided to use MSVC just because can't compile Kokkos with clang-cl because of:

```log
D:/Develop/kokkos/bcuda/generated/Kokkos_Version_Info.cpp
-- Kokkos is configured for CMake languages CUDA compilation (using NVIDIA version 12.6.20)
-- Using -std:c++20 for C++20 standard as feature
CMake Error at cmake/kokkos_test_cxx_std.cmake:40 (message):
CMake wants to use -std:c++20 which is not supported by NVCC.
```

and because of I don't think that mix up 2 different ABIs (Kokkos CUDA build on MSVC <==> Kokkos-Kernels CUDA build on Clang-cl) is a good idea.

### 3rd problem

```log
D:\Develop\kokkos-kernels\batched\dense\impl\KokkosBatched_Trsm_Team_Impl.hpp(157): error: identifier "Kokkos::View< ::Kokkos::complex<float>  **,  ::Kokkos::LayoutRight,  ::Kokkos::Device< ::Kokkos::Serial,
::Kokkos::HostSpace> ,  ::Kokkos::MemoryTraits<(unsigned int)3u>  > ::rank" is undefined in device code                                                                                                                                                                       
D:\Develop\kokkos-kernels\batched\dense\impl\KokkosBatched_Trsm_Team_Impl.hpp(231): error: identifier "Kokkos::View< ::Kokkos::complex<float>  **,  ::Kokkos::LayoutRight,  ::Kokkos::Device< ::Kokkos::Serial,  ::Kokkos::HostSpace> ,  ::Kokkos::MemoryTraits<(unsigned int)3u>  > ::rank" is undefined in device code
2 errors detected in the compilation of "D:/Develop/kokkos-kernels/bcuda/sparse/eti/generated_specializations_cpp/spiluk_numeric/Sparse_spiluk_numeric_eti_COMPLEX_FLOAT_ORDINAL_INT_OFFSET_INT_LAYOUTLEFT_EXECSPACE_SERIAL_MEMSPACE_HOSTSPACE.cpp".
```

### 4th problem

The same problem that I described in #3073, so I needed to rebuilt it again using static libraries

image

### 5th problem with undefined `rank()` in device code

The same as in 3rd one.

image

### 6th problem is a **hard blocker** - `LNK1248` (see _Questions No 1_)

image

## Questions

1. What about `LNK1189` I see several options (for #3073 I can propose the same approaches):
        1.1. Split library into parts, for example it would be: [kokkoskernels_core.dll (`/common` dir), kokkoskernels_batched.dll, kokkoskernels_blas.dll, kokkoskernels_lapack.dll, kokkoskernels_sparse.dll, etc. (smth else if I skipped)] - in my opinion is the best way, but it need to refactor and more time for implimentation
        1.2. Use `-DCMAKE_UNITY_BUILD=ON` + `-DCMAKE_UNITY_BUILD_BATCH_SIZE=<N>` - _I myself learned about this quite recently when I began to study the issue of the `LNK1189` linker problem in more depth_; as I understand this is not the best option because of **side effects - violation ODR and harder to isolating errors, but as advantage it has faster builds**. In any case, I would choose this option last.
        1.3. `FATAL_ERROR` + hint message about to prefer static libraries instead of shared - this is the most straightforward but not the best and clearest way
        
> All these prevent wasting time!

2. Does kokkos-kernels needs to restrict compilation on clang-cl as host + device CUDA 12.6 with `/std:c++20`, because of Kokkos restricts it in https://github.com/kokkos/kokkos/blob/b958ea229eb86387ee5c228d3692a036ea821628/cmake/kokkos_test_cxx_std.cmake#L34-L45? Or maybe **vice versa - to allow clang-cl on Kokkos?** (need someone from Kokkos team to dicuss about it)
3. Clang-cl's compilation output shows many of warnings where most of them is about data loss by downcasting from bigger type to lower, for example ` warning: implicit conversion loses integer precision: 'index_type' (aka 'unsigned long long') to 'int' [-Wshorten-64-to-32]`, is this ok? I can propose the solution for all of them in next PRs (if necessary and relevant) smth like this https://github.com/ViNN280801/LumexCore/blob/develop/lumex/core/utility/LumexSafeNumericComparator.hpp, all the documentation is on Russian, but here is minimal example how to use it:

```cpp
#if defined(KOKKOSKERNELS_ENABLE_SAFE_NUMERIC_COMPARATOR_DEMO) && !defined(__CUDA_ARCH__) \
  && !defined(__HIP_DEVICE_COMPILE__)
/// \example Host-only SafeNumericComparator usage (enable with
/// -DKOKKOSKERNELS_ENABLE_SAFE_NUMERIC_COMPARATOR_DEMO).
///
/// ArithTraits<char>::abs() must return char, not int (avoids -Wimplicit-int-conversion
/// at KokkosKernelsAbs). After you have a narrow magnitude, comparing it to a wider
/// threshold is where SafeNumericComparator helps.
inline bool KokkosKernelsExampleMagnitudeExceedsThreshold(char x, int threshold) {
  const char magnitude = integer_abs<char>::abs(x);

  SafeComparator<char, false> safeMagnitude(magnitude);
  return safeMagnitude.safe_greater_equal(threshold);
}

/// Classic SafeNumericComparator case: unsigned char magnitude vs int limit.
/// Naive `unsigned char(200) >= 300` is wrong because 300 wraps to 44 mod 256.
inline bool KokkosKernelsExampleUCharMagnitudeExceedsThreshold(unsigned char x, int limit) {
  const unsigned char magnitude = integer_abs<unsigned char>::abs(x);

  SafeComparator<unsigned char, false> safeMagnitude(magnitude);
  if (safeMagnitude.safe_greater_equal(limit)) {
    return true;
  }
  return safe_greater_equal(magnitude, limit);  // free-function form
}
#endif
```

Or just write minimal template helper exclusively for downcasting and use it in all relevant places, smth like this:

```cpp
#ifndef KOKKOSKERNELS_NARROWCAST_HPP
#define KOKKOSKERNELS_NARROWCAST_HPP

#include <Kokkos_Macros.hpp>
#include <type_traits>
#include <limits>

namespace KokkosKernels {
namespace Impl {

namespace narrow_cast_detail {

template <class Dst, class Src>
struct IsLosslessIntegralCast {
  static constexpr bool value =
      std::is_integral<Dst>::value && std::is_integral<Src>::value &&
      (sizeof(Dst) > sizeof(Src) ||
       (sizeof(Dst) == sizeof(Src) &&
        std::is_signed<Dst>::value == std::is_signed<Src>::value));
};

template <class Dst, class Src>
KOKKOS_FUNCTION typename std::enable_if<
    IsLosslessIntegralCast<Dst, Src>::value, Dst>::type
NarrowCastImpl(Src value) {
  return static_cast<Dst>(value);
}

template <class Dst, class Src>
KOKKOS_FUNCTION typename std::enable_if<
    !IsLosslessIntegralCast<Dst, Src>::value &&
        std::is_integral<Dst>::value && std::is_integral<Src>::value,
    Dst>::type
NarrowCastImpl(Src value) {
  const Dst result = static_cast<Dst>(value);
  if (static_cast<Src>(result) != value) {
    Kokkos::abort("KokkosKernels::narrow_cast: value out of range for destination type");
  }
  return result;
}

}  // namespace narrow_cast_detail

/// @brief Checked integral downcast for device/host code (C++17).
/// @details Uses round-trip identity (GSL narrow pattern). Aborts on overflow.
/// @note Integral types only; use explicit static_cast for floating-point.
template <class Dst, class Src>
KOKKOS_FUNCTION typename std::enable_if<
    std::is_integral<Dst>::value && std::is_integral<Src>::value, Dst>::type
narrow_cast(Src value) {
  return narrow_cast_detail::NarrowCastImpl<Dst>(value);
}

}  // namespace Impl

using Impl::narrow_cast;

}  // namespace KokkosKernels
```

Or more simplier with Boost:

```cpp
#ifndef KOKKOSKERNELS_NARROWCAST_BOOST_HPP
#define KOKKOSKERNELS_NARROWCAST_BOOST_HPP

#include <boost/numeric/conversion/cast.hpp>
#include <type_traits>

namespace KokkosKernels {
namespace Host {
template <class Dst, class Src>
Dst narrow_cast(Src value) {
  static_assert(std::is_arithmetic<Dst>::value && std::is_arithmetic<Src>::value,
                "narrow_cast requires arithmetic types");
  return boost::numeric_cast<Dst>(value);
}
}  // namespace Host
}  // namespace KokkosKernels
#endif
```

But as I see kokkos-kernels does not use boost libraries, so we can prefer pure C++ implementation instead.

This approach can't affect on performance, because `static_assert` performs check entirely at compile-time, so, performance loss is 0%.
But in fact this change will require enormous global refactor (I'm ready for this) and reviewing (it's will be on reviewers, thanks them for theirs expensive time) judging by the size of the compilation warnings log.


ViNN280801:
Forgot to mention about {OpenMP | MPI | CUDA} versions:

| Item       | Value                                            |
| ---------- | ------------------------------------------------ |
| OS         | Windows 11 Pro x64, Build 26100                  |
| CPU        | Intel Core i9-12900H (20 logical cores)          |
| CMake      | 3.31.2                                           |
| MSVC       | 19.44.35211 (VS 2022 Community)                  |
| clang-cl   | 21.1.5;x86_64-pc-windows-msvc;thread-model:posix |
| Ninja      | 1.12.1                                           |
|OpenMP|202011 (aka 5.1)|
|MPI|10.0.12498.5|
|CUDA|12.9; driver: 576.28|
| Build type | RelWithDebInfo, shared libs, C++20               |

### How I checked these versions on Windows

#### OpenMP

```cmd
echo #include ^<iostream^> > test.cpp && echo #include ^<omp.h^> >> test.cpp && echo int main() { std::cout ^<^< _OPENMP ^<^< std::endl; } >> test.cpp
clang-cl -Xclang -fopenmp test.cpp
test.exe
```

### MPI

```cmd
mpiexec /?
```

### CUDA

```cmd
nvidia-smi
```

ViNN280801:
Also, about my mentionned 3rd solution:

> 1.3. `FATAL_ERROR` + hint message about to prefer static libraries instead of shared - this is the most straightforward but not the best and clearest way

It seems to entirely block the CUDA build with message on future, and the best solution is **1.1**

ViNN280801:
gentle ping

JBludau (approved):
Both the code changes and the build system changes look reasonable

cmake/kokkoskernels_windows.cmake
```diff
@@ -0,0 +1,99 @@
+# Detect MSVC ABI once; result persists for the second include (below /bigobj).
+# cl.exe, clang-cl, Intel icl/icx on Windows all use the MSVC C++ ABI.
+# MinGW and Cygwin use the GCC/POSIX model – excluded intentionally.
+if(WIN32 AND NOT CYGWIN AND NOT DEFINED _KK_WINDOWS_MSVC_ABI_CXX)
```
lucbv: I believe `_UPPERCASE` is reserved by the compiler, can you remove the leading `_`
```suggestion
if(WIN32 AND NOT CYGWIN AND NOT DEFINED KK_WINDOWS_MSVC_ABI_CXX)
```

cmake/kokkoskernels_windows.cmake
```diff
@@ -0,0 +1,99 @@
...
+  elseif(CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM")
+    set(_KK_WINDOWS_MSVC_ABI_CXX TRUE)
+  endif()
+endif()
+
+# x64 -------------------------------------------------------------------------
+# 16-byte host atomics need CMPXCHG16B / _InterlockedCompareExchange128 (x64 only).
+# clang-cl: -mcx16 below.  MSVC: Compare_Exchange_MSVC.hpp (no -mcx16).
+# Do not key off -A x64 alone: Ninja + Hostx64/x64/cl.exe is valid without it.
+if(_KK_WINDOWS_MSVC_ABI_CXX AND NOT DEFINED _KK_WINDOWS_X64_CHECK_DONE)
```
lucbv: Same as above

cmake/kokkoskernels_windows.cmake
```diff
@@ -0,0 +1,99 @@
...
+  set(_KK_WINDOWS_X64_CHECK_DONE TRUE)
+endif()
+
+# /EHsc -----------------------------------------------------------------------
+# Required for stack-unwinding semantics in throw/catch on the MSVC ABI.
+# Without it the compiler generates non-unwind tables and <chrono>/<stdexcept>
+# emit C4530; behaviour on actual throws is undefined.
+# https://learn.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model
+# https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4530
+if(_KK_WINDOWS_MSVC_ABI_CXX AND NOT DEFINED _KK_WINDOWS_EHSC_DONE)
```
lucbv: Same as above

cmake/kokkoskernels_windows.cmake
```diff
@@ -0,0 +1,99 @@
...
+endif()
+
+# /bigobj (CUDA only) ---------------------------------------------------------
+# C1128 on the host pass of nvcc (.cudafe1.cpp): template-heavy ETI TUs exceed
+# the default COFF section limit. With COMPILE_AS_CMAKE_LANGUAGE=CUDA, library
+# sources compile as CUDA; cl.exe needs /bigobj via -Xcompiler=/bigobj.
+# https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1128
+# Included a second time (after kokkos_backends.cmake) so KOKKOS_ENABLE_CUDA is set.
+if(_KK_WINDOWS_MSVC_ABI_CXX AND KOKKOS_ENABLE_CUDA AND
+   NOT DEFINED _KK_WINDOWS_BIGOBJ_DONE)
```
lucbv: Same as above

cmake/kokkoskernels_windows.cmake
```diff
@@ -0,0 +1,99 @@
...
+endif()
+
+# -mcx16 -----------------------------------------------------------------------
+# LNK2001 __atomic_compare_exchange_16: clang/clang-cl on Windows uses GCC-style
+# __atomic_* builtins (desul atomics). Without -mcx16, 16-byte compare-exchange
+# (e.g. Kokkos::complex<double>, Dummy16ByteValue reductions) lowers to libatomic
+# which is not linked in the MSVC link environment.
+# MSVC (cl.exe) uses _InterlockedCompareExchange128 instead; -mcx16 is clang-only.
+# Kokkos adds the same flag in cmake/kokkos_arch.cmake for Clang+WIN32.
+if(_KK_WINDOWS_MSVC_ABI_CXX AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
```
lucbv: Same as above

lucbv (changes requested):
Some changes are needed for the macros.
Also there is not testing possible for all of this so I am not quite sure that any of it will stay functional or it if it as proposed?

ViNN280801:
@lucbv, thanks for the review!

> Some changes are needed for the macros.

Done ✅

> Also there is not testing possible for all of this so I am not quite sure that any of it will stay functional or it if it as proposed?

I have tested these changes locally, and they successfully resolve the compilation issues on Windows. However, I did run into a subsequent hard blocker with shared libraries: the `LNK1189` / `LNK1248` linker error, because the final image size exceeds the 4GB limit. I believe that splitting the project into smaller sub-libraries, as proposed in issue #3073, is the correct long-term solution for this.

Regarding the lack of automated CI for this specific Windows configuration (especially with `clang-cl` + CUDA): since we don't have CI coverage for it, how would you prefer to handle the risk of future regressions? 
Would it be acceptable to merge this as a "best-effort" fix for Windows users, or should I look into adding a specific CI job / rely on community manual testing to keep it functional? Maybe I can add windows runner to CI.

lucbv (approved):
Okay, I think this is reasonable for now and until we get proper testing on windows. We could setup a CI build with github runners but reproducing errors and doing development will be hard so still no promise on windows support for now

ViNN280801:
@lucbv, glad it got merged!

Regarding Windows CI: I'd be happy to work on adding a GitHub Actions runner for Windows (MSVC + clang-cl). I already have a working local setup and build scripts, so I have a reasonable baseline to start from. If the team is open to it, I can open a separate PR or issue to discuss the setup. No promises on timeline, but I'm willing to take ownership of this or just provide all the help I can. Also, feel free to tag me if Windows-specific issues come up in the future - I'll do my best to help reproduce and fix them.

batched/KokkosBatched_Util.hpp
```diff
@@ -58,6 +58,17 @@ namespace KokkosBatched {
...
+// nvcc+MSVC device code cannot resolve Kokkos::View::rank via instance call;
+// use ViewType::rank() (compile-time) for fixed-rank View, v.rank() for DynRankView.
+template <typename ViewType>
+KOKKOS_INLINE_FUNCTION size_t view_rank(const ViewType &v) {
+  if constexpr (Kokkos::is_view_v<ViewType>) {
+    return ViewType::rank();
+  } else {
+    return v.rank();
+  }
+}
```
dalg24, line 70: There is `Kokkos::rank` free function with overloads for View and DynRankView

ViNN280801: Oh, sorry, I didn't know about it. Will change it soon.

ViNN280801: see #3174

PR #3120, opened 2026-05-26
Add batched norm

yasahi-hpc:
This PR aims at adding batched norm API.

- [x] Adding a new tag to specify norm type `Norm::L1`, `Norm::L2`, `Norm::LInf`
- [x] Support norm operations [dnrm2](https://www.netlib.org/blas/dnrm2.f90) and [dasum](https://www.netlib.org/blas/dasum.f)
- [x] Add unit-tests 
- [x] Improve unit-testing tools (should be done in a separate [PR](#3121)) 
- [x] Fix LInf Team/TeamVector tests

### DIFF
--- a/batched/dense/impl/KokkosBatched_Nrm_Internal.hpp
+++ b/batched/dense/impl/KokkosBatched_Nrm_Internal.hpp
@@ -9,13 +9,22 @@
 namespace KokkosBatched {
 namespace Impl {
 
-template <typename ValueType, typename NrmValueType>
-KOKKOS_INLINE_FUNCTION NrmValueType l1_norm(const ValueType &x) {
+template <typename NrmValueType, typename ValueType>
+KOKKOS_INLINE_FUNCTION NrmValueType asum(const ValueType &x) {
   if constexpr (KokkosKernels::ArithTraits<ValueType>::is_complex) {
     return KokkosKernels::ArithTraits<NrmValueType>::abs(KokkosKernels::ArithTraits<ValueType>::real(x)) +
            KokkosKernels::ArithTraits<NrmValueType>::abs(KokkosKernels::ArithTraits<ValueType>::imag(x));
   } else {
     return KokkosKernels::ArithTraits<NrmValueType>::abs(x);
+  }
+}
+
+template <typename NrmValueType, typename ValueType>
+KOKKOS_INLINE_FUNCTION NrmValueType square(const ValueType &x) {
+  if constexpr (KokkosKernels::ArithTraits<ValueType>::is_complex) {
+    return KokkosKernels::ArithTraits<ValueType>::conj(x) * x;
+  } else {
+    return x * x;
   }
 }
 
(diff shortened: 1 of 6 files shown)


batched/dense/impl/KokkosBatched_Nrm_Internal.hpp
```diff
@@ -0,0 +1,152 @@
...
+  NrmValueType nrm = 0;
+
+  if constexpr (std::is_same_v<NrmType, Norm::L1>) {
+    for (int i = 0; i < n; ++i) {
+      nrm += l1_norm<ValueType, NrmValueType>(x[i * xs0]);
+    }
+  } else if constexpr (std::is_same_v<NrmType, Norm::L2>) {
+    for (int i = 0; i < n; ++i) {
+      const NrmValueType abs_val = KokkosKernels::ArithTraits<ValueType>::abs(x[i * xs0]);
+      nrm += abs_val * abs_val;
```
lucbv: How about we skip the `abs` value that will compute a square root just to square the results again?
```suggestion
      nrm += x[i * xs0] * Kokkos::conj(x[i * xs0]);
```

batched/dense/impl/KokkosBatched_Nrm_Internal.hpp
```diff
@@ -0,0 +1,152 @@
...
+    Kokkos::parallel_reduce(
+        Kokkos::TeamThreadRange(member, n),
+        [&](const int i, NrmValueType &thread_nrm) { thread_nrm += l1_norm<ValueType, NrmValueType>(x[i * xs0]); },
+        nrm);
+  } else if constexpr (std::is_same_v<NrmType, Norm::L2>) {
+    Kokkos::parallel_reduce(
+        Kokkos::TeamThreadRange(member, n),
+        [&](const int i, NrmValueType &thread_nrm) {
+          const NrmValueType abs_val = KokkosKernels::ArithTraits<ValueType>::abs(x[i * xs0]);
+          thread_nrm += abs_val * abs_val;
```
lucbv: Same as above

batched/dense/impl/KokkosBatched_Nrm_Internal.hpp
```diff
@@ -0,0 +1,152 @@
...
+    Kokkos::parallel_reduce(
+        Kokkos::TeamVectorRange(member, n),
+        [&](const int i, NrmValueType &thread_nrm) { thread_nrm += l1_norm<ValueType, NrmValueType>(x[i * xs0]); },
+        nrm);
+  } else if constexpr (std::is_same_v<NrmType, Norm::L2>) {
+    Kokkos::parallel_reduce(
+        Kokkos::TeamVectorRange(member, n),
+        [&](const int i, NrmValueType &thread_nrm) {
+          const NrmValueType abs_val = KokkosKernels::ArithTraits<ValueType>::abs(x[i * xs0]);
+          thread_nrm += abs_val * abs_val;
```
lucbv: Same as above

lucbv (changes requested):
Just a small question around the computation of the 2-nrom, I think we can improve the performance with a slightly different implementation.

yasahi-hpc:
@lucbv Thank you for the review
I have updated accordingly by introducing a helper function.
Actually, [blas](https://www.netlib.org/lapack/explore-html/d1/d2a/group__nrm2_ga7f9f9febc6dc1836c9f5e7c1aa00b743.html) uses more stable method to avoid overflow. 
Should we do that as well?

lucbv (approved):
I think we can go ahead with this version although if you want to look at the scaling for stability, you can go ahead, that's not a bad idea. That could just be a second PR?

yasahi-hpc:
> I think we can go ahead with this version although if you want to look at the scaling for stability, you can go ahead, that's not a bad idea. That could just be a second PR?

OK, let's make it in a second PR.

PR #3125, opened 2026-06-01
[Docs] batched dot

yasahi-hpc:
- [x] Update dense-index.rst
- [x] Add docs for batched dot
- [x] Add a minimal working example which is embedded in docs
- [x] Minor fix in `rotmg` documentation
- [x] Described in the [deprecation page](https://kokkos.org/kokkos-kernels/docs/deprecation_page.html)


lucbv (changes requested):
Since we have marked some functions as deprecated, can you update [this page](https://kokkos.org/kokkos-kernels/docs/deprecation_page.html) as well?

lucbv (approved):
Thanks for the changes, looks good!

PR #3129, opened 2026-06-02
Turning off failure in ctest on Aurora

lucbv:
Currently the pipeline fails when running ctest and we are not planning on fixing all the tests just now but we would like to push the results of the benchmarks after ctest runs.

### DIFF
--- a/.gitlab/aurora-ctest.cmake
+++ b/.gitlab/aurora-ctest.cmake
@@ -27,9 +27,12 @@ set(CTEST_CONFIGURE_COMMAND "cmake -S ${CTEST_SOURCE_DIRECTORY} \
 
 set(CTEST_BUILD_COMMAND "cmake --build ${CTEST_BINARY_DIRECTORY} --parallel")
 
+set(build_error 0)
+set(test_error 0)
+
 ctest_start(Nightly)
 ctest_update(SOURCE "${CTEST_SOURCE_DIRECTORY}")
 ctest_configure(BUILD "${CTEST_BINARY_DIRECTORY}")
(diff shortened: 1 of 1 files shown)


.gitlab/aurora-ctest.cmake
```diff
@@ -27,9 +27,12 @@ set(CTEST_CONFIGURE_COMMAND "cmake -S ${CTEST_SOURCE_DIRECTORY} \
...
 
+set(build_error 0)
+set(test_error 0)
+
 ctest_start(Nightly)
 ctest_update(SOURCE "${CTEST_SOURCE_DIRECTORY}")
 ctest_configure(BUILD "${CTEST_BINARY_DIRECTORY}")
-ctest_build(BUILD "${CTEST_BINARY_DIRECTORY}")
-ctest_test(BUILD "${CTEST_BINARY_DIRECTORY}")
+ctest_build(BUILD "${CTEST_BINARY_DIRECTORY}" CAPTURE_CMAKE_ERROR ${build_error})
```
masterleinad, line 36: Do you actually want to ignore build errors? https://my.cdash.org/builds/3575312/build seems to have the same warnings as https://my.cdash.org/builds/3580836/build but not errors so it's not clear to me if that error is real or not.

lucbv: I am on Aurora and launching a build to see if I can reproduce this error, but if it is a "fake error" then capturing the return value might be reasonable? I believe that it will still show up as an error on the dashboard just won't be returned to the pipeline process?

lucbv:
Latest build seems to be fine especially after some of the fixes for experimental namespace and gtest casting so let's close this!

PR #3130, opened 2026-06-02
Add scaled l2 norm computation

yasahi-hpc:
This PR introduces a new class for batched norm computation.
Based on the [older blas implementation](https://www.netlib.org/lapack/lapack-3.1.1/html/dznrm2.f.html), l2 norm can be computed without overflow/underflow.

- [x] Add a new class `ScaledL2`
- [x] Add an unit test to show that `ScaledL2` gives correct results without overflow while it gives the same results as `L2`

### DIFF
--- a/batched/dense/src/KokkosBatched_Nrm.hpp
+++ b/batched/dense/src/KokkosBatched_Nrm.hpp
@@ -15,8 +15,7 @@ namespace KokkosBatched {
 /// \tparam NrmType: one of Norm::L1, Norm::L2, Norm::LInf, Norm::ScaledL2
 template <typename NrmType>
 struct SerialNrm {
-  static_assert(std::is_same_v<NrmType, Norm::L1> || std::is_same_v<NrmType, Norm::L2> ||
-                    std::is_same_v<NrmType, Norm::LInf> || std::is_same_v<NrmType, Norm::ScaledL2>,
+  static_assert(is_norm_v<NrmType>,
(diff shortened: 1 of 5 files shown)


batched/dense/src/KokkosBatched_Nrm.hpp
```diff
@@ -10,13 +10,14 @@ namespace KokkosBatched {
...
 /// No nested parallel_for is used inside of the function.
-/// \tparam NrmType: one of Norm::L1, Norm::L2, Norm::LInf
+/// \tparam NrmType: one of Norm::L1, Norm::L2, Norm::LInf, Norm::ScaledL2
 template <typename NrmType>
 struct SerialNrm {
   static_assert(std::is_same_v<NrmType, Norm::L1> || std::is_same_v<NrmType, Norm::L2> ||
-                    std::is_same_v<NrmType, Norm::LInf>,
-                "KokkosBatched::SerialNrm: NrmType must be one of Norm::L1, Norm::L2, Norm::LInf");
+                    std::is_same_v<NrmType, Norm::LInf> || std::is_same_v<NrmType, Norm::ScaledL2>,
+                "KokkosBatched::SerialNrm: NrmType must be one of Norm::L1, Norm::L2, Norm::LInf, Norm::ScaledL2");
```
lucbv, line 19: Maybe write a templated helper function in `KokkosBatched_Util.hpp` to check if a template is valid norm?

lucbv (changes requested):
Just a small change to simplify the code

yasahi-hpc:
@lucbv Thanks. I added a new trait accordingly

lucbv (approved):

PR #3131, opened 2026-06-02
Sparse - SpMV: updating MKL SYCL gemv call

lucbv:
@masterleinad 
The current call is deprecated in modern mkl version so this update should remove warnings and potentially errors that are popping up on Aurora.


masterleinad (approved):

lucbv:
The build on Aurora did not report any errors with the updated mkl API... one more thing would be to check when the new API was introduced and the old one deprecated. 
I will push one more commit to actually guard the change correctly and use the appropriate API based on the MKL version as identified in the documentation [here](https://www.intel.com/content/www/us/en/developer/articles/release-notes/onemkl/2025.html#inpage-nav-1-3)

yasahi-hpc (approved):
Looks good
Is this the reference for new API?
https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2026-0/oneapi-mkl-sparse-set-csr-data.html

I am very surprised by the old API without `nnz` argument.

lucbv:
Yes, I believe you have the correct documentation link above, I had just pointed at the deprecation announcement in my previous comment.

PR #3135, opened 2026-06-06
Benchmark - Blas1: result of nrminf on host

lucbv:
It appears norm inf does not allow the results to be stored on device. This seems odd but for the time being let's just fix the benchmark.
This was tested on blake H100 node and works fine, no more segfault.

### DIFF
--- a/docs/source/API/blas/blas1_nrminf.rst
+++ b/docs/source/API/blas/blas1_nrminf.rst
@@ -70,4 +70,4 @@ output:
 
 .. code::
 
-   X_nrm: 300 Expected: 300
+   X_nrm: 50.011 Expected: 50.011
--- a/docs/source/API/blas/blas1_nrminf.rst
+++ b/docs/source/API/blas/blas1_nrminf.rst
@@ -63,7 +63,7 @@ Type Requirements
 Example
 =======
 
-.. literalinclude:: ../../../../example/wiki/blas/KokkosBlas1_wiki_nrminf.cpp
+.. literalinclude:: ../../../../example/docs/blas/KokkosBlas1_docs_nrminf.cpp
   :language: c++
 
 output:
--- a/docs/source/API/blas/blas1_nrminf.rst
+++ b/docs/source/API/blas/blas1_nrminf.rst
@@ -1,0 +1,73 @@
+KokkosBlas::nrminf
+##################
+
+Defined in header: :code:`KokkosBlas1_nrminf.hpp`
+
+.. code:: c++
+
+  template <class execution_space, class XVector, typename std::enable_if<Kokkos::is_execution_space_v<execution_space>, int>::type = 0>
+  typename KokkosKernels::Details::InnerProductSpaceTraits<typename XVector::non_const_value_type>::mag_type
+  nrminf(const execution_space& space, const XVector& x);
+
+  template <class XVector>
+  typename KokkosKernels::Details::InnerProductSpaceTraits<typename XVector::non_const_value_type>::mag_type
+  nrminf(const XVector& x);
+
+  template <class execution_space, class RV, class XMV>
+  void nrminf(const execution_space& space, const RV& R, const XMV& X,
+              typename std::enable_if<Kokkos::is_view_v<RV>, int>::type = 0);
+
+  template <class RV, class XMV>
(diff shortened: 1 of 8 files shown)


yasahi-hpc (approved):
I do not know why, but the results in `nrminf` seem to be stored in `HostView`

1. May be better to check the consistency with other norms. If not consistent, we may fix the behavior later
2. Documentation should also be updated?

ndellingwood (approved):
Thanks for adding the documentation as well

docs/source/API/blas/blas1_nrminf.rst
```diff
@@ -0,0 +1,73 @@
...
+
+  - ``RV::rank == XMV::rank - 1``
+  - ``std::is_same_v<typename RV::value_type, typename RV::non_const_value_type> == true``
+  - ``std::is_same_v<typename RV::value_type, typename KokkosKernels::Details::InnerProductSpaceTraits<typename XMV::non_const_value_type>::mag_type> == true``
+  - ``RV::memory_space == Kokkos::HostSpace``
+
+Example
+=======
+
+.. literalinclude:: ../../../../example/wiki/blas/KokkosBlas1_wiki_nrminf.cpp
```
ndellingwood: The example below is in `example/docs/blas/KokkosBlas1_docs_nrminf.cpp` , I'm not sure if you want to update the path here, or move the example file to match this wiki path?

From the docs build:

```
` /Users/runner/work/kokkos-kernels/kokkos-kernels/kokkos-kernels/docs/source/API/blas/blas1_nrminf.rst:66: WARNING: Include file '/Users/runner/work/kokkos-kernels/kokkos-kernels/kokkos-kernels/example/wiki/blas/KokkosBlas1_wiki_nrminf.cpp' not found or reading it failed [docutils]`
```

ndellingwood:
```suggestion
.. literalinclude:: ../../../../example/docs/blas/KokkosBlas1_docs_nrminf.cpp
```

if you opt to update the path here

lucbv: Yep just got it done : )

docs/source/API/blas/blas1_nrminf.rst
```diff
@@ -0,0 +1,73 @@
...
+  nrminf(const XVector& x);
+
+  template <class execution_space, class RV, class XMV>
+  void nrminf(const execution_space& space, const RV& R, const XMV& X,
+              typename std::enable_if<Kokkos::is_view_v<RV>, int>::type = 0);
+
+  template <class RV, class XMV>
+  void nrminf(const RV& R, const XMV& X, typename std::enable_if<Kokkos::is_view_v<RV>, int>::type = 0)
+
+Computes the norm inf (maximum absolute value) of the X vector(s) and returns or stores the result in R.
```
yasahi-hpc, line 23: Can we put the term `L_\infty` here?

docs/source/API/blas/blas1_nrminf.rst
```diff
@@ -0,0 +1,73 @@
...
+
+1. returns the largest magnitude in ``X`` and fences the ``space`` instance.
+2. returns the largest magnitude in ``X`` and fences the default instance of ``typename XVector::execution_space``.
+3. returns the largest magnitude in each column of ``X``, stores it in the corresponding entry of ``R`` and fences the ``space`` instance.
+4. returns the largest magnitude in each column of ``X``, stores it in the corresponding entry of ``R`` and fences the default instance of ``typename XVector::execution_space``.
+
+The result (returned or stored value) is undefined if ``X`` has no entries.
+
+Note: For 3. and 4. the return view is expected to be templated on `Kokkos::HostSpace`, the function will fail with a segfault at runtime otherwise.
```
yasahi-hpc, line 32: This behavior is a bit confusing and may be fixed in a future PR. I would at least prefer to make assertion compile time.

```suggestion
.. note::

For 3. and 4. the return view is expected to be templated on `Kokkos::HostSpace`, the function will fail with a segfault at runtime otherwise.
```

yasahi-hpc (approved):
Mostly good
Assuming this is a temporary solution, all my points can be addressed in a subsequent PR

lucbv:
Yes, my goal is to address these other issues in a subsequent PR, it might involve relaxing the benchmark constraint just introduced for the time being.

PR #3149, opened 2026-06-11
Improve performance of RCB and diagonal blocks extraction with device implementations

vqd8a:
This PR is to improve performance of RCB and the associated diagonal blocks extraction utility:
- RCB: replaced the host implementation with a new implementation where all the tasks run on GPU, including finding the most elongated dimension for partitioning, bisecting partition on the most elongated dimension, shuffling coordinates,  generating ```perm```, ```reverse_perm``` vectors
- Extract diagonal blocks: added a new function ```kk_extract_diagonal_blocks_crsmatrix_with_rcb(...)``` that first permutes the original matrix using RCB ordering information and then extracts the diagonal blocks on GPU

When tested with Aria matrix and coordinates (89310 x 89310, nnz = 1,343,982, number of coordinates: 89310), the new RCB is ~2x faster and the new diagonal blocks extraction for 8 blocks is ~19x faster.

Note: streaming implementations can be considered in the future for further performance improvement.


vqd8a:
@lucbv This PR is ready. Please take a look. Thanks.

graph/impl/KokkosGraph_RCB_impl.hpp
```diff
@@ -21,278 +28,315 @@ struct FillOneIncrementFunctor {
...
-    if (val < lminmax.min_val) lminmax.min_val = val;
-    if (val > lminmax.max_val) lminmax.max_val = val;
+template <typename view_type, typename ordinal_t>
+struct ColumnMinMaxFunctor {
+  using scalar_type = typename view_type::value_type;
+  using layout_type = typename view_type::array_layout;
+  using device_type = typename view_type::device_type;
+  using internal_view_type =
+      Kokkos::View<const scalar_type **, layout_type, device_type, Kokkos::MemoryTraits<Kokkos::RandomAccess>>;
+  using value_type = MinMaxPair<scalar_type>[];
```
lucbv: Just a stylistic point of view but I think `value_type` is going to lead to confusing in this context, better to be more explicit so no misunderstanding happens in maintenance down the road
```suggestion
  using minmax_type = MinMaxPair<scalar_type>*;
```

graph/impl/KokkosGraph_RCB_impl.hpp
```diff
@@ -21,278 +28,315 @@ struct FillOneIncrementFunctor {
...
+  internal_view_type A;
+  ordinal_t value_count;
+
+  ColumnMinMaxFunctor(const view_type &A_, const ordinal_t &num_cols_) : A(A_), value_count(num_cols_) {}
+
+  KOKKOS_INLINE_FUNCTION void init(value_type dst) const {
+    for (ordinal_t c = 0; c < value_count; c++) {
+      dst[c].min_val = std::numeric_limits<scalar_type>::infinity();
+      dst[c].max_val = -std::numeric_limits<scalar_type>::infinity();
```
lucbv: I have a few issues with this;

1. infinity() is not guaranteed to be defined for all numeric types, only float, double and long double. This could be UB for half_t and is zero for integral types...
2. do we know if -infinity() is actually well defined?
3. why not use the Kokkos overloads that are portable?

@dalg24 do you have any knowledge on 1. and 2. ?

lucbv (changes requested):
Not done with this put sending this early feedback, I'm still reading through `KokkosGraph_RCB_impl.hpp`

sparse/src/KokkosSparse_Utils.hpp
```diff
@@ -2454,6 +2596,173 @@ void kk_extract_diagonal_blocks_crsmatrix_with_rcb_sequential(
...
+    } else {
+      // A_nrows >= 1
+      if ((n_blocks < 1) || (A_nrows < n_blocks)) {
+        std::ostringstream os;
+        os << "The number of diagonal blocks (" << n_blocks
+           << ") should be >=1 and <= the number of rows of the matrix A (" << A_nrows << ")";
+        throw std::runtime_error(os.str());
+      }
+
+      if (static_cast<ordinal_type>(std::pow(2, static_cast<int>(std::log2(n_blocks)))) != n_blocks) {
```
lucbv: Cant this be done by checking that only one bit is non-zero? See `has_single_bti` in [Kokkos_BitManipulation.hpp](https://kokkos.org/kokkos-core-wiki/API/core/numerics/bit-manipulation.html)

sparse/src/KokkosSparse_Utils.hpp
```diff
@@ -2454,6 +2596,173 @@ void kk_extract_diagonal_blocks_crsmatrix_with_rcb_sequential(
...
+      if (static_cast<ordinal_type>(reverse_perm_rcb.extent(0)) != A_nrows) {
+        std::ostringstream os;
+        os << "The size of the reverse permutation array (" << reverse_perm_rcb.extent(0)
+           << ") must be equal to the number of rows of the matrix A (" << A_nrows << ')';
+        throw std::runtime_error(os.str());
+      }
+
+      // 1. Permute the original matrix using RCB ordering
+      out_row_map_type row_map_rcb(Kokkos::view_alloc(Kokkos::WithoutInitializing, "row_map_rcb"), A_nrows + 1);
+      out_entries_type entries_rcb(Kokkos::view_alloc(Kokkos::WithoutInitializing, "entries_rcb"), A.nnz());
```
lucbv: Why did you bother to create a copy of `A.numRows()` but not `A.nnz()`? seems a bit inconsistent?

lucbv (changes requested):
Some more comments, otherwise looks reasonable to me

sparse/src/KokkosSparse_Utils.hpp
vqd8a: @lucbv I just do not want calling ```numRows()``` multiple times. To make it consistent, I now create a copy of A.nnz() too.

sparse/src/KokkosSparse_Utils.hpp
vqd8a: @lucbv Thanks. I did not know Kokkos has it. The code has been changed.

graph/impl/KokkosGraph_RCB_impl.hpp
vqd8a: @lucbv You are right. I have changed to using ```Kokkos::reduction_identity```
```std::numeric_limits<scalar_type>::infinity()``` --> ```Kokkos::reduction_identity<scalar_type>::min()```
```-std::numeric_limits<scalar_type>::infinity()``` --> ```Kokkos::reduction_identity<scalar_type>::max()```

File: graph/impl/KokkosGraph_RCB_impl.hpp (hunk 1 above)
Comment (vqd8a) [reply graph/impl/KokkosGraph_RCB_impl.hpp]: Thanks, @lucbv. If I use ```MinMaxPair<scalar_type>*```, the code will fail to compile. It looks like Kokkos requires ```Type[]```  to handle multiple reduction values. So I keep ```[]``` as it is. But, I have changed to using ```minmax_type``` and mapping it to ```value_type``` (as ```value_type``` is a mandatory type alias for the data type of the reduction result)
```
using minmax_type = MinMaxPair<scalar_type>[];
using value_type  = minmax_type;
```

vqd8a:
@lucbv I addressed your comments. Please take a quick look. Thanks.

graph/impl/KokkosGraph_RCB_impl.hpp
lucbv: Thanks for the explanation and update

lucbv (approved):

vqd8a:
@lucbv @ndellingwood I have quick questions. When will be the best time to have this in Trilinos? Can we have a patch for this PR in Trilinos?

lucbv:
@vqd8a the best time will be for Kokkos 4.3 (around November) since this is a new feature, at best we could consider it for the patch release 4.2.2 but that's pushing a bit the envelop of what patch release are for...
Note that you can easily build Trilinos against Kokkos Core and Kernels to test this feature, of course if you need to modify Trilinos that's something you might want to do in a branch?

vqd8a:
Thanks, @lucbv. Let's wait for Kokkos 4.3. There is a new diagonal block extraction function, so I am going to create a Trilinos branch for testing.

PR #3159, opened 2026-06-29
Explicitly include Kokkos_Half.hpp when using Kokkos::Experimental::[b]half_t

masterleinad:
Fixes https://my.cdash.org/builds/3680847/build by including the `Kokkos_Half.hpp` in all source files that don't include `Kokkos_Core.hpp` already.

### DIFF
--- a/common/src/KokkosKernels_ArithTraits.hpp
+++ b/common/src/KokkosKernels_ArithTraits.hpp
@@ -21,6 +21,7 @@
 /// \brief Declaration and definition of KokkosKernels::ArithTraits
 
 #include <KokkosKernels_config.h>
+#include <Kokkos_Half.hpp>
 #include <Kokkos_NumericTraits.hpp>
 #include <Kokkos_MathematicalFunctions.hpp>
 #include <Kokkos_Complex.hpp>
--- a/common/src/KokkosKernels_ArithTraits.hpp
+++ b/common/src/KokkosKernels_ArithTraits.hpp
@@ -21,11 +21,14 @@
 /// \brief Declaration and definition of KokkosKernels::ArithTraits
 
 #include <KokkosKernels_config.h>
+
+#include <Kokkos_Macros.hpp>
+#if KOKKOS_VERSION_GREATER_EQUAL(5, 1, 1)
 #include <Kokkos_Half.hpp>
+#endif
 #include <Kokkos_NumericTraits.hpp>
 #include <Kokkos_MathematicalFunctions.hpp>
 #include <Kokkos_Complex.hpp>
-#include <Kokkos_Macros.hpp>
 
 #include <impl/Kokkos_QuadPrecisionMath.hpp>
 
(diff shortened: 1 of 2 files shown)


lucbv (approved):

ndellingwood:
Compilation errors like:

```
 /home/runner/_work/kokkos-kernels/kokkos-kernels/kokkos/install/include/Cuda/Kokkos_Cuda_Half_Conversion.hpp(16): error: function "Kokkos::Experimental::cast_to_half(float)" has already been defined (previous definition at line 936 of /home/runner/_work/kokkos-kernels/kokkos-kernels/kokkos/install/include/impl/Kokkos_Half_FloatingPointWrapper.hpp)
  half_t cast_to_half(half_t val) { return val; }
         ^

/home/runner/_work/kokkos-kernels/kokkos-kernels/kokkos/install/include/Cuda/Kokkos_Cuda_Half_Conversion.hpp(19): error: function "Kokkos::Experimental::cast_to_half(float)" has already been defined (previous definition at line 936 of /home/runner/_work/kokkos-kernels/kokkos-kernels/kokkos/install/include/impl/Kokkos_Half_FloatingPointWrapper.hpp)
  half_t cast_to_half(float val) { return __float2half(val); }
...
```

kokkos-kernels CI  tests against the most recent kokkos release (5.1.1 here) rather than develop, are some changes needed for backward compatibility?

masterleinad:
> kokkos-kernels CI tests against the most recent kokkos release (5.1.1 here) rather than develop, are some changes needed for backward compatibility?

Ah, that makes sense. Let me fix that real quick.

masterleinad:
I just added a version guard now.

common/src/KokkosKernels_ArithTraits.hpp
```diff
@@ -21,10 +21,14 @@
 /// \brief Declaration and definition of KokkosKernels::ArithTraits
 
 #include <KokkosKernels_config.h>
+
+#include <Kokkos_Macros.hpp>
+#if KOKKOS_VERSION_GREATER_EQUAL(5, 1, 1)
```
lucbv: Should it be strictly greater than 5.1.1 or will it work with 5.1.1 already?

masterleinad:
```suggestion
#if KOKKOS_VERSION_GREATER(5, 1, 1)
```

sparse/src/KokkosSparse_IOUtils.hpp
```diff
@@ -7,6 +7,11 @@
 #include "KokkosSparse_CrsMatrix.hpp"
 #include "KokkosSparse_SellMatrix.hpp"
 
+#include <Kokkos_Macros.hpp>
+#if KOKKOS_VERSION_GREATER_EQUAL(5, 1, 1)
```
masterleinad:
```suggestion
#if KOKKOS_VERSION_GREATER(5, 1, 1)
```

common/src/KokkosKernels_ArithTraits.hpp
masterleinad: Yes, it should be past 5.1.1.

lucbv (approved):

PR #3160, opened 2026-06-30
Fix #3153: spgemm tpls not getting called

### LINKED ISSUE #3153: Choice of algorithm for SpGEMM
@brian-kelley 

It appears as though the default behavior for TPL usage has changed with #3060. It seems that `SPGEMM_DEFAULT` results in calling the native KokkosKernels implementation on Cuda, not the cuSparse TPL.

Currently, we are calling
https://github.com/kokkos/kokkos-kernels/blob/d72cc895c681dea693d7d000d9bff35314caf5dd/sparse/src/KokkosSparse_spgemm_handle.hpp#L787-L806
with `"SPGEMM_DEFAULT"` which returns `SPGEMM_KK`.

I then see `KokkosSparse::spgemm_symbolic[NATIVE]` in profiling output. I also see 
```
Tpetra::Details::IntRowPtrHelper<KokkosSparse::CrsMatrix<double, int, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, void, long unsigned int>, KokkosSparse::CrsMatrix<double, int, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, void, int> >::*)(const Kokkos::View<const long unsigned int*, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, Kokkos::MemoryTraits<0> >&), &Tpetra::Details::IntRowPtrHelper<KokkosSparse::CrsMatrix<double, int, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, void, long unsigned int>, KokkosSparse::CrsMatrix<double, int, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, void, int> >::fillRowptrsInt, 1>, void(int), Kokkos::View<int*, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, Kokkos::MemoryTraits<0> >, const Kokkos::View<const long unsigned int*, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, Kokkos::MemoryTraits<0> > > [for]
```
which should mean that the rowptr does get converted to the correct type for cusparse.

brian-kelley:
Fix #3153, where spgemm TPLs were not getting called as expected.

- In SPGEMMHandle constructor, do not call ``choose_default_algorithm`` which overwrote ``SPGEMM_DEFAULT`` with a specific native algorithm. Instead have the native impl internally pick the default.
- Fix ``KokkosSparse::StringToSPGEMMAlgorithm`` (again, preserve ``SPGEMM_DEFAULT``)
- Same thing in spgemm perf test: passing ``--algorithm DEFAULT`` now goes to ``SPGEMM_DEFAULT`` which can be a TPL

Tested locally:
- built with and without cusparse
- ran spgemm perf test with ``--algorithm DEFAULT`` under the kernel logger
- also tried with algorithm chosen like Tpetra does: ``KokkosSparse::StringToSPGEMMAlgorithm("SPGEMM_DEFAULT")``
- checked that the native or cusparse path is taken as expected

### DIFF
--- a/sparse/impl/KokkosSparse_spgemm_impl_memaccess.hpp
+++ b/sparse/impl/KokkosSparse_spgemm_impl_memaccess.hpp
@@ -235,8 +235,7 @@ void KokkosSPGEMM<HandleType, a_row_view_t_, a_lno_nnz_view_t_, a_scalar_nnz_vie
   this->create_read_write_hg(overall_flops, c_flop_rowmap, c_comp_a_net_index, c_comp_b_net_index, c_comp_row_index,
                              c_comp_col_index);
 
-  int write_type                   = 0;  // 0 -- KKMEM, 1-KKSPEED, 2- KKCOLOR 3-KKMULTICOLOR 4-KKMULTICOLOR2
-  SPGEMMAlgorithm spgemm_algorithm = this->handle->get_spgemm_handle()->get_algorithm_type();
+  int write_type = 0;  // 0 -- KKMEM, 1-KKSPEED, 2- KKCOLOR 3-KKMULTICOLOR 4-KKMULTICOLOR2
 
   if (spgemm_algorithm == KokkosKernels::Experimental::Graph::SPGEMM_KK_COLOR) {
     write_type = 2;
--- a/perf_test/sparse/KokkosSparse_spgemm.cpp
+++ b/perf_test/sparse/KokkosSparse_spgemm.cpp
@@ -106,7 +106,7 @@ void print_options() {
   std::cerr << "\t[Required] INPUT MATRIX: '--amtx [left_hand_side.mtx]' -- for C=AxA" << std::endl;
 
   std::cerr << "\t[Optional] '--algorithm "
-               "[DEFAULT=KKDEFAULT=KKSPGEMM|KKMEM|KKDENSE]' --> to choose algorithm. "
+               "[DEFAULT|KKDEFAULT=KKSPGEMM|KKMEM|KKDENSE]' --> to choose algorithm. "
                "KKMEM is outdated, use KKSPGEMM instead."
             << std::endl;
   std::cerr << "\t[Optional] --bmtx [righ_hand_side.mtx]' for C = AxB" << std::endl;
@@ -202,7 +202,7 @@ int parse_inputs(KokkosKernels::Experiment::Parameters& params, int argc, char**
       // because there are pre- post processing in these TPL kernel wraps.
     } else if (perf_test::check_arg_str(i, argc, argv, "--algorithm", algoStr)) {
       if (0 == Test::string_compare_no_case(algoStr, "DEFAULT")) {
-        params.algorithm = KokkosSparse::SPGEMM_KK;
+        params.algorithm = KokkosSparse::SPGEMM_DEFAULT;
       } else if (0 == Test::string_compare_no_case(algoStr, "KKDEFAULT")) {
         params.algorithm = KokkosSparse::SPGEMM_KK;
       } else if (0 == Test::string_compare_no_case(algoStr, "KKSPGEMM")) {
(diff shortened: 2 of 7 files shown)


sparse/impl/KokkosSparse_spgemm_impl_memaccess.hpp
```diff
@@ -235,8 +235,7 @@ void KokkosSPGEMM<HandleType, a_row_view_t_, a_lno_nnz_view_t_, a_scalar_nnz_vie
   this->create_read_write_hg(overall_flops, c_flop_rowmap, c_comp_a_net_index, c_comp_b_net_index, c_comp_row_index,
                              c_comp_col_index);
 
-  int write_type                   = 0;  // 0 -- KKMEM, 1-KKSPEED, 2- KKCOLOR 3-KKMULTICOLOR 4-KKMULTICOLOR2
-  SPGEMMAlgorithm spgemm_algorithm = this->handle->get_spgemm_handle()->get_algorithm_type();
+  int write_type = 0;  // 0 -- KKMEM, 1-KKSPEED, 2- KKCOLOR 3-KKMULTICOLOR 4-KKMULTICOLOR2
 
   if (spgemm_algorithm == KokkosKernels::Experimental::Graph::SPGEMM_KK_COLOR) {
```
lucbv, line 240: Is that still correct considering the definition of `spgemm_algorithm` was removed a couple lines above?

lucbv (approved):
Looks reasonable, just a quick question

sparse/impl/KokkosSparse_spgemm_impl_memaccess.hpp
lucbv: Oh, do we need to update the documentation to reflect this change or was the previous implementation just a bug and now everything is in line with what was stated?

brian-kelley: ``spgemm_algorithm`` is also a member of the KokkosSPGEMM classs. It's set up here in the constructor:
```
  KokkosSPGEMM(HandleType *handle_, nnz_lno_t m_, nnz_lno_t n_, nnz_lno_t k_, const_a_lno_row_view_t row_mapA_,
               const_a_lno_nnz_view_t entriesA_, bool transposeA_, const_b_lno_row_view_t row_mapB_,
               const_b_lno_nnz_view_t entriesB_, bool transposeB_)
...
        use_dynamic_schedule(handle_->is_dynamic_scheduling()),
        KOKKOSKERNELS_VERBOSE(handle_->get_verbose()),
        MyEnumExecSpace(this->handle->get_handle_exec_space()),
        spgemm_algorithm(select_algorithm(this->handle)),
        spgemm_accumulator(this->handle->get_spgemm_handle()->get_accumulator_type())
  //,row_mapC(), entriesC(), valsC()
  {}
```
``select_algorithm`` just gets the algo choice out of the handle, and if that is SPGEMM_DEFAULT, it picks the default native algo. Before, the handle's algo could never be SPGEMM_DEFAULT because its constructor would do this selection.

So for this ``write_type`` logic, we now want to use ``KokkosSPGEMM::spgemm_algorithm `` instead of ``handle->get_spgemm_handle()->get_algorithm_type()``.

brian-kelley: It wasn't a bug although it was a bit redundant since the local ``spgemm_algorithm`` shadowed the member, and it always had the same value as the member. The overall behavior of the native impl doesn't change by this PR

lucbv: Okay, thanks for the added explanations : )

ndellingwood:
Thanks @brian-kelley ! h2d counts in MueLu tests are back to previous expected values

Issue #3161, opened 2026-07-01
Various unit test failures with deprecated code 5 disabled


ndellingwood:
Following merge of https://github.com/kokkos/kokkos/pull/9076 various unit tests are failing when KOKKOS_ENABLE_DEPRECATED_CODE_5=OFF is set with failure outputs:

```
Asserted condition `static_cast<int>(r) < static_cast<int>(rank())` evaluated false.
19:05:14 Error at "/home/jenkins/blake-new/workspace/KokkosEco_Trilinos_Blake_LLVM1507_Serial_debug-new_view-dep_code5_off/Trilinos/kokkos/core/src/View/Kokkos_BasicView.hpp":169
```

Failing tests from llvm/15 serial debug build:
```
         48 - KokkosKernels_common_serial (Subprocess aborted)
         51 - KokkosKernels_batched_dla_serial (Subprocess aborted)
         53 - KokkosKernels_batched_sla_serial (Subprocess aborted)
         54 - KokkosKernels_blas_serial (Subprocess aborted)
         57 - KokkosKernels_sparse_serial (Subprocess aborted)
         58 - KokkosKernels_blocksparse_serial (Subprocess aborted)
         59 - KokkosKernels_ode_serial (Subprocess aborted)
         60 - KokkosKernels_docs_blas1_abs (Subprocess aborted)
         61 - KokkosKernels_docs_blas1_scal (Subprocess aborted)
         68 - KokkosKernels_sparse_spiluk (Subprocess aborted)
         74 - KokkosKernels_wiki_blas1_rotg_rot (Subprocess aborted)
         75 - KokkosKernels_wiki_blas1_rotmg_rotm (Subprocess aborted)
         81 - KokkosKernels_wiki_spmv (Subprocess aborted)
         84 - KokkosKernels_wiki_gauss_seidel (Subprocess aborted)
         89 - KokkosKernels_gmres_test_prec (Subprocess aborted)
```

Cross-reference: https://github.com/kokkos/kokkos/pull/9076#issuecomment-4785264966

lucbv:
How are we doing on this front after the merge of #3162 , are there any issues remaining?

ndellingwood:
@lucbv I think we're good to close this

PR #3162, opened 2026-07-02
Compatibility updates for deprecated out-of-rank extent usages

ndellingwood:
This PR includes work toward addressing https://github.com/kokkos/kokkos-kernels/issues/3161

Compatibility update when building with `Kokkos_ENABLE_DEPRECATED_CODE_5=OFF` following https://github.com/kokkos/kokkos/pull/9076

### DIFF
--- a/batched/dense/unit_test/Test_Batched_VectorView.hpp
+++ b/batched/dense/unit_test/Test_Batched_VectorView.hpp
@@ -25,7 +25,7 @@ namespace Test {
 
 template <int extent_id, typename VectorViewType>
 inline int get_extent(const VectorViewType& v) {
-  if constexpr (VectorViewType::rank == extent_id + 1) {
+  if constexpr (extent_id < VectorViewType::rank) {
     return v.extent(extent_id);
   } else
     return 1;
--- a/batched/dense/unit_test/Test_Batched_VectorView.hpp
+++ b/batched/dense/unit_test/Test_Batched_VectorView.hpp
@@ -23,28 +23,37 @@ using namespace KokkosBatched;
 
 namespace Test {
 
+template <int rank, typename VectorViewType>
+inline int get_extent(const VectorViewType& v) {
+  if constexpr (VectorViewType::rank == rank+1) {
+    return v.extent(rank);
+  }
+  else
+    return 1;
+}
+
 template <typename VectorViewType>
 void impl_init_vector_view(const VectorViewType& a) {
   int cnt = 0;
-  for (int i0 = 0, i0end = a.extent(0); i0 < i0end; ++i0)
-    for (int i1 = 0, i1end = a.extent(1); i1 < i1end; ++i1)
-      for (int i2 = 0, i2end = a.extent(2); i2 < i2end; ++i2)
-        for (int i3 = 0, i3end = a.extent(3); i3 < i3end; ++i3)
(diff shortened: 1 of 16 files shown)


blas/impl/KokkosBlas1_update_spec.hpp
```diff
@@ -230,7 +230,7 @@ struct Update<execution_space, XV, YV, ZV, 1, false, KOKKOSKERNELS_IMPL_COMPILE_
 #endif
 
     const size_type numRows = X.extent(0);
-    const size_type numCols = X.extent(1);
+    const size_type numCols = 1;  // X.extent(1);
```
ndellingwood: L217 indicates the rank should be 1 here, and L214 indicates the rank should match for XV, YV, and ZV , hopefully this is a reasonable update?

ndellingwood:
First round of testing was a Serial debug build, no TPLs enabled in testing yet

batched/dense/unit_test/Test_Batched_VectorView.hpp
```diff
@@ -23,28 +23,36 @@ using namespace KokkosBatched;
 
 namespace Test {
 
+template <int rank, typename VectorViewType>
+inline int get_extent(const VectorViewType& v) {
+  if constexpr (VectorViewType::rank == rank + 1) {
```
brian-kelley: Should this be
```suggestion
  if constexpr (rank < VectorViewType::rank) {
```
? ``VectorViewType::rank == rank + 1`` would be true only when getting the last dimension.

ndellingwood: @brian-kelley I updated the template parameter name from `rank` to `extend_id` to improve clarity of what is being passed in and compared

blas/impl/KokkosBlas1_update_spec.hpp
lucbv: Actually if you removed this line and the usage of numCols on line 252 then we would not have to worry about it anymore!

lucbv (changes requested):
Just a small change to make one update even simpler to maintain

lucbv (approved):

batched/dense/unit_test/Test_Batched_VectorView.hpp
brian-kelley: I'm still not sure about this part. The views in this test can be rank 1 up to rank 8 (with each extent being hardcoded to 4). So I think it does matter that the ``get_extent<k>`` returns ``v.extent(k)`` if ``k < VectorViewType::rank``.

For this version of get_extent, this prints out "myView is 1x8" instead of "myView is 9x8"
```
Kokkos::View<int**> myView("myView", 9, 8); 
std::cout << "myView is " << get_extent<0>(myView) << "x" << get_extent<1>(myView) << '\n';
```

ndellingwood: Good catch, I'll get it fixed

ndellingwood: @brian-kelley updated in [b698a97](https://github.com/kokkos/kokkos-kernels/pull/3162/commits/b698a9716b1b41ecc99bd9823d92158995d86cae)

lucbv (approved):

Issue #1465, opened 2022-07-14
Inconsistent result of `KokkosBlas::nrm2` for `CudaSpace` and `CudaUVMSpace`


romintomasetti:
Using the following code snippet leads to inconsistent result of `KokkosBlas::nrm2` with `Kokkos::CudaSpace` and `Kokkos::CudaUVMSpace`.

It seems that:
* when the view is in `Kokkos::CudaUVMSpace` memory space, `KokkosBlas::nrm2` branches on the `KokkosBlas`'s functor,
* when the view is in `Kokkos::CudaSpace` memory space, it uses `cublasDnrm2`.

The results are different because `KokkosBlas`'s functor implements the "naïve" approach (might be related to #1076) while the one of `cublas` `uses a multiphase model of accumulation to avoid intermediate underflow and overflow` (https://docs.nvidia.com/cuda/cublas/index.html#cublas-lt-t-gt-nrm2).

Here is the snippet:
```c++
using scalar_t = double;
using mem_space_A = Kokkos::CudaSpace;
using mem_space_B = Kokkos::CudaUVMSpace;

using view_t_A = Kokkos::View<scalar_t*,mem_space_A>;
using view_t_B = Kokkos::View<scalar_t*,mem_space_B>;

view_t_A a("view without UVM",10000); Kokkos::deep_copy(a,1.0);
view_t_B b("view with    UVM",10000); Kokkos::deep_copy(b,1.0);

scalar_t a_nrm2 = KokkosBlas::nrm2(a);
scalar_t b_nrm2 = KokkosBlas::nrm2(b);

scalar_t res_a, res_b;
cublasHandle_t handle;
cublasCreate(&handle);
cublasDnrm2(handle,a.size(),a.data(),1,&res_a);
cublasDnrm2(handle,b.size(),b.data(),1,&res_b);

std::cout << std::setprecision(25) << std::scientific;
std::cout << "> Nrm2 of a with KokkosBlas : " << a_nrm2 << std::endl;
std::cout << "> Nrm2 of b with KokkosBlas : " << b_nrm2 << std::endl;
std::cout << "> Nrm2 of a with cublasDnrm2: " << res_a  << std::endl;
std::cout << "> Nrm2 of b with cublasDnrm2: " << res_b  << std::endl;
```
that gives:
```bash
> Nrm2 of a with KokkosBlas : 1.0000000000000001421085472e+02
[...]

PR #3168, opened 2026-07-07
BLAS: remove memory space from unification layer

### LINKED ISSUE #775: cuBLAS + UVM
I am trying to use cuBLAS from EMPIRE / Trilinos / Tpetra. Since KokkosKernels lacks specializations for cuBLAS + UVM that currently isn't possible. I tried adding them, similar to https://github.com/kokkos/kokkos-kernels/pull/759, but there still seems to be a mismatch.

Is there a reason for not supporting this use case?

brian-kelley:
- Remove memspace from BLAS ETI (all views use either HostSpace or ExecSpace as their device)
  - Now for example, CudaSpace and CudaUVMSpace views both get unified to just Cuda as their device type
  - Resolves #775 and #1465
  - ran 1465 reproducer and checked that CUBLAS nrm2 is called for both spaces
- Consistently use GetUnifiedLayoutPreferring to maximize ETI matching
- Fix link error in LAPACK gesv unit test when instantiating for SharedSpace
- Make reduction-like functions (dot, sum, iamax, nrm1, nrm2, nrm2w, nrminf) more consistent in how their ETI works
- Fix overly broad templating on ExecSpace in host BLAS tpl avail/decl

Testing:
- workstation
  - serial+openmp with host blas
  - serial+openmp without host blas
  - cuda with TPLs, and CudaSpace instantiated
  - cuda with TPLs, and CudaUVMSpace instantiated
- eldorado
  - rocblas/rocsparse, HIPSpace instantiated
  - rocblas/rocsparse, HIPManagedSpace


brian-kelley:
Obviously this is a lot to review and I would be happy to split the changes into smaller chunks (each kernel can be changed independently). However, I would like to see if the full refactor passes autotester.

brian-kelley:
All required checks passed so I'll start splitting this one up.

PR #3171, opened 2026-07-08
BLAS ETI/TPL refactor: core changes

brian-kelley:
This is the PR 1 in a refactor of the BLAS unification layer (splitting up the large draft PR #3168 for easier review).

The main change is to instantiate on ExecSpace instead of Device. This way, Views with ``<Cuda, CudaSpace>``, ``<Cuda, CudaUVMSpace>`` and ``<Cuda, CudaHostPinnedSpace>`` are all unified to the same implementation or TPL. This works towards resolving #775 and #1465.

This PR does:
- Add ``EXEC_SPACES`` as an ETI type list in cmake
- Add new helper which casts an unmanaged View to desired layout and device type
  - because we now can't always assign user's input views to the "_Internal" view types directly
- Refactor axpby to by ETI'd on ExecSpace only instead of Device. Similar changes coming for all other kernels.
- Fix LAPACK gesv test when instantiating UVM/HIPManaged only

### DIFF
--- a/blas/CMakeLists.txt
+++ b/blas/CMakeLists.txt
@@ -53,6 +53,13 @@ endif()
 #Generate @X@ variables in the template X.hpp.in and X.cpp.in
 #files containing the list of all needed macros
 
+macro(gen_blas_eti_new FUNC_NAME SUBDIR)
+  kokkoskernels_generate_eti(${FUNC_NAME} ${SUBDIR}
+    COMPONENTS  blas
+    HEADER_LIST ETI_HEADERS
+    SOURCE_LIST SOURCES
+    TYPE_LISTS  FLOATS LAYOUTS EXEC_SPACES)
+endmacro()
 macro(gen_blas_eti FUNC_NAME SUBDIR)
   kokkoskernels_generate_eti(${FUNC_NAME} ${SUBDIR}
     COMPONENTS  blas
@@ -67,8 +74,8 @@ gen_blas_eti(Blas1_scal           scal)
 gen_blas_eti(Blas1_scal_mv        scal)
 gen_blas_eti(Blas1_dot            dot)
 gen_blas_eti(Blas1_dot_mv         dot)
-gen_blas_eti(Blas1_axpby          axpby)
-gen_blas_eti(Blas1_axpby_mv       axpby)
+gen_blas_eti_new(Blas1_axpby          axpby)
+gen_blas_eti_new(Blas1_axpby_mv       axpby)
 gen_blas_eti(Blas1_update         update)
 gen_blas_eti(Blas1_update_mv      update)
 gen_blas_eti(Blas1_sum            sum)
--- a/blas/tpls/KokkosBlas1_axpby_tpl_spec_avail.hpp
+++ b/blas/tpls/KokkosBlas1_axpby_tpl_spec_avail.hpp
@@ -20,44 +20,51 @@ namespace Impl {
 // Generic Host side BLAS (could be MKL or whatever)
 #ifdef KOKKOSKERNELS_ENABLE_TPL_BLAS
 
-#define KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(SCALAR, LAYOUT, MEMSPACE)                                             \
-  template <class ExecSpace>                                                                                        \
-  struct axpby_tpl_spec_avail<                                                                                      \
-      ExecSpace, SCALAR,                                                                                            \
-      Kokkos::View<const SCALAR*, LAYOUT, Kokkos::Device<ExecSpace, MEMSPACE>,                                      \
-                   Kokkos::MemoryTraits<Kokkos::Unmanaged> >,                                                       \
-      SCALAR,                                                                                                       \
-      Kokkos::View<SCALAR*, LAYOUT, Kokkos::Device<ExecSpace, MEMSPACE>, Kokkos::MemoryTraits<Kokkos::Unmanaged> >, \
-      1> {                                                                                                          \
-    enum : bool { value = true };                                                                                   \
+#define KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(EXEC_SPACE, SCALAR, LAYOUT)                                            \
+  template <>                                                                                                        \
+  struct axpby_tpl_spec_avail<                                                                                       \
+      EXEC_SPACE, SCALAR, Kokkos::View<const SCALAR*, LAYOUT, EXEC_SPACE, Kokkos::MemoryTraits<Kokkos::Unmanaged> >, \
+      SCALAR, Kokkos::View<SCALAR*, LAYOUT, EXEC_SPACE, Kokkos::MemoryTraits<Kokkos::Unmanaged> >, 1> {              \
+    enum : bool { value = true };                                                                                    \
   };
 
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(double, Kokkos::LayoutLeft, Kokkos::HostSpace)
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(float, Kokkos::LayoutLeft, Kokkos::HostSpace)
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::complex<double>, Kokkos::LayoutLeft, Kokkos::HostSpace)
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::complex<float>, Kokkos::LayoutLeft, Kokkos::HostSpace)
+#ifdef KOKKOS_ENABLE_SERIAL
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::Serial, double, Kokkos::LayoutLeft)
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::Serial, float, Kokkos::LayoutLeft)
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::Serial, Kokkos::complex<double>, Kokkos::LayoutLeft)
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::Serial, Kokkos::complex<float>, Kokkos::LayoutLeft)
+#endif
+#ifdef KOKKOS_ENABLE_OPENMP
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::OpenMP, double, Kokkos::LayoutLeft)
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::OpenMP, float, Kokkos::LayoutLeft)
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::OpenMP, Kokkos::complex<double>, Kokkos::LayoutLeft)
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::OpenMP, Kokkos::complex<float>, Kokkos::LayoutLeft)
+#endif
+#ifdef KOKKOS_ENABLE_THREADS
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::Threads, double, Kokkos::LayoutLeft)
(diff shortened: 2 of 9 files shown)


blas/tpls/KokkosBlas1_axpby_tpl_spec_avail.hpp
```diff
@@ -20,44 +20,51 @@ namespace Impl {
...
+      Kokkos::View<const SCALAR*, LAYOUT, Kokkos::Cuda, Kokkos::MemoryTraits<Kokkos::Unmanaged> >, SCALAR, \
+      Kokkos::View<SCALAR*, LAYOUT, Kokkos::Cuda, Kokkos::MemoryTraits<Kokkos::Unmanaged> >, 1> {          \
+    enum : bool { value = true };                                                                          \
   };
 
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_CUBLAS(double, Kokkos::LayoutLeft, Kokkos::CudaSpace)
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_CUBLAS(float, Kokkos::LayoutLeft, Kokkos::CudaSpace)
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_CUBLAS(Kokkos::complex<double>, Kokkos::LayoutLeft, Kokkos::CudaSpace)
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_CUBLAS(Kokkos::complex<float>, Kokkos::LayoutLeft, Kokkos::CudaSpace)
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_CUBLAS(double, Kokkos::LayoutLeft)
```
lucbv: If we are only instantiating layout left, then we should hard code it in the macro definition and remove it from the macro arguments.

blas/tpls/KokkosBlas1_axpby_tpl_spec_avail.hpp
```diff
@@ -20,44 +20,51 @@ namespace Impl {
...
+      SCALAR, Kokkos::View<SCALAR*, LAYOUT, EXEC_SPACE, Kokkos::MemoryTraits<Kokkos::Unmanaged> >, 1> {              \
+    enum : bool { value = true };                                                                                    \
   };
 
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(double, Kokkos::LayoutLeft, Kokkos::HostSpace)
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(float, Kokkos::LayoutLeft, Kokkos::HostSpace)
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::complex<double>, Kokkos::LayoutLeft, Kokkos::HostSpace)
-KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::complex<float>, Kokkos::LayoutLeft, Kokkos::HostSpace)
+#ifdef KOKKOS_ENABLE_SERIAL
+KOKKOSBLAS1_AXPBY_TPL_SPEC_AVAIL_BLAS(Kokkos::Serial, double, Kokkos::LayoutLeft)
```
lucbv: Is there no way to unify all these host execution spaces in the case of BLAS since it all goes down to `KokkosBlas::HostBlas<>` which does nothing with the execution space if I recall correctly?

blas/CMakeLists.txt
```diff
@@ -67,8 +74,8 @@ gen_blas_eti(Blas1_scal           scal)
 gen_blas_eti(Blas1_scal_mv        scal)
 gen_blas_eti(Blas1_dot            dot)
 gen_blas_eti(Blas1_dot_mv         dot)
-gen_blas_eti(Blas1_axpby          axpby)
-gen_blas_eti(Blas1_axpby_mv       axpby)
+gen_blas_eti_new(Blas1_axpby          axpby)
```
brian-kelley, line 77: When reviewing, note that the ``gen_blas_eti_new`` is just temporary and added so that the refactor can be split into multiple PRs. The last PR will make ``gen_blas_eti`` use the new system and get rid of ``gen_blas_eti_new``.

common/src/KokkosKernels_helpers.hpp
```diff
@@ -47,10 +47,61 @@ struct GetUnifiedScalarViewType<T, TX, true, true> {
...
+  } else {
+    if constexpr (std::is_same_v<typename V1::array_layout, typename V2::array_layout>) {
+      return V1(v.data(), v.layout());
+    }
+    // If layout types aren't the same, first convert to an intermediate type with the target layout
+    Kokkos::View<typename V2::data_type, typename V1::array_layout, typename V2::device_type,
+                 Kokkos::MemoryTraits<Kokkos::Unmanaged>>
+        temp = v;
+    // Then we can we use View assignment to change to V1's device type and traits
+    return V1(temp.data(), temp.layout());
```
lucbv, line 101: Isn't `temp.layout()` the same as `v.layout()`? In which case, what are we fixing with the intermediate view construction?

blas/tpls/KokkosBlas1_axpby_tpl_spec_decl.hpp
```diff
@@ -182,165 +179,145 @@ KOKKOSBLAS1_CAXPBY_BLAS(Kokkos::LayoutLeft, Kokkos::HostSpace, false)
...
 
-KOKKOSBLAS1_ZAXPBY_CUBLAS(Kokkos::LayoutLeft, Kokkos::CudaSpace, true)
-KOKKOSBLAS1_ZAXPBY_CUBLAS(Kokkos::LayoutLeft, Kokkos::CudaSpace, false)
+KOKKOSBLAS1_ZAXPBY_CUBLAS(Kokkos::LayoutLeft, true)
+KOKKOSBLAS1_ZAXPBY_CUBLAS(Kokkos::LayoutLeft, false)
 
-KOKKOSBLAS1_CAXPBY_CUBLAS(Kokkos::LayoutLeft, Kokkos::CudaSpace, true)
-KOKKOSBLAS1_CAXPBY_CUBLAS(Kokkos::LayoutLeft, Kokkos::CudaSpace, false)
+KOKKOSBLAS1_CAXPBY_CUBLAS(Kokkos::LayoutLeft, true)
+KOKKOSBLAS1_CAXPBY_CUBLAS(Kokkos::LayoutLeft, false)
```
lucbv: Since we are only using `Kokkos::LayoutLeft`, let us just remove it from the macro interface.

```suggestion
#define KOKKOSBLAS1_CAXPBY_CUBLAS(ETI_SPEC_AVAIL)                                                             \
  template <>                                                                                                         \
  struct Axpby<                                                                                                       \
      Kokkos::Cuda, Kokkos::complex<float>,                                                                           \
      Kokkos::View<const Kokkos::complex<float>*, Kokkos::LayoutLeft, Kokkos::Cuda, Kokkos::MemoryTraits<Kokkos::Unmanaged> >,    \
Kokkos::complex<float>,                                                                                         \
      Kokkos::View<Kokkos::complex<float>*, Kokkos::LayoutLeft, Kokkos::Cuda, Kokkos::MemoryTraits<Kokkos::Unmanaged> >, 1, true, \
      ETI_SPEC_AVAIL> {                                                                                               \
typedef Kokkos::complex<float> AV;                                                                                \
typedef Kokkos::complex<float> BV;                                                                                \
    typedef Kokkos::View<const Kokkos::complex<float>*, Kokkos::LayoutLeft, Kokkos::Cuda,                                         \
Kokkos::MemoryTraits<Kokkos::Unmanaged> >                                                    \
XV;                                                                                                           \
    typedef Kokkos::View<Kokkos::complex<float>*, Kokkos::LayoutLeft, Kokkos::Cuda, Kokkos::MemoryTraits<Kokkos::Unmanaged> > YV; \
typedef typename XV::size_type size_type;                                                                         \
\
    static void axpby(const Kokkos::Cuda& space, const AV& alpha, const XV& X, const BV& beta, const YV& Y) {         \
Kokkos::Profiling::pushRegion("KokkosBlas::axpby[TPL_CUBLAS,complex<float>]");                                  \
const size_type numElems = X.extent(0);                                                                         \
if ((numElems < static_cast<size_type>(INT_MAX)) && (beta == 1.0f)) {                                           \
axpby_print_specialization<AV, XV, BV, YV>();                                                                 \
const int N                            = static_cast<int>(numElems);                                          \
constexpr int one                      = 1;                                                                   \
KokkosBlas::Impl::CudaBlasSingleton& s = KokkosBlas::Impl::CudaBlasSingleton::singleton();                    \
KOKKOSBLAS_IMPL_CUBLAS_SAFE_CALL(cublasSetStream(s.handle, space.cuda_stream()));                             \
KOKKOSBLAS_IMPL_CUBLAS_SAFE_CALL(cublasCaxpy(s.handle, N, reinterpret_cast<const cuComplex*>(&alpha),         \
reinterpret_cast<const cuComplex*>(X.data()), one,               \
reinterpret_cast<cuComplex*>(Y.data()), one));                   \
KOKKOSBLAS_IMPL_CUBLAS_SAFE_CALL(cublasSetStream(s.handle, NULL));                                            \
} else                                                                                                          \
        Axpby<Kokkos::Cuda, AV, XV, BV, YV, YV::rank, false, ETI_SPEC_AVAIL>::axpby(space, alpha, X, beta, Y);        \
Kokkos::Profiling::popRegion();                                                                                 \
}                                                                                                                 \
};

KOKKOSBLAS1_DAXPBY_CUBLAS(true)
KOKKOSBLAS1_DAXPBY_CUBLAS(false)

KOKKOSBLAS1_SAXPBY_CUBLAS(true)
KOKKOSBLAS1_SAXPBY_CUBLAS(false)

KOKKOSBLAS1_ZAXPBY_CUBLAS(true)
KOKKOSBLAS1_ZAXPBY_CUBLAS(false)

KOKKOSBLAS1_CAXPBY_CUBLAS(true)
KOKKOSBLAS1_CAXPBY_CUBLAS(false)
```

blas/tpls/KokkosBlas1_axpby_tpl_spec_decl.hpp
lucbv, line 1: Remove all the `LAYOUT` from our ETI macros since we enforce `Kokkos::LayoutLeft`, one less thing that can go wrong later on...

lucbv (changes requested):
Some questions and I think some clean-up as well but fundamentally good with me : )

common/src/KokkosKernels_helpers.hpp
brian-kelley: Not always, the v -> temp step does the same thing as our unification always did (e.g. ``X_Internal x_internal = x;``). So it can switch between LayoutLeft and LayoutRight for 0D and 1D views. The second step is the new part, where we force the returned view to have the device we want, ``V1::device_type``. For BLAS this is now ``execution_space``. Kokkos won't let you directly convert a view with device_type ``Device<Cuda, CudaHostPinnedSpace>`` view to one with device_type ``Cuda`` so this lets us continue supporting host pinned.

blas/tpls/KokkosBlas1_axpby_tpl_spec_avail.hpp
brian-kelley: We used to template the eti_spec_avail and implementations on ExecSpace, but this ended up being wrong if you enable host BLAS but not device TPLs. The compiler sees the template and plugs in ExecSpace=Cuda.

I tried using enable_if via default template parameter on condition ``is_same_v<ExecSpace::memory_space, Kokkos::HostSpace>``, but it seems you can't use a default template parameter on partial specializations. I will look at this again and see if there's another way. Maybe using ``requires`` will work...

blas/tpls/KokkosBlas1_axpby_tpl_spec_decl.hpp
brian-kelley: ugh, github is not letting me apply your patch. I'll make this change though

blas/tpls/KokkosBlas1_axpby_tpl_spec_avail.hpp
brian-kelley: Actually, I think I'm wrong about when the issue happens. But something with this template caused failures on the HIP MI210 TPL build. That has both host blas and rocblas enabled.

brian-kelley: OK, using ``requires`` seems to work. The host exec spaces are back to being covered by one template.

brian-kelley:
Think this is ready for a new review. Kumquat hasn't run yet but I have tested locally on pascal/cuda12 so it'll probably be good

lucbv (approved):
Okay that looks good to me, thanks for looking into the requires clause for the host execution spaces, it is a lot cleaner in my opinion : )

Issue #2739, opened 2025-08-19
Sycl test failures with intel oneapi 2024.2.1


ndellingwood:
The following tests are failing when testing with the Sycl backend on PVC arch in containerized testing:

```
The following tests FAILED:
   11 - blas_sycl (Failed)
   17 - sparse_sycl (Failed)
   18 - blocksparse_sycl (Failed)
```

More detailed output:

**blas_sycl**:
```
[ RUN      ] sycl_test.gemv_complex_double
/var/kokkos-kernels/blas/unit_test/Test_Blas2_gemv.hpp:89: expected(i)=(-6.3513,5.11578), h_y(i)=(-0.182747,-0.221999)
/var/kokkos-kernels/blas/unit_test/Test_Blas2_gemv.hpp:89: expected(i)=(-5.73633,-8.32032), h_y(i)=(-2.78235,-6.85526)
/var/kokkos-kernels/blas/unit_test/Test_Blas2_gemv.hpp:89: expected(i)=(-2.99421,6.36861), h_y(i)=(-3.32041,5.05302)
/var/kokkos-kernels/blas/unit_test/Test_Blas2_gemv.hpp:89: expected(i)=(4.34386,2.34132), h_y(i)=(9.68928,-0.660492)
/var/kokkos-kernels/blas/unit_test/Test_Blas2_gemv.hpp:89: expected(i)=(-14.9297,-3.40846), h_y(i)=(-4.4837,-2.31755)
/var/kokkos-kernels/blas/unit_test/Test_Blas2_gemv.hpp:89: expected(i)=(-1.23367,1.26546), h_y(i)=(-3.58353,-1.58239)
/var/kokkos-kernels/blas/unit_test/Test_Blas2_gemv.hpp:89: expected(i)=(1.18362,-5.81078), h_y(i)=(-0.666861,-4.31589)
/var/kokkos-kernels/blas/unit_test/Test_Blas2_gemv.hpp:89: expected(i)=(-1.9987,4.08174), h_y(i)=(-0.2554,7.11416)
/var/kokkos-kernels/blas/unit_test/Test_Blas2_gemv.hpp:89: expected(i)=(3.94619,-0.926831), h_y(i)=(5.45093,1.06786)
/var/kokkos-kernels/blas/unit_test/Test_Blas2_gemv.hpp:89: expected(i)=(-5.12418,-4.75785), h_y(i)=(-1.60102,0.164905)
[...]

PR #3173, opened 2026-07-09
Drop KokkosBatched::view_rank and use Kokkos::rank instead

dalg24:
Following up on https://github.com/kokkos/kokkos-kernels/pull/3113/changes#r3548453312

Kokkos Core already provides a free function that accepts both View and DunRankView so there is little point in introducing a helper function for it.


yasahi-hpc (approved):
Good to know this

ndellingwood:
* The pv ci job (not required to pass) has known unit test failures https://github.com/kokkos/kokkos-kernels/issues/2739

ndellingwood (approved):

PR #3174, opened 2026-07-09
[Batched][Refactor] Replaced `view_rank()` on `Kokkos::rank()`

ViNN280801:
As @dalg24 noticed in #3113:

> There is Kokkos::rank free function with overloads for View and DynRankView.

This PR contains replacement of excess implementation of `rank` on existing `Kokkos::rank()`.

### DIFF
--- a/batched/KokkosBatched_Util.hpp
+++ b/batched/KokkosBatched_Util.hpp
@@ -58,17 +58,6 @@ namespace KokkosBatched {
 #define Int2String(A) Int2StringHelper(A)
 #define StringCat(A, B) A B
 
-// nvcc+MSVC device code cannot resolve Kokkos::View::rank via instance call;
-// use ViewType::rank() (compile-time) for fixed-rank View, v.rank() for DynRankView.
-template <typename ViewType>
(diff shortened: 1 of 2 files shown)


ViNN280801:
Also, I noticed that after updating Visual Studio to 2026 (MSVC 19.51) my previous CUDA Toolkit 12.6 becomes invalid, I faced with the same problem as described [here](https://developercommunity.microsoft.com/t/11091998).

So, for now I can't made fast check even the compilation. But for real, I don't think that these changes can break smth. BUT, if checking from my side is necessarily, I do it tomorrow.

ViNN280801:
I faced with another problem because I need to rebuild Kokkos to specify the kokkos isntall dir for the kokkos-kernels, so:

[compilation-output-msvc2022-ninja-cuda-err.log](https://github.com/user-attachments/files/29859506/compilation-output-msvc2022-ninja-cuda-err.log)

lucbv:
I believe this was already done in PR #3173 so I will go ahead and close this.

ViNN280801:
Sorry, didn't see this.

Issue #3184, opened 2026-07-15
gemm degenarate A execution space is not respected


jakub-homola:
Hi.

blas/src/KokkosBlas3_gemm.hpp, line 144:

https://github.com/kokkos/kokkos-kernels/blob/7012d301bd2b4a9f534f4338801ba5380e6f9da5/blas/src/KokkosBlas3_gemm.hpp#L144

```
  // Simply scale C if A matrix is degenerated
  if (A.extent(1) == 0) {
    scal(C, beta, C);
    return;
  }
```

the scal should be
```
scal(space, C, beta, C);
```

The execution space parameter is missing.

Please also check other kernels for a similar issue, I'm not really familiar with the codebase.

Don't ask how long I've been debugging to find this.

I'm using KokkosKernels 5.1.1, but also see this issue in the current develop branch.

Thanks,

Jakub

lucbv:
@jakub-homola PR #3185 proposes the obvious fix for the problem you reported. We are going to try to get this tested quickly and added in the release candidate branch for 5.2 so you can use the fixed code soon.
Thank you for taking the time to analyze your issue with Kokkos Kernels and give a very concise description, that helps us a lot making generating the fix quickly.
Can I ask you how you are using Kokkos Kernels? We are always interested in understanding what people are doing with the library : )

jakub-homola:
Thanks @lucbv

> ... how you are using Kokkos Kernels?

I am just writing up some question and comments that I intend to post in the Slack channel, I will also include my usecase.

ndellingwood:
Should be resolved by #3185

jakub-homola:
not sure what the traditions are here - should I, as the creator, close the issue, or leave it up to you?

lucbv:
I can close this one, in general I like waiting for the person who asked that the question/issue has indeed been resolved to their expectation.

PR #3185, opened 2026-07-15
BLAS - GEMM: fixing call to scal missing space param

lucbv:
Fixes issue #3184

@jakub-homola reports this problem in the above issue. It obviously can lead to weird synchronization issues and performance bugs as well. Looked at the rest of the main gemm code path and did not notice any other similar issues but will keep this in mind when working on other part of the code...

I am going to mark this as "BlocksPromotion" so it gets picked in our current release branch and included in 5.2.0 which should come out in a week or so...


ndellingwood (approved):

ndellingwood:
Cherry-picked to 5.2.0 as part of #3186

PR #3187, opened 2026-07-17
Encapsulate sparse unit test managing of random seeds

jgfouca:
The problem that randomly showed up in CI testing was this:
`FAILED Cuda.sparse_coo2crs` with seed `583998535`.

I was unable to reproduce this failure but I did notice that managing seeds was a little difficult in the unit tests; you have to change code to select a specific seed. I wrote a little infrastructure to make this easier and add support for the flag `--gtest_random_seed=...`.

In an effort to reproduce the fail, I ran the test above with 100+ different seeds and it passed every time.

### DIFF
--- a/common/unit_test/Test_Common_TestUtils.hpp
+++ b/common/unit_test/Test_Common_TestUtils.hpp
@@ -1,0 +1,178 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
+
+/// \file Test_Common_TestUtils.hpp
+/// \brief Tests for KokkosKernels_TestUtils.hpp random-generation utilities.
+///
+/// Verifies that constructing RandCooMat and RandCsMatrix twice with the same
+/// seed produces bit-for-bit identical data.  A FNV-1a hash of each view's raw
+/// bytes is computed and compared.  If the hashes differ, the test fails and
+/// the offending seed is reported so the failure can be reproduced with
+/// --gtest_random_seed=<seed>.
+
+#ifndef TEST_COMMON_TESTUTILS_HPP
+#define TEST_COMMON_TESTUTILS_HPP
+
+#include <gtest/gtest.h>
+#include <Kokkos_Core.hpp>
+#include <KokkosKernels_TestUtils.hpp>
+
+namespace {
+
+/// FNV-1a hash over the raw bytes of a contiguous host-accessible range.
+template <class T>
+uint64_t fnv1a(const T* data, size_t count) {
+  constexpr uint64_t kBasis = 14695981039346656037ULL;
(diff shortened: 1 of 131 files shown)


lucbv (commented):
Looks reasonable to me but copilot seem to have some reservation regarding the use of the random generator.
Also you might want a test that leverages your feature so that we know it does set the seed correctly...

test_common/KokkosKernels_TestUtils.hpp
```diff
@@ -4,6 +4,7 @@
 #ifndef KOKKOSKERNELS_TEST_UTILS_HPP
 #define KOKKOSKERNELS_TEST_UTILS_HPP
 
+#include <chrono>
 #include <random>
```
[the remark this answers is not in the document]
brian-kelley: Yep this is a real issue on some compiler/stdlib installations

jgfouca:
@lucbv , good idea. I will add a test to confirm that running with same seed gets identical result.

common/unit_test/Test_Common_TestUtils.hpp
```diff
@@ -0,0 +1,178 @@
...
+  for (size_t i = 0; i < count * sizeof(T); ++i) {
+    h ^= static_cast<uint64_t>(bytes[i]);
+    h *= kPrime;
+  }
+  return h;
+}
+
+/// Copy view to host and hash its contiguous data.
+template <class View>
+uint64_t hashView(const View& v) {
```
brian-kelley, line 38: Duplicate of ``KokkosKernels::Impl::hashView`` in KokkosKernels_SimpleUtils.hpp?

test_common/KokkosKernels_TestUtils.hpp
brian-kelley, line 374: Keep in mind that for ExeSpaceType other than Serial, ``Kokkos::fill_random`` is non-deterministic even if the seed is the same ([the note here](https://kokkos.org/kokkos-core-wiki/API/algorithms/Random-Number.html#construction-and-initialization) applies to Kokkos::fill_random, since it also uses the generator states in a parallel_for).

If determinism is required, this should also be replaced with a solution like in https://github.com/kokkos/kokkos/pull/1066, or with the ``std::mt19937`` generator on host.

brian-kelley (changes requested):

common/unit_test/Test_Common_TestUtils.hpp
brian-kelley: Actually, forget about this for now. The existing hashView has an overly strict static_assert that doesn't let it work with floating point data. We can merge these later.

test_common/KokkosKernels_TestUtils.hpp
jgfouca: Good catch, thanks

jgfouca:
Closing this PR. There is still a fair bit of work to do and I don't want every push to notify reviewers or trigger CI.

Issue #2083, opened 2024-01-05
HIP -O0 -g: spgemm producess incorrect entries


brian-kelley:
On rocm 5.6.0, MI50 (``AMD_GFX906``) and with ``-O0 -g``, spgemm reliably produces incorrect entries in the sparse unit test. This only happens with ``-O0`` (so ``-g`` by itself is not the cause).

```
[ RUN      ] hip.sparse_spgemm_double_int_size_t_TestDevice
entries are different.
1 2 7 18 19 24 26 31 35 38 46 51 53 58 63 72 86 96 100 101 ... ... ... 2282 2283 2285 2288 2307 2308 2310 2328 2343 2353 2358 2360 2362 2366 2369 2380 2389 2407 2423 2433 
1 2 7 18 19 24 26 31 35 38 46 51 53 58 63 72 86 96 100 101 ... ... ... 2282 2283 2285 2288 2307 2308 2310 2328 2343 2353 2358 2360 2362 2366 2369 2380 2389 2407 2423 2433 
/home/bmkelle/AMD_KK/kokkos-kernels/sparse/unit_test/Test_Sparse_spgemm.hpp:335: Failure
Value of: is_identical
  Actual: false
Expected: true
```

brian-kelley:
Unfortunately this is still a problem (now on rocm 6.2.0 and MI250x, a combination supported by Kokkos now)

khuck-AMD:
@brian-kelley I think I found a solution to this and will be submitting a PR later today: 
```
  * The key fix is letting the first-level hash accept inserts up to its actual full capacity
    (team_cuckoo_key_size) instead of cutting off early at max_first_level_hash_size.
  * So the hash table itself isn’t fundamentally bigger; we stop spilling too early.
  * I also kept the earlier safety fix that avoids undersized aliasing of L2 storage (c_row_size <
    pow2_hash_size), which is about correct storage usage rather than increasing nominal hash size.
```

lucbv:
Excellent, thanks for letting us know, we have had subtle issues with spgemm for a while, it's often hard to diagnose exactly what causes them. I'll be happy to have a look at the PR as well : )

lucbv:
@brian-kelley should we re-run this and check if that issue is now resolved?
I don't know that we still have rocm 5.6.0 but we can probably try on caraway with something like 6.2 or so?
If you don't have time to look at it let me know and I can kick-off a build

brian-kelley:
[...]

PR #3209, opened 2026-07-27
Fixing spgemm due to hash spilling

khuck-AMD:
With some help from Codex 5.3 Medium, iterated over the test to confirm that the test failures are related to spilling the hashes and some uninitialized variables. When failures do happen, additional diagnostic output is given and I kept that code.
The fixes include:

  * The key fix is letting the first-level hash accept inserts up to its actual full capacity (team_cuckoo_key_size) instead of cutting off early at max_first_level_hash_size.
  * So the hash table itself isn’t fundamentally bigger; we stop spilling too early.
  * I also kept the earlier safety fix that avoids undersized aliasing of L2 storage (c_row_size < pow2_hash_size), which is about correct storage usage rather than increasing nominal hash size.

Fix for #2083

### DIFF
--- a/sparse/unit_test/Test_Sparse_spgemm.hpp
+++ b/sparse/unit_test/Test_Sparse_spgemm.hpp
@@ -312,8 +312,9 @@ void test_spgemm(lno_t m, lno_t k, lno_t n, size_type nnz, lno_t bandwidth, lno_
       EXPECT_TRUE((res == 0)) << algo;
       bool is_identical = is_same_matrix<crsMat_t, device>(output_mat, output_mat2);
       EXPECT_TRUE(is_identical) << "SpGEMM result incorrect: algo=" << algo << ", callMode=" << int(callMode)
-                  << ", testReuse=" << int(testReuse) << ", m=" << m << ", k=" << k << ", n=" << n << ", nnz=" << nnz
-                  << ", bandwidth=" << bandwidth << ", row_size_variance=" << row_size_variance;
+                                << ", testReuse=" << int(testReuse) << ", m=" << m << ", k=" << k << ", n=" << n 
+                                << ", nnz=" << nnz << ", bandwidth=" << bandwidth 
+                                << ", row_size_variance=" << row_size_variance;
       // EXPECT_TRUE( equal) << algo;
     }
     // std::cout << "algo:" << algo << " spgemm_time:" << spgemm_time << "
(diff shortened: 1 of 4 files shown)


khuck-AMD:
before anyone reviews, my AMD colleagues are thinking there may be a race related to the hash-key probes, so trying some experiments with that too. Will amend this commit/PR as needed. Thanks!

sparse/unit_test/Test_Sparse_spgemm.hpp
```diff
@@ -312,6 +311,11 @@ void test_spgemm(lno_t m, lno_t k, lno_t n, size_type nnz, lno_t bandwidth, lno_
     if (!is_expected_to_fail) {
       EXPECT_TRUE((res == 0)) << algo;
       bool is_identical = is_same_matrix<crsMat_t, device>(output_mat, output_mat2);
+      if (!is_identical) {
+        std::cout << "SpGEMM mismatch context: algo=" << algo << ", callMode=" << int(callMode)
+                  << ", testReuse=" << int(testReuse) << ", m=" << m << ", k=" << k << ", n=" << n << ", nnz=" << nnz
+                  << ", bandwidth=" << bandwidth << ", row_size_variance=" << row_size_variance << std::endl;
+      }
       EXPECT_TRUE(is_identical) << algo;
```
brian-kelley:
```suggestion
      EXPECT_TRUE(is_identical) << "SpGEMM result incorrect: algo=" << algo << ", callMode=" << int(callMode)
                  << ", testReuse=" << int(testReuse) << ", m=" << m << ", k=" << k << ", n=" << n << ", nnz=" << nnz
                  << ", bandwidth=" << bandwidth << ", row_size_variance=" << row_size_variance;
```

brian-kelley:
@khuck-AMD Thanks for the PR! The rationale for the fix definitely sounds reasonable. I just added the github label that will allow autotesting to run. I'll be happy to approve when it passes.

brian-kelley:
@khuck-AMD Looks like this is mostly good to go except for formatting from my suggested change (can click on the github-FORMAT check to get a patch to fix this) and a few warnings like -Wsign-compare on the OSX builds.

lucbv (approved):

lucbv:
I'll let this go through testing but I expect to merge it in an hour or so! : )

PR #3210, opened 2026-07-27
KokkosSparse: improve ParILUT reuse-path numeric performance

MalachiTimothyPhillips:
Cache the sorted U^T pattern and U-to-U^T permutation for ParILUT reuse-numeric-pattern cache hits, allowing each iteration to refresh only transpose values instead of rebuilding and sorting the transpose structure.

Also make threshold-filter assignment team-parallel and specialize the compute_sum sparse merge for the non-async path. Exact threshold-selection behavior is unchanged.

Assisted-by: SandiaAI:ChatGPT-5.5

@jgfouca @lucbv

### DIFF
--- a/sparse/impl/KokkosSparse_par_ilut_numeric_impl.hpp
+++ b/sparse/impl/KokkosSparse_par_ilut_numeric_impl.hpp
@@ -114,6 +114,66 @@ struct IlutWrap {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jgfouca (approved):

jgfouca:
@lucbv, this PR looks good to me and so does CI. I think this can be merged.

lucbv (approved):

PR #3227, opened 2026-08-04
CodeQL: grouping dependabot updates of CodeQL

lucbv:
This should allow these updates to actually pass CI without having to create new PRs manually which also mean I won't need to bug a second person to review the manual PR either...
This follows the pattern of the same PR [#9399](https://github.com/kokkos/kokkos/pull/9399) that was made against Kokkos Core.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,10 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    group:
+      codeql-action:
+        pattern:
+	  - "github/codeql-action/*"
 
   - package-ecosystem: pip
     directory: /docs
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,6 +4,10 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    group:
+      codeql-action:
+        pattern:
+	  - "github/codeql-action/*"
```
dalg24:
```suggestion
        patterns:
          - "github/codeql-action/*"
```

dalg24: Pattern(s) plural and replaced tab with whitespaces

dalg24 (approved):

PR #3230, opened 2026-08-06
Perf Tests: removing perf tests for Benchmarks

lucbv:
Since we are using Benchmarks to monitor performance and not perf tests anymore, I would like to slowly replace the perf tests and remove them to avoid confusion and misconfiguration in testing. This is part of the long term effort to have the library properly performance tested and monitored using benchmarks.


jgfouca (approved):
I assume these BLAS routines already have benchmarks?

lucbv:
Yes they do, that's why I selected these and left alone the trtri and trmm ones for now but eventually I want to tackle them as well in Benchmark so we can migrate it all!

brian-kelley (approved):

lucbv:
The testing seems to be stuck on this PR after the GitHub issues from last week, I'll rebase the branch to have it run properly so we can merge

brian-kelley (approved):

ndellingwood (approved):

jgfouca (approved):

PR #3247, opened 2026-08-13
Make all TPL specializations partial

### LINKED ISSUE #2041: TPL wrappers instantiate more often than necessary
Common pattern in TPL wrappers:
```
KOKKOSBLAS2_DGEMV_CUBLAS(Kokkos::LayoutLeft, Kokkos::LayoutLeft,
                         Kokkos::LayoutLeft, Kokkos::CudaSpace, true)
KOKKOSBLAS2_DGEMV_CUBLAS(Kokkos::LayoutLeft, Kokkos::LayoutLeft,
                         Kokkos::LayoutLeft, Kokkos::CudaSpace, false)
```
where the final argument is used as the "eti_spec_avail" template argument in the unification layer:
```
  template <class ExecSpace>                                                  \
  struct GEMV<ExecSpace,                                                      \
              Kokkos::View<const Kokkos::complex<float>**, LAYOUTA,           \
                           Kokkos::Device<ExecSpace, MEM_SPACE>,              \
                           Kokkos::MemoryTraits<Kokkos::Unmanaged> >,         \
              Kokkos::View<const Kokkos::complex<float>*, LAYOUTX,            \
                           Kokkos::Device<ExecSpace, MEM_SPACE>,              \
                           Kokkos::MemoryTraits<Kokkos::Unmanaged> >,         \
              Kokkos::View<Kokkos::complex<float>*, LAYOUTY,                  \
                           Kokkos::Device<ExecSpace, MEM_SPACE>,              \
                           Kokkos::MemoryTraits<Kokkos::Unmanaged> >,         \
              true, ETI_SPEC_AVAIL> {                                         \
...
```
But we know what this should really be, it's just ``KokkosBlas::Impl::gemv_eti_spec_avail<ExecSpace, ...>::value``. Because the unification layor (spec.hpp) files always include eti_spec_avail before tpl_spec_decl, this definition is available.

Another thing we could do is compile each tpl_spec_decl file as a separate object file and link it into the library, so that we know anything there only gets compiled once. This would reduce the amount of code pulled in by ``#include "KokkosBlas2_gemv.hpp`` etc. in user code, and reduce the total build time for kernels that are used in multiple places.

brian-kelley:
Make all TPL specializations partial by templating them on ``bool ETI_SPEC_AVAIL``.

This accomplishes two things:
- Make TPL wrappers instantiate lazily, instead of eagerly in every translation unit where they're included. This speeds up the build, especially for wrappers than contain fallback kernels.
- Avoids the need to call DECL macros twice (with "true" and "false" plugged in for eti_spec_avail), but also without using the tedious eti_spec_avail expression with all kernel template arguments.

This would essentially resolve #2041 in a simpler way, IMO.

I tested a large configuration (serial+openmp+cuda, with TPLs) before and after to see the impact on build times and library sizes.
Before/develop:
- ``make -j8`` took 31:53
- libkokkoskernels.a was 230.5 MB
- The overall build directory was 1.75 GB

After:
- ``make -j8`` took 30:13
- libkokkoskernels.a was 170.8 MB (26% smaller)
- The build directory was 1.36 GB (22% smaller)

This is not the only solution to the issues identified in #2041. The other I can see would be moving the TPL specializations out of the headers and into their own .cpp files. This would then require more work:
- separate DECL/INST macros
- some more CMake logic
- using the long expressions to get the correct ``blah_eti_spec_avail<...>::value`` to plug into the specializations

I'm not sure all this is worth it, since the majority of the possible build time reduction is already achieved here with many fewer lines changed.

While working on this, it was also shown that our ETI system does work as intended. Object files that will be linked with ``libkokkoskernels`` (perf tests, for example) do contain ETI'd kernels as undefined symbols, which means they didn't re-compile them unnecessarily. If these object files did recompile kernels, they'd show up as weak symbols.

### DIFF
--- a/blas/tpls/KokkosBlas1_axpby_tpl_spec_decl.hpp
+++ b/blas/tpls/KokkosBlas1_axpby_tpl_spec_decl.hpp
@@ -26,137 +26,119 @@ inline void axpby_print_specialization() {
 namespace KokkosBlas {
 namespace Impl {
 
-#define KOKKOSBLAS1_DAXPBY_BLAS(ETI_SPEC_AVAIL)                                                                        \
-  template <typename ExecSpace>                                                                                        \
-    requires(std::is_same_v<typename ExecSpace::memory_space, Kokkos::HostSpace>)                                      \
-  struct Axpby<ExecSpace, double,                                                                                      \
-               Kokkos::View<const double*, Kokkos::LayoutLeft, ExecSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged> >,   \
-               double, Kokkos::View<double*, Kokkos::LayoutLeft, ExecSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged> >, \
-               1, true, ETI_SPEC_AVAIL> {                                                                              \
-    typedef double AV;                                                                                                 \
-    typedef double BV;                                                                                                 \
-    typedef Kokkos::View<const double*, Kokkos::LayoutLeft, ExecSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged> > XV;   \
-    typedef Kokkos::View<double*, Kokkos::LayoutLeft, ExecSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged> > YV;         \
-                                                                                                                       \
-    static void axpby(const ExecSpace& space, const AV& alpha, const XV& X, const BV& beta, const YV& Y) {             \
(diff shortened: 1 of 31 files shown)


lucbv:
@brian-kelley this looks reasonable to me but we need to resolve the conflict first, sorry for the slow review that probably made this stale. Let's retest tomorrow and we should be able to merge this!

brian-kelley:
No worries, it was a simple conflict. Rebased now.

brian-kelley:
Went ahead and rebased again to get #3256

lucbv (approved):

PR #3255, opened 2026-08-19
Add batched syrk

yasahi-hpc:
This PR adds batched [syrk](https://www.netlib.org/lapack/explore-html/d4/d6e/group__herk_ga09ec411a6b845c47efb0bc8bd3283b03.html) 

- [x] Serial/Team/TeamVector implementations
- [x] Corresponding unit-tests 
- [x] A new class `ConjNoTranspose` added to support [zherk](https://www.netlib.org/lapack/explore-html/d4/d6e/group__herk_ga5bd7d028f21ee95764845c1197baa09f.html#ga5bd7d028f21ee95764845c1197baa09f).

### DIFF
--- a/batched/dense/src/KokkosBatched_Syrk.hpp
+++ b/batched/dense/src/KokkosBatched_Syrk.hpp
@@ -1,0 +1,126 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
+#ifndef KOKKOSBATCHED_SYRK_HPP_
+#define KOKKOSBATCHED_SYRK_HPP_
+
+#include <KokkosBatched_Util.hpp>
+
+/// \author Yuuichi Asahi (yuuichi.asahi@cea.fr)
+
+namespace KokkosBatched {
+
+/// \brief Serial Batched Syrk:
+/// Performs one of the symmetric rank k operation
+///   C := alpha*A*A**T + beta*C
+///   C := alpha*A*A**H + beta*C
+///   C := alpha*A**T*A + beta*C
+///   C := alpha*A**H*A + beta*C
+///    where alpha and beta are real scalars, C is an n by n symmetric or hermitian matrix and
+///    A is an n by k matrix in the first case and a k by n matrix in the second case.
+///
+/// \tparam ArgUplo: Type indicating whether the upper (Uplo::Upper) or lower (Uplo::Lower) triangular part of A is
+/// modified
+/// \tparam ArgTrans: Type indicating whether the A (Trans::NoTranspose), or A**T (Trans::Transpose), or A**H
+/// (Trans::ConjTranspose), or AH (Trans::ConjNoTranspose) is used.
+///
+/// \tparam ScalarType: Input type for the scalar alpha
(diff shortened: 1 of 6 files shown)


batched/dense/src/KokkosBatched_Syrk.hpp
lucbv, line 1: Why not add static assertions here to check that A and C are rank-2 views?
That would make it clear from our public interface header what is expected from the users.

lucbv (approved):
Good in general and I realize that we are usually putting our static assertions in Impl headers. I wonder if there is value for the users to see them in the src header that is public and documents that API?

batched/dense/src/KokkosBatched_Syrk.hpp
yasahi-hpc: It depends where we implement the definition of the `invoke` method. Since `Views` are the template parameters of the `invoke` method (not class itself), we can only check them inside `invoke`. We can check their ranks if we move the definition of the `invoke` method from `KokkosBatched_Syrk_Impl.hpp` to `KokkosBatched_Syrk.hpp`.

yasahi-hpc: Also, it is possible to make them template parameters and partially deduce the types. Then, we can check in class definition.

```C++
template <typename ArgUplo, typename ArgTrans, typename ScalarType, typename AViewType, typename CViewType>
struct SerialSyrk {
  static_assert(Kokkos::is_view_v<AViewType>, "KokkosBatched::syrk: AViewType is not a Kokkos::View.");
  static_assert(Kokkos::is_view_v<CViewType>, "KokkosBatched::syrk: CViewType is not a Kokkos::View.");
  static_assert(AViewType::rank == 2, "KokkosBatched::syrk: AViewType must have rank 2.");
  static_assert(CViewType::rank == 2, "KokkosBatched::syrk: CViewType must have rank 2.");
  KOKKOS_INLINE_FUNCTION static int invoke(const ScalarType alpha, const AViewType &A, const ScalarType beta,
                                           const CViewType &C);
};
```

yasahi-hpc: > Also, it is possible to make them template parameters and partially deduce the types. Then, we can check in class definition.
> 
> ```c++
> template <typename ArgUplo, typename ArgTrans, typename ScalarType, typename AViewType, typename CViewType>
> struct SerialSyrk {
>   static_assert(Kokkos::is_view_v<AViewType>, "KokkosBatched::syrk: AViewType is not a Kokkos::View.");
>   static_assert(Kokkos::is_view_v<CViewType>, "KokkosBatched::syrk: CViewType is not a Kokkos::View.");
>   static_assert(AViewType::rank == 2, "KokkosBatched::syrk: AViewType must have rank 2.");
>   static_assert(CViewType::rank == 2, "KokkosBatched::syrk: CViewType must have rank 2.");
>   KOKKOS_INLINE_FUNCTION static int invoke(const ScalarType alpha, const AViewType &A, const ScalarType beta,
>                                            const CViewType &C);
> };
> ```

This is not possible. We do not have a constructor.

lucbv: Yeah, that's fine as it is, the only reasonable way to achieve a public interface that self document the requirements would be to use concepts and have a require and view ranks, etc...

PR #3262, opened 2026-08-20
fix(ode): replace static pivoting with partial in BDF solver

bcaddy:
# Summary

The BDF solver in Kokkos Kernels has several issues that were leading it being extremely slow for stiff ODE systems with one dense row

1. It used static pivoting that was very fragile with stiff ODE systems. I've replaced this with partial pivoting that is more fault tolerant
2. Check `LIN_SOLVE_FAIL` before applying the output of the linear solver in the Newton iterate
3. Make the BDF driver halve dt on LIN_SOLVE_FAIL, same as MAX_ITER. Only reachable if the Jacobian is singular

## Adding to older versions of Kokkos Kernels

Is it possible to add this patch to older versions of Kokkos Kernels? I'm currently stuck on Kokkos 4.7 for reasons outside of my control, if it's possible to add this to 4.7.05 that would be very helpful

## Note

I do not have a background in ODE solvers, numerical linear algebra, or the internals of Kokkos/Kokkos Kernels. All the code in this patch was written by Claude. I've done my best to verify it and it does produce correct results in my code (AthenaK) but I would really appreciate someone with more expertise taking a look. Thanks

### DIFF
--- a/ode/impl/KokkosODE_Newton_impl.hpp
+++ b/ode/impl/KokkosODE_Newton_impl.hpp
@@ -58,8 +58,35 @@ KOKKOS_FUNCTION KokkosODE::Experimental::newton_solver_status NewtonSolve(
     // compute LHS
     sys.jacobian(y0, J);
 
-    // solve linear problem
-    int linSolverStat = KokkosBatched::SerialGesv<KokkosBatched::Gesv::StaticPivoting>::invoke(J, update, rhs, tmp);
+    // Solve the linear problem J*update = rhs using LU with partial
+    // (row) pivoting. The static pivoting implemented in
+    // KokkosBatched::SerialGesv uses a greedy row-to-column assignment
+    // that can fail spuriously on well-conditioned sparse systems
+    // (e.g. chemistry Jacobians with one dense row), so it is only kept
+    // as a fallback for systems larger than the stack pivot buffer.
+    int linSolverStat = 0;
+    constexpr int pivot_buffer_size = 32;
+    if (sys.neqs <= pivot_buffer_size) {
+      int pivot_buffer[pivot_buffer_size];
+      Kokkos::View<int*, Kokkos::LayoutRight, Kokkos::AnonymousSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged>> piv(
+          pivot_buffer, sys.neqs);
+      for (int idx = 0; idx < sys.neqs; ++idx) update(idx) = rhs(idx);
+      linSolverStat = KokkosBatched::SerialGetrf<KokkosBatched::Algo::Getrf::Unblocked>::invoke(J, piv);
+      if (linSolverStat == 0) {
+        linSolverStat =
+            KokkosBatched::SerialGetrs<KokkosBatched::Trans::NoTranspose, KokkosBatched::Algo::Getrs::Unblocked>::invoke(
+                J, piv, update);
+      }
+    } else {
+      linSolverStat = KokkosBatched::SerialGesv<KokkosBatched::Gesv::StaticPivoting>::invoke(J, update, rhs, tmp);
+    }
+
+    // Return before touching y0 if the linear solve failed: applying the
+    // (stale or partial) update would corrupt the iterate.
+    if (linSolverStat != 0) {
+      return newton_solver_status::LIN_SOLVE_FAIL;
+    }
+
     KokkosBlas::SerialScale::invoke(-1, update);
 
     // update solution // x = x + alpha*update
--- a/ode/impl/KokkosODE_Newton_impl.hpp
+++ b/ode/impl/KokkosODE_Newton_impl.hpp
@@ -13,6 +13,8 @@
 #include "KokkosBlas1_axpby.hpp"
 #include "KokkosKernels_InnerProductSpaceTraits.hpp"
 
+#include "KokkosBatched_Getrf.hpp"
+#include "KokkosBatched_Getrs.hpp"
 #include "KokkosODE_Types.hpp"
 
 namespace KokkosODE {
--- a/ode/impl/KokkosODE_Newton_impl.hpp
+++ b/ode/impl/KokkosODE_Newton_impl.hpp
@@ -67,12 +67,11 @@ KOKKOS_FUNCTION KokkosODE::Experimental::newton_solver_status NewtonSolve(
     // (e.g. chemistry Jacobians with one dense row), so it is only kept
     // as a fallback for systems larger than the stack pivot buffer.
     int linSolverStat = 0;
-    constexpr int pivot_buffer_size = 32;
-    if (sys.neqs <= pivot_buffer_size) {
-      int pivot_buffer[pivot_buffer_size];
+    using mat_value_type = typename mat_type::non_const_value_type;
+    if (sys.neqs * sizeof(int) <= tmp.size() * sizeof(mat_value_type)) {
       Kokkos::View<int*, Kokkos::LayoutRight, Kokkos::AnonymousSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged>> piv(
-          pivot_buffer, sys.neqs);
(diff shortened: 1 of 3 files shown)


ode/impl/KokkosODE_Newton_impl.hpp
```diff
@@ -58,8 +58,35 @@ KOKKOS_FUNCTION KokkosODE::Experimental::newton_solver_status NewtonSolve(
...
+    // KokkosBatched::SerialGesv uses a greedy row-to-column assignment
+    // that can fail spuriously on well-conditioned sparse systems
+    // (e.g. chemistry Jacobians with one dense row), so it is only kept
+    // as a fallback for systems larger than the stack pivot buffer.
+    int linSolverStat = 0;
+    constexpr int pivot_buffer_size = 32;
+    if (sys.neqs <= pivot_buffer_size) {
+      int pivot_buffer[pivot_buffer_size];
+      Kokkos::View<int*, Kokkos::LayoutRight, Kokkos::AnonymousSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged>> piv(
+          pivot_buffer, sys.neqs);
```
yasahi-hpc: I think static allocation is less flexible. I would rather reuse `tmp` buffer.

```suggestion
   using mat_value_type = typename mat_type::non_const_value_type;
   if (sys.neqs * sizeof(int) <= tmp.size() * sizeof(mat_value_type)) {
      Kokkos::View<int*, Kokkos::LayoutRight, Kokkos::AnonymousSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged>> piv(
          reinterpret_cast<int *>(tmp.data()), sys.neqs);
```

yasahi-hpc (changes requested):
Could you introduce an unit-test that requires a partial-pivoting?

ode/impl/KokkosODE_Newton_impl.hpp
yasahi-hpc, line 15:
```suggestion
#include "KokkosBatched_Getrf.hpp"
#include "KokkosBatched_Getrs.hpp"
```

yasahi-hpc (changes requested):

ode/impl/KokkosODE_Newton_impl.hpp
bcaddy: fixed in 9ab429929

ode/impl/KokkosODE_Newton_impl.hpp
bcaddy: Fixed in e192ea729

bcaddy:
I've added a test in 916258fc5 that fails with static pivoting and passes with partial pivoting.

ode/unit_test/Test_ODE_BDF.hpp
```diff
@@ -389,6 +444,81 @@ void test_BDF_StiffChemistry() {
...
+  vec_type y0("initial conditions", mySys.neqs), y_new("solution", mySys.neqs);
+  auto y0_h    = Kokkos::create_mirror_view(y0);
+  auto y_new_h = Kokkos::create_mirror_view(y_new);
+
+  // Reference solution
+  Kokkos::View<scalar_type*, Kokkos::HostSpace> y_ref_h("reference solution", mySys.neqs);
+  Kokkos::deep_copy(y0_h, 1.0);
+  mySys.solution(t_end, y0_h, y_ref_h);
+
+  Kokkos::RangePolicy<execution_space> myPolicy(0, 1);
```
yasahi-hpc: We prefer variables in snake case.

```suggestion
  Kokkos::RangePolicy<execution_space> my_policy(0, 1);
```

ode/unit_test/Test_ODE_BDF.hpp
```diff
@@ -389,6 +444,81 @@ void test_BDF_StiffChemistry() {
...
+    mv_type kstack("Startup RK vectors", 6, mySys.neqs);
+    mv_type y_vecs("history vectors", mySys.neqs, 1);
+
+    Kokkos::deep_copy(scale, 1);
+    Kokkos::deep_copy(y0, 1.0);
+
+    BDFSolve_wrapper<SparseLinearSystem, KokkosODE::Experimental::BDF_type::BDF1, vec_type, mv_type, mat_type,
+                     scalar_type>
+        solve_wrapper(mySys, t_start, t_end, num_steps, y0, y_new, rhs, update, scale, y_vecs, kstack, temp, jac);
+    Kokkos::parallel_for(myPolicy, solve_wrapper);
```
yasahi-hpc:
```suggestion
    Kokkos::parallel_for(my_policy, solve_wrapper);
```

ode/unit_test/Test_ODE_BDF.hpp
```diff
@@ -389,6 +444,81 @@ void test_BDF_StiffChemistry() {
...
+  // Adaptive time step BDFSolve
+  {
+    mat_type temp("buffer1", mySys.neqs, 23 + 2 * mySys.neqs + 4), temp2("buffer2", 6, 7);
+
+    Kokkos::deep_copy(y0, 1.0);
+    Kokkos::deep_copy(y_new, 0.0);
+
+    BDF_Solve_wrapper bdf_wrapper(mySys, t_start, t_end, scalar_type(0.0), (t_end - t_start) / 10, y0, y_new, temp,
+                                  temp2);
+    Kokkos::parallel_for(myPolicy, bdf_wrapper);
```
yasahi-hpc:
```suggestion
    Kokkos::parallel_for(my_policy, bdf_wrapper);
```

ode/unit_test/Test_ODE_BDF.hpp
```diff
@@ -389,6 +444,81 @@ void test_BDF_StiffChemistry() {
...
+    mat_type jac("jacobian", mySys.neqs, mySys.neqs), temp("temp storage", mySys.neqs, mySys.neqs + 4);
+    mv_type kstack("Startup RK vectors", 6, mySys.neqs);
+    mv_type y_vecs("history vectors", mySys.neqs, 1);
+
+    Kokkos::deep_copy(scale, 1);
+    Kokkos::deep_copy(y0, 1.0);
+
+    BDFSolve_wrapper<SparseLinearSystem, KokkosODE::Experimental::BDF_type::BDF1, vec_type, mv_type, mat_type,
+                     scalar_type>
+        solve_wrapper(mySys, t_start, t_end, num_steps, y0, y_new, rhs, update, scale, y_vecs, kstack, temp, jac);
```
yasahi-hpc, line 491: Can you use CTAD here as well as in line 510?

```suggestion
    BDFSolve_wrapper solve_wrapper(mySys, t_start, t_end, num_steps, y0, y_new, rhs, update, scale, y_vecs, kstack, temp, jac);
```

yasahi-hpc (changes requested):
LGTM

Some small suggestions

ode/unit_test/Test_ODE_BDF.hpp
bcaddy: Fixed in c9cf5ff. That issue occurred in several other tests so I fixed it there too.

I also added `build/` to the gitignore so I didn't have diffs of 750+ files after a build.

ode/unit_test/Test_ODE_BDF.hpp
bcaddy: I can't use it there since `bdf_type` is a non-type template parameter. Other tests in that same file use it like this as well.

Line 510 can use CTAD because all template arguments are types and they all appear in the constructor arguments.

yasahi-hpc: Sure

yasahi-hpc (approved):
LGTM
Can you revert changes in .gitignore
I am fine with your suggestions but it should be made in a separate PR

bcaddy:
Good point, done

lucbv (approved):

PR #3268, opened 2026-08-27
chore: expand gitignore to include many other commonly ignored files

bcaddy:
My primary goal with this is to ignore the `build/` directory so I don't have 750+ files show up in git after a build. As long as I was at it though I added the standard set of things I add to all my gitignore files. I'm happy revert this to just adding `build/` to the ignore list if you'd prefer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,21 +1,62 @@
-# Standard ignores
-*~
+# Files specific to this repo #
+##############################
+docs/generated_docs/
+build/
+compile_commands.json
+
+# Compiled source #
+###################
+*.com
+*.class
+*.dll
+*.exe
+*.o
+*.so
+*.mod
+*.a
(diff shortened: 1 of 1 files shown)


brian-kelley (approved):

brian-kelley:
We generally do our builds outside the source tree, but there's no good reason for that to be a requirement. I like this change.

bcaddy:
Two of the tests are failing. I'm not sure how changing the gitignore could break anything. Is that really related to this change or is it something else?

brian-kelley:
@bcaddy Those always fail - either something isn't set up correctly with the job, or there are other issues we have to deal with. They aren't required checks though so I went ahead and merged this.

bcaddy:
Good to know. Thanks

PR #3272, opened 2026-08-28
Implement batched symv

yasahi-hpc:
This PR adds batched [symv](https://www.netlib.org/blas/dsymv.f) 

- [x] Serial/Team/TeamVector implementations
- [x] Corresponding unit-tests


batched/dense/src/KokkosBatched_Symv.hpp
```diff
@@ -0,0 +1,131 @@
...
+  /// \param[in] alpha: Scalar alpha
+  /// \param[in] A: A is a dimension ( lda, n ) matrix. Before entry with uplo = Uplo::Upper, the leading n by n upper
+  /// triangular part of the array A must contain the upper triangular part of the symmetric matrix and the strictly
+  /// lower triangular part of A is not referenced. Before entry with uplo = Uplo::Lower, the leading n by n lower
+  /// triangular part of the array A must contain the lower triangular part of the symmetric matrix and the strictly
+  /// upper triangular part of A is not referenced. \param[in] x: x is a dimension ( n ) vector \param[in] beta: Scalar
+  /// beta \param[in,out] y: y is a dimension ( n ) vector. Before entry, y must contain the vector y. On exit, y is
+  /// overwritten by the result ( alpha*op( A )*x + beta*y )
+  ///
+  /// No nested parallel_for is used inside of the function.
```
lucbv: Is that true for the Team level implementation, that would probably waste some resources?

batched/dense/src/KokkosBatched_Symv.hpp
```diff
@@ -0,0 +1,131 @@
...
```
lucbv: Same as above, if we do not use nested parallelism why not only implement the Serial variant?

lucbv (changes requested):
I think some of the comments are wrong and should be corrected a bit, otherwise this looks good to me.

lucbv (approved):

PR #3273, opened 2026-08-28
Batched Serial Householder: use abs instead of real in norm calculation

maartenarnst:
This PR is motivated by a compilation error we see in a Trilinos build with Stokhos enabled, and with the option "Stokhos_ENABLE_Ensemble_Reduction" set to "ON".

```
INFO:root:#19 2788.6 /Trilinos-sources/packages/kokkos-kernels/batched/dense/impl/KokkosBatched_Householder_Serial_Internal.hpp:38:22: error: no match for 'operator+=' (operand types are 'mag_type' {aka 'double'} and 'KokkosKernels::ArithTraits<Sacado::MP::Vector<Stokhos::StaticFixedStorage<int, double, 16, Kokkos::Serial> > >::val_type' {aka 'Sacado::MP::Vector<Stokhos::StaticFixedStorage<int, double, 16, Kokkos::Serial> >'})
INFO:root:#19 2788.6    38 |       norm_x2_square += KAT::real(KAT::conj(x2_at_i) * x2_at_i);
```

The issue appears to be that with ensemble reduction, the `norm_x2_square` has magnitude type and is "reduced to a" scalar, whereas the conjugate and the real part are left as ensembles.

The proposed fix is to replace `KAT::real` with `KAT::abs` in this line. Indeed, with ensemble reduction, the `abs` function reduces to a scalar.

Tagging @cgcgcg as the last PR that had modified this line was:
- #3051

@etphipp

### DIFF
--- a/batched/dense/impl/KokkosBatched_Householder_Serial_Internal.hpp
+++ b/batched/dense/impl/KokkosBatched_Householder_Serial_Internal.hpp
@@ -35,7 +35,7 @@ struct SerialLeftHouseholderInternal {
     mag_type norm_x2_square = zero;
     for (int i = 0; i < m_x2; ++i) {
       const auto x2_at_i = x2[i * x2s];
-      norm_x2_square += KAT::real(KAT::conj(x2_at_i) * x2_at_i);
+      norm_x2_square += KAT::abs(KAT::conj(x2_at_i) * x2_at_i);
     }
 
     /// if norm_x2 is zero, return with trivial values


lucbv:
@maartenarnst this looks wrong to me, the `norm_x2_square` should be an ensemble as each vector in the ensemble is different and hence has a different norm. I think the real issue is that mag_type is not defined correctly to be an ensemble type.
This is probably something that needs to be looked at in the overload of `KokkosKernels::ArithTraits` in Stokhos?

lucbv (changes requested):
Blocking this until we can have confirmation that the proposed fix indeed implements the proper math.

etphipp:
> @maartenarnst this looks wrong to me, the `norm_x2_square` should be an ensemble as each vector in the ensemble is different and hence has a different norm. I think the real issue is that mag_type is not defined correctly to be an ensemble type. This is probably something that needs to be looked at in the overload of `KokkosKernels::ArithTraits` in Stokhos?

@lucbv Stokhos supports two modes for the ensemble scalar type:  one where magnitude type is a scalar and one where it is an ensemble.  With Stokhos_ENABLE_Ensemble_Reduction=ON, it is a scalar.  In this mode, Stokhos views the ensemble system as a block-diagonal system, in which case magnitudes should be scalars.  It has the advantage of being simpler without requiring special orthogonalization implementations for ensembles, but has the disadvantage of coupling the spectra of the ensembles together.  So I think the change @maartenarnst is proposing is correct.

maartenarnst:
Hey @lucbv and @etphipp.

Some additional context is that the compilation error arises when Stokhos instantiates the MueLu inverse approximation factory in `MueLu_InverseApproximationFactory_MP_Vector_16_Serial.cpp`. This happens as part of how the Stokhos ETI instantiates a large number of MueLu classes. But I think there is currently no Stokhos test or example that uses the inverse-approximation factory. The Stokhos test that solves a linear problem with MueLu uses a Chebychev preconditioner. 

I should have clarified that I meant the PR as a reasonable way to unblock the build. The PR keeps the behavior unchanged for builds that currently compile. And it unblocks the build for the case of ensemble propagation with reduction. It does this by calculating the variable in a way that is consistent with how ensemble propagation with reduction calculates norms.

I agree that the mathematics are a valid concern. I did not mean the PR to address the mathematics of Householder transformations under ensemble propagation. I think it would be a significant study, such as how the transformation is defined with the reduced inner product and how the surrounding solver uses it. We'd also have to look at other issues in the code such as the ternary and the existing `abs` computations (currently not sums of squares).  

In our own Trilinos builds, we have implemented the proposed patch, and our pipelines are green. We do not currently use MueLu with ensembles, so there is no urgency from our side. 

To me it's fine to proceed with the PR or to look for another solution that you are more comfortable with. Perhaps a more targeted approach to select MueLu classes to instantiate from the Stokhos ETI could be an option.

lucbv:
@etphipp @maartenarnst 
I am fine going with this since you are in agreement on the math, I'll rebase, run the CI and see where this stands

lucbv (approved):

PR #3274, opened 2026-08-28
  Implement right-transpose SerialApplyQ

### LINKED ISSUE #3087: Implement batched SerialApplyQ on the right with transpose
I am using the batched `SerialApplyQ` in an application where applying $Q^T$ on the right would let me use less device scratch memory for intermediate computations. However, that is the only one of the four left/right, transpose/no-transpose combinations that is [not implemented](https://github.com/kokkos/kokkos-kernels/blob/25c1ed6b034a579e528660535b5b3093bb543c23/batched/dense/impl/KokkosBatched_ApplyQ_Serial_Impl.hpp).

alvaroborras:
Fixes #3087.

Adds the missing `SerialApplyQ<Side::Right, Trans::Transpose, Algo::ApplyQ::Unblocked>` specialization.

The implementation reuses the existing right-side Householder operation, traversing the reflectors in reverse order and conjugating `tau` for complex scalars.

The existing SerialQR tests now verify that applying `Q` followed by `Q^H` from the right recovers the original matrix. Existing coverage includes square and rectangular factorizations and the supported real and complex scalar types.

Testing:
- SerialQR batch tests pass for all four scalar types.
- Full `batched_dla_serial` test suite passes.

### DIFF
--- a/batched/dense/impl/KokkosBatched_ApplyQ_Serial_Impl.hpp
+++ b/batched/dense/impl/KokkosBatched_ApplyQ_Serial_Impl.hpp
@@ -110,6 +110,38 @@ KOKKOS_INLINE_FUNCTION int SerialApplyQ<Side::Right, Trans::NoTranspose, Algo::A
                                                    B.stride(1), w.data());
 }
 
+template <>
+template <typename AViewType, typename tViewType, typename BViewType, typename wViewType>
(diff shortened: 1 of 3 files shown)


lucbv (approved):
Looks good to me, thanks for contributing the feature needed, we appreciate your help a lot!

PR #3276, opened 2026-08-28
fix(ode): norm was not accumulating or resetting

bcaddy:
`norm_new` in `NewtonSolve` was not accumulating on each inner loop iteration and wasn't being reset to zero between outer loop iterations. Fixed both issues.

This was causing an issue when I was using the solver. I'm solving a system of ODEs as part of a larger time step and if I varied the sizes of the larger time steps but evolved to the same time it would cause significant differences in the results.


lucbv (approved):
Thanks that's a good catch

PR #3277, opened 2026-08-28
fix(ode): reject all newton solves that don't succeed

bcaddy:
Before only a subset of failure modes triggered a step size reduction and retry. Now anything that isn't `NLS_SUCCESS` triggers a halving of `dt` and a retry

### DIFF
--- a/ode/impl/KokkosODE_BDF_impl.hpp
+++ b/ode/impl/KokkosODE_BDF_impl.hpp
@@ -366,7 +366,8 @@ KOKKOS_FUNCTION void BDFStep(ode_type& ode, scalar_type& t, scalar_type& dt, sca
       update(eqIdx) = y_new(eqIdx) - y_predict(eqIdx);
     }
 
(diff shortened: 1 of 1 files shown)


bcaddy:
In testing this in my code I've found that it exposes some other issues since a bunch of unconverged steps are no longer accepted. I'm working on figuring out exactly what the issue(s) are and will post a PR once I do.

bcaddy:
> In testing this in my code I've found that it exposes some other issues since a bunch of unconverged steps are no longer accepted. I'm working on figuring out exactly what the issue(s) are and will post a PR once I do.

I believe I've sorted these out. I need to clean up my code but I will submit the fixes in a series of PRs next week.

lucbv (approved):

lucbv:
I suspect that we could be a bit more accurate and could potentially decide to propagate errors that are not recoverable from reducing the time step but we can think about it later.

PR #3295, opened 2026-09-11
fix(ode): `param.iters` never set so `safety` was constant

bcaddy:
Since `param.iters` was never update the calculation for `safety` in the BDF solver was just computing a constant and so taking overly aggressive steps. This change fixes that and adds a regression test for this bug.

Assisted-by: Claude:fable-5


ode/src/KokkosODE_Newton.hpp
lucbv, line 1: I do think that the user parameters should be taken as const if possible. This gives some guarantees that we will not overwrite them internally be it on purpose or on accident. If the only thing we want here is to report the number of Newton iteration taken maybe we should just split that up from the Newton_params struct?

lucbv (changes requested):
Unfortunately I think it is a design flaw to have added iters to the parameters. It is not really a parameter for the algorithm, just a tracked state so it should be split up and the API needs to change a bit instead of making params non-const.
Let me know if you have strong thoughts on this?

ode/src/KokkosODE_Newton.hpp
bcaddy: I agree. Splitting it out is much better. I'll make that change and let you know when it's fixed

bcaddy: I've removed `iters` from `params` and replaced it with a return parameter in the Newton solver. Is this more what you were thinking?

lucbv (approved):
Yes, I like that more!

lucbv:
Looks like on host there is some interaction between this change and the convergence of the solver. Both host builds are seemingly in infinite loops while the gpu ones are fine.

PR #3298, opened 2026-09-11
fix(ode): infinite loop caused by too small `min_step`

bcaddy:
`min_step` was `ArithTraits<double>::min()` (≈2.2e-308 for fp64). When steps keep getting rejected, `dt` shrinks until it drops below the floating-point spacing of `t`, at which point `t + dt == t` and `dt = t_new - t` becomes exactly 0. In this case, `BDFSolve`'s while loop spins forever. This fix limits `dt` to be ~10x larger than the size between `t` and `t+1ulp` and makes `BDFStep` and `BDFSolve` return a status just like the Newton and RK solvers.

Updated all calls to `BDFStep` and `BDFSolve` to accept the returned status. If that call is in a test it will check for success.

Added a regression test to check for this bug. Since the bug manifests as an infinite loop the test checks if a certain number of iterations has been exceeded and if it has it calls `Kokkos::abort`. Not ideal since it basically crashes that test but it does work without breaking the the execution of other tests.

Assisted-by: Claude:fable-5


lucbv (approved):
Okay, that's reasonable and thanks for adding a new test that checks this infinite loop condition

PR #3302, opened 2026-09-14
fix(ode): add a noise guard

bcaddy:
When the solution is at equilibrium sometimes floating point noise can cause iterations to be incorrectly classified as diverging. I added a check to the divergence branch to additionally require that the new norm, and therefore the update, must be larger than the tolerance to trigger a divergence failure.

Assisted-by: Claude:fable-5


lucbv (approved):
Reasonable, although this is changing the divergence criteria so it could disturb some of the unit-tests checks, will need to observe the CI results closely

PR #3309, opened 2026-09-16
Batched - Syrk: adding special case for early return

lucbv:
When A is degenerate (length 0) or alpha == 0, we only need to scale the matrix C and return. Doing this early avoids getting a warning for lda < 1 which happens with a degenrate A.


batched/dense/impl/KokkosBatched_Syrk_Impl.hpp
yasahi-hpc, line 33:
```suggestion
if (lda < Kokkos::max(0, nrowa)) {
Kokkos::printf(
"KokkosBatched::syrk: leading dimension of A must not be smaller than "
"max(0, nrowa): "
"lda = %d, nrowa = %d\n",
lda, nrowa);
return 1;
}
```

batched/dense/impl/KokkosBatched_Syrk_Impl.hpp
```diff
@@ -61,6 +61,22 @@ KOKKOS_INLINE_FUNCTION int SerialSyrk<ArgUplo, ArgTrans>::invoke(const ScalarTyp
...
+      }
+    } else {
+      for (int colIdx = 0; colIdx < n; ++colIdx) {
+        for (int rowIdx = 0; rowIdx < colIdx + 1; ++rowIdx) {
+          C(rowIdx, colIdx) = beta * C(rowIdx, colIdx);
+        }
+      }
+    }
+    return 0;
+  }
```
yasahi-hpc, line 79:
```suggestion
```

batched/dense/impl/KokkosBatched_Syrk_Impl.hpp
```diff
@@ -105,6 +121,20 @@ KOKKOS_INLINE_FUNCTION int TeamSyrk<MemberType, ArgUplo, ArgTrans>::invoke(const
...
+                             [&](const int rowIdx) { C(rowIdx, colIdx) *= beta; });
+      });
+    } else {
+      Kokkos::parallel_for(Kokkos::TeamThreadRange(member, n), [&](const int colIdx) {
+        Kokkos::parallel_for(Kokkos::ThreadVectorRange(member, 0, colIdx + 1),
+                             [&](const int rowIdx) { C(rowIdx, colIdx) *= beta; });
+      });
+    }
+    return 0;
+  }
```
yasahi-hpc, line 137:
```suggestion
```

batched/dense/impl/KokkosBatched_Syrk_Impl.hpp
```diff
@@ -150,6 +180,22 @@ KOKKOS_INLINE_FUNCTION int TeamVectorSyrk<MemberType, ArgUplo, ArgTrans>::invoke
...
+      });
+    } else {
+      Kokkos::parallel_for(Kokkos::TeamVectorRange(member, n), [&](const int colIdx) {
+        for (int rowIdx = 0; i < colIdx + 1; ++rowIdx) {
+          C(rowIdx, colIdx) *= beta;
+        }
+      });
+    }
+    return 0;
+  }
```
yasahi-hpc, line 198:
```suggestion
```

batched/dense/impl/KokkosBatched_Syrk_Impl.hpp
yasahi-hpc: `SyrkInternal` will automatically compute this.

yasahi-hpc (changes requested):
If we would like to allow A (N, 0) and C (N, N) case, we also need to update the test.

https://github.com/kokkos/kokkos-kernels/blob/957002b67629a6ee9af91e87a2b9e4c24b6ec7a8/batched/dense/unit_test/Test_Batched_Syrk.hpp#L320-#L334 

should be suppressed as well

PR #3310, opened 2026-09-16
Fix out-of-rank extent access in Blas1 dot, sum and Utils permute vector

ndellingwood:
Detected in Trilinos testing with Kokkos_ENABLE_DEPRECATED_CODE_5=OFF

Reported by @cgcgcg, see e.g. https://my.cdash.org/tests/478171938


brian-kelley (approved):

lucbv (approved):
This is fine but it points out the fact that these utility functions could probably use some refactor to assert some properties at compile time and run time such as ranks, size compatibility, etc...

ndellingwood:
Going to push one more fix shortly for KokkosBlas1_dot.hpp

ndellingwood:
Blas1 dot fix pushed in [378649b](https://github.com/kokkos/kokkos-kernels/pull/3310/commits/378649bab7bf8ae57684e799c23ae5ac38179972) , ready for re-review

ndellingwood:
Blas1 sum fix pushed here [378649b](https://github.com/kokkos/kokkos-kernels/pull/3310/commits/378649bab7bf8ae57684e799c23ae5ac38179972)

lucbv (changes requested):
To simplify the check remove `&& RV::rank == 1` on line 187.
Then consolidate the if and else if close, I think you just need the if close with some checks on line 195 to avoid calling extent when RV::rank == 0?

ndellingwood:
@lucbv thanks, I simplified the changes in [7cc7208](https://github.com/kokkos/kokkos-kernels/pull/3310/commits/7cc7208c25958f4e13ec87a97a023ea1e1b69e38) I think the way you asked and retested, ready for re-review

blas/src/KokkosBlas1_dot.hpp
```diff
@@ -184,25 +184,46 @@ void dot(const execution_space& space, const RV& R, const XMV& X, const YMV& Y,
...
+  if constexpr (XMV::rank == 2 && YMV::rank == 2) {
+    if (X.extent(0) != Y.extent(0)) {
+      dimsMatch = false;
+    } else if (X.extent(1) != Y.extent(1) && X.extent(1) != 1 && Y.extent(1) != 1) {
+      // Numbers of columns don't match, and neither X nor Y have one column.
+      dimsMatch = false;
+    }
+    if constexpr (RV::rank == 1) {
+      const auto maxNumCols = X.extent(1) > Y.extent(1) ? X.extent(1) : Y.extent(1);
+      if (RV::rank == 1 && R.extent(0) != maxNumCols) {
```
lucbv: The compiler already did the static check on RV::rank == 1, let's skip it here and save a comparison instruction
```suggestion
      if (R.extent(0) != maxNumCols) {
```

blas/src/KokkosBlas1_dot.hpp
```diff
@@ -184,25 +184,46 @@ void dot(const execution_space& space, const RV& R, const XMV& X, const YMV& Y,
...
+        dimsMatch = false;
+      }
+      if (!dimsMatch) {
+        std::ostringstream os;
+        os << "KokkosBlas::dot: Dimensions of R, X, and Y do not match: ";
+        if (RV::rank == 1) {
+          os << "R: " << R.extent(0) << " x " << X.extent(1) << ", ";
+        }
```
lucbv: Same as above
```suggestion
        os << "R: " << R.extent(0) << " x " << X.extent(1) << ", ";
```

lucbv (changes requested):
A little more cleaning to avoid checking at compile time and run time that `RV::rank == 1`

blas/src/KokkosBlas1_dot.hpp
ndellingwood: Good call, applied in 22a7e055da485bcca657d017ee837565950ca84e

blas/src/KokkosBlas1_dot.hpp
ndellingwood: Good call, applied in 22a7e055da485bcca657d017ee837565950ca84e

lucbv (approved):

PR #3307, opened 2026-09-15
Fix bugs in graph distance 2 color

jgfouca:
1. VB_BIT d1-check used `>` instead of `>=`, missing conflicts when a d1 neighbor's color equaled `offset`.

2. resolveConflictsSerial allocated `forbidden[nr]` but could read index `nr`

Bug 1 fed into Bug 2, which is why the crash only showed up on `VB_BIT`.

PR #3312, opened 2026-09-18
Add clique test for D2 coloring

brian-kelley:
Follows #3307 which fixed two bugs in D2 coloring discovered due to failures on randomly generated graphs.

This PR adds a couple of non-random test cases for D2 coloring on cliques of nodes (N=64 and N=130). Since they're cliques, they need exactly N colors. The N=64 case catches the case where we were doing an out-of-bounds write to `forbidden` in resolveConflictsSerial. This would typically only get caught when running under valgrind or address sanitizer. The N=130 case exercises the case where we have to loop over multiple color sets for VBBIT, where a color set is 64 colors.

Since all the parallel D2 coloring algorithms (VB, VBBIT, NB) are nondeterministic, these cases don't always trigger the bugs, but it's better than nothing.


lucbv (approved):

jgfouca:
Thanks, @brian-kelley ! I was out the last half of last week, so that's why I didn't comment here.

PR #3314, opened 2026-09-21
Citation: adding our initial citation file

lucbv:
This points to zenodo and might in the future point to a journal article. The list of authors/contributors are the current members of the team on github.


CITATION.cff
```diff
@@ -0,0 +1,61 @@
...
+  coarse-grained operations; they take a lot of work and
+  make sense to parallelize inside using Kokkos. Kokkos
+  Kernels can be the building block of a parallel linear
+  algebra library that uses MPI and threads for parallelism,
+  or it can be used stand-alone in your application.
+keywords:
+  - Performance portability
+  - High Performance Computing
+  - Mathematics Library
+license: Apache-2.0
```
dalg24: WITH LLVM-exception?

CITATION.cff
```diff
@@ -0,0 +1,61 @@
...
+    family-names: Kelley
+    affiliation: Sandia National Laboratories
+    orcid: 'https://orcid.org/0000-0003-3607-360X'
+  - given-names: Ichitaro
+    family-names: Yamazaki
+    affiliation: Sandia National Laboratories
+    orcid: 'https://orcid.org/0009-0003-7248-573X'
+identifiers:
+  - type: doi
+    value: 10.5281/zenodo.17641723
```
dalg24, line 42: This one points to version 5.1.1

dalg24 (commented):
Do you want to mention the arXiv paper mentioned in https://kokkos.org/citing-kokkos/#kokkos-kernels ?

CITATION.cff
lucbv: I tried to add it in the generator tool but it is still a license that's not generally recognized so I debated adding it manually?

CITATION.cff
lucbv: I respectfully disagree, this is the "Concept DOI" that points to all versions of the library according to Zenodo.
[https://doi.org/10.5281/zenodo.17641723](https://doi.org/10.5281/zenodo.17641723)
It currently points to 5.2.1 but it will update itself overtime to the latest version I believe?

dalg24: Weird, you are right, sorry about the noise.

CITATION.cff
lucbv: Added the LLVM Exception mention

